PR #2395, opened 2012-09-21
Fix for http://framework.zend.com/issues/browse/ZF2-571

delatbabel:
Hopefully this is the correct pull request to fix http://framework.zend.com/issues/browse/ZF2-571

There should only be one commit here, one line change to Form/Element/Checkbox.php


stillatmylinux:
There is a work around for this.  Instead of changing the boolean value in the core file you can just add a InputFilter

    namespace MyModule\Form;
    use Zend\InputFilter\InputFilter;

    class MyFormFilter extends InputFilter {
        public function __construct() {

            $this->add(array(
                'name'       => 'myCheckboxes',
                'required'   => false,
            ));
        }
    }

PR #3338, opened 2013-01-03
Hostname RouteInterface non standard port handling

richardjennings:
Added test "non-standard-port" to demonstrate failed handling of non standard port specifications for Hostname. 
eg. foo.example.com:8080

Modification of  'testMatching' to use non standard port if specified.

Modification of Hostname RouteInterface to allow matching and assembling of non standard ports.

Modification of Hostname RouteInterface to match non standard port if specified.


kenkataiwa:
Was this matter ever resolved?

Ocramius:
@kenkataiwa no, but as you can see, @DASPRiD is proposing to use a `Port` route. You should give it a try yourself: shouldn't be too complicated.

PR #3438, opened 2013-01-15
Add @method tags to the AbstractController

micheh:
This pull request adds the convenience methods for the controller plugins to the AbstractController via the @method tag, which enables full code completion in the IDE.


apoca:
I'm using phpstorm and code completion don't work for ZF2 framework... How can I solve this!?

PR #4122, opened 2013-03-26
Update EM->trigerUntil to be an alias of trigger

EvanDotPro:
They contained the exact same code/logic. The only difference was the
conditional around checking if the callback was valid, but the outcome
is the same for both. This simply reduces duplicate code and causes no
change in behavior at all.


weierophinney:
Marked as a BC break, as it emits a deprecation notice (which can and does break unit tests on code that exercises the method!).

PR #4149, opened 2013-03-30
[ZF3] FQCN in parameter type hints

Ocramius:
This PR introduces FQCNs for type-hints in generated method signatures. This is a BC break, but it is necessary to use the code generator when the generated code is not placed on a dedicated file.

Before:

``` php
public function doFoo(Foo $foo) {}
```

After

``` php
public function doFoo(\Foo $foo) {}
```

This PR also takes into account the new PHP 5.4 `callable` data type when generating code.


GeeH:
New issue can be found at: https://api.github.com/repos/zendframework/zendframework/issues/4149

PR #4176, opened 2013-04-04
"X-HTTP-Method-Override" header support in AbstractRestfulController

etodanik:
Working with Backbone.js I encountered the need to use the "X-HTTP-Method-Override" header to simulate the different HTTP verbs. This little patch takes the header into account if it exists.

### DIFF
--- a/library/Zend/Mvc/Controller/AbstractRestfulController.php
+++ b/library/Zend/Mvc/Controller/AbstractRestfulController.php
@@ -248,6 +248,12 @@ abstract class AbstractRestfulController extends AbstractController
 
         // RESTful methods
         $method = strtolower($request->getMethod());
+        
+        // Check if we're emulating REST HTTP (for servers that don't support all HTTP verbs)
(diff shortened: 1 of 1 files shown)


r0mdau:
Another need : receiving a form in a RestfulController to modify an existing object. Or delete it.

To solve, I add a hidden input with name of x-http-method-override to route POSTed form to PUT or DELETE AbstractRestfulController methods.

And override AbstractRestfulController::onDispatch(MvcEvent $e) including : 
```php
// RESTful methods
$method = strtolower($request->getMethod());

if ($method == "post") {
    $override = strtolower($request->getPost('x-http-method-override'));
    if (in_array($override, ['delete', 'put'])) {
        $method = $override;
    }
}
```

weierophinney:
@r0mdau Please do not comment on closed issues; additionally, please note that issues have moved to the individual repositories for the components. If you have changes you would like to request for the AbstractRestfulController, or a patch you would like to submit, please do so against the [zend-mvc](https://github.com/zendframework/zend-mvc) repository.

PR #4238, opened 2013-04-16
ValueGenerator constant detection

3axap4eHko:
For example, we need to get a file like this:

```
<?php
return array(
    'module' => array(
        'moduleDir' => __DIR__ . '/../module'
    )
);
```

but we cant pass argument `__DIR__`, because this constant will be parsed and replaced to value, and we must quote it '**DIR**' . And finally, after generation, we will not get this:

```
        'moduleDir' => __DIR__ . '/../module'
```

generated value will be like it:

```
        'moduleDir' => '__DIR__ . \'/../module\''
```

Changes:
Add ability to define environment constants by method

```
public function initEnvironmentConstants();
```

Also add ability to define custom constants by methods

```
public function addConstant($constant);
public function deleteConstant($constant)
```

property `$constants` must be type of `ArrayObject` for memory saving when pass it as argument for recursive generation by reference.

### DIFF
--- a/library/Zend/Code/Generator/ValueGenerator.php
+++ b/library/Zend/Code/Generator/ValueGenerator.php
@@ -9,6 +9,8 @@
 
 namespace Zend\Code\Generator;
 
+use Zend\Stdlib\ArrayObject;
+
 class ValueGenerator extends AbstractGenerator
 {
     /**#@+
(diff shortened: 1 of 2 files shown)


library/Zend/Code/Generator/ValueGenerator.php
```diff
@@ -239,7 +318,7 @@ public function generate()
                 );
                 foreach ($rii as $curKey => $curValue) {
                     if (!$curValue instanceof ValueGenerator) {
-                        $curValue = new self($curValue);
+                        $curValue = new self($curValue, self::TYPE_AUTO, self::OUTPUT_MULTIPLE_LINE, $this->getConstants());
```
Maks3w: @3axap4eHko Did you use ArrayObject for this line?

I don't understand the performance improvements you claim. Non return by reference methods implemented on StdLib\ArrayObject are invoked.

The only thing I see the constants are shared by the instances created in this line, so the only need I see $this->constants must be provided as reference. Alternative solutions may

* Use PHP native `ArrayObject`
* Refactor getConstants for return by reference.

3axap4eHko: Constants is a very big array, containing about 1k string values. Let's length of every string length is 10, therefore total  size of array in memory would be greater then 10KB for one instance of ValueGenerator, that means for 1 property of generated class. So for 100 classes with 100 properties total size of constant arrays would be 10MB instead of 10KB for reference value

3axap4eHko: Also you should keep in mind of allocating memory for every array in method. That 100 classes with 100 properties would be have a 10000 arrays

Maks3w: So native `ArrayObject` is suitable for this case and `Stdlib` version is not required, isn't it?

3axap4eHko: Stdlib\ArrayObject is a wrapper for backward version capability and nothing more

PR #4726, opened 2013-06-26
Add maxLength to ParameterContainer

postalservice14:
The Oci8/Statement class currently sets the max length parameter of oci_bind_by_name to -1.  I need to be able to manually set this parameter.

This is my approach to doing this.


weierophinney:
Merged to develop for release with 2.4.0

PR #4774, opened 2013-07-02
[ZF3] copy resources/languages/lang/Zend_Validate.php to resources/languages/lang/Zend_Validator.php

samsonasik:
I keep resources/languages/lang/Zend_Validate.php to keep no BC. it is can be a preparation for ZF3 (for consistency of naming with component name) , which resources/languages/lang/<b>Zend_Validate.php</b> can be completely removed and changed with resources/languages/lang/<b>Zend_Validator.php</b>.


samsonasik:
moved to https://github.com/zendframework/zend-i18n-resources/pull/1

PR #4962, opened 2013-08-14
added "ControllerManager" Manager, and make "ControllerLoader" as alias of it

samsonasik:
for consistency, "ControllerLoader" should be named "ControllerManager". It not BC break because of I made new Factory named ControllerManagerFactory and point ControllerLoader as alias of ControllerManager. It can be a preparation for ZF3. When ZF3 released, the "ControllerLoader" manager can be completely removed.


samsonasik:
re-created in https://github.com/zendframework/zend-mvc/pull/6

PR #5080, opened 2013-09-05
Added separator to model in renderPartial function

serpent7776:


froschdesign:
:-1: 
This is not needed!

Use in your partial script:
```php
$this->navigation()->breadcrumbs()->getSeparator();
```

PR #5229, opened 2013-10-05
[ZF3] Callback handler for PHP 5.4

bakura10:
As ZF3 will be PHP 5.4 only, we can optimize this class quite heavily.

All tests pass (of course Travis fails because of PHP 5.3).


Maks3w:
This class has been removed on V3

PR #5230, opened 2013-10-05
[ZF3] Clean Stdlib

bakura10:
- Remove fluent interface for options class
- Remove the Hydrator classes from Stdlib as it will have its own component (https://github.com/zendframework/zf2/pull/5000)
- Remove deprecated classes
- Updated tests
- Remove InitializableInterface because this interface was introduced for doing bad things (the infamous init method).
- CallbackHandler class is a bigger refactoring (see [this PR](https://github.com/zendframework/zf2/pull/5229)).

Some notes while checking the code:
- I don't remember the status about ArrayObject. I thought we wanted to throw it out no?

Don't hesitate to quickly check this namespace. It's pretty quick and most classes are small.


tests/ZendTest/Stdlib/OptionsTest.php
```diff
@@ -88,10 +88,11 @@ public function testGetThrowsBadMethodCallException()
 
     public function testSetFromArrayAcceptsArray()
     {
-        $array = array('test_field' => 3);
+        $array   = array('test_field' => 3);
         $options = new TestOptions();
+        $options->setFromArray($array);
 
-        $this->assertSame($options, $options->setFromArray($array));
+        $this->assertSame($options, $options);
```
Maks3w: Is this a joke? :)

Maks3w:
All of this is done stdlib or there are PRs proposing the same except removal of Fluent interfaces.

PR #5231, opened 2013-10-05
[ZF3] Rbac component cleaning

bakura10:
ping @spiffyjr

Another PR for ZF3 as I have lot of time today :) !

This one could have been included in ZF2 but I think it would make a very very small BC so here it is. This component is pretty small and has nearly not evolved in ZF2 as it is nearly perfect (nice work @spiffyjr!):
- Rbac class no longer extends AbstractIterator. In fact while workign on my InputFilter refacotr, I realized that implementing IteratorAggregate is MUCH faster than implementing all the recursive iterator methods (because the C code must call a lot of userland code, while when returning a RecursiveArrayIterator, all is done in C). Furthermore, it does not make a lot of sense for Rbac component to implement RecursiveIterator (in fact, the roles are iterated).
- Roles no longer implement RecursiveIteraotr, but only the simpler IteratorAggregate that returns a RecursiveArrayIterator.
- Variable "children" in rbac was renamed to "roles" to better inforce what it does.

Side note: it's not useful to create custom RecursiveIterator implementation unless you have specific and custom needs for the getChildren class.

All tests pass.

NOTE to spiffyjr: I checked your implementation, and the "setParent" method in the Role only accept a RoleInterface object, while the addChild can accept a string or a role (and create a new Role instance if it is a string). I think we should make this more coherent by either accepting string in setParent OR typehint RoleInterface for addChild. But as it is today it's a bit confusing.

Also: I don't like the fact that getRole method in Rbac throws an exception if it cannot found a role. Would not it make more sense to the getRole method returns "null" if no role can be found?

EDIT: also: shouldn't the AssertionInterface have the role and the permission in the assert method too?


bakura10:
I suppose ZF3 won't have anymore RBAC compoennt. ZfcRbac uses [Zfr-Rbac](https://github.com/zf-fr/rbac) as component.

PR #5386, opened 2013-10-30
[Cache] Add adhoc TTL support, fix per-item ttl handling.

Thinkscape:
### Why can't I specify TTL ?

Since I remember working with code-level cache solutions there was always an easy method to specify TTL on per-item basis. It is a [common](http://us3.php.net/apc_store) [feature](http://us3.php.net/manual/en/function.wincache-ucache-add.php) [among](http://www.php.net/manual/en/memcache.add.php) most well-known libraries and extensions and was present in [ZF1 Zend_Cache](https://github.com/zendframework/zf1/blob/master/library/Zend/Cache/Backend/File.php#L243). Sadly, zf2 is missing this.

``` php
// APC
apc_store ( string $key , mixed $var [, int $ttl = 0 ] );

// Memcache
Memcache::add( string $key , mixed $var [, int $flag [, int $expire ]] );

// Wincache
wincache_ucache_add ( string $key , mixed $value [, int $ttl = 0 ] );

// Zend Server
zend_shm_cache_store($internalKey, $value, $ttl);

// Zend_Cache
Zend_Cache_Backend_File::save($data, $id, $tags = array(), $specificLifetime = false);

// etc.
```
### Problem 1

ZF2 Cache Storage does not allow to store items with specific TTL.

The first problem was missing `$ttl` parameter in all [Cache\StorageInterface](https://github.com/zendframework/zf2/blob/master/library/Zend/Cache/Storage/StorageInterface.php#L99) write methods (i.e. `setItem()`, `replaceItem()` etc.). Without this, the only way to set an item with a specific TTL was:

``` php
/* @var $cache StorageInterface */
$oldTtl = $cache->getOptions()->getTtl();
$cache->getOptions()->setTtl(60); // set to 1 minute
$cache->setItem('foo','bar'); // store the item
$cache->getOptions()->setTtl($oldTtl); // restore previous ttl
// *phew*
```
### Problem 2

Unfortunately, the above code does not work properly. This is because many ZF2 adapters do not support or understand item-level TTL, unless it is handled internally by the backend (i.e. `apc_store()` natively supports TTL and it's getting fed with the one from options).

For example,  `Storage\Memory` and `Storage\Filesystem` would exhibit unexpected behavior when discarding old items or fetching items if we tried to manipulate ttl:

``` php
/* @var $cache StorageInterface */
$cache = new \Zend\Cache\Storage\Adapter\Memory();
$cache->getOptions()->setTtl(3600); // set TTL to 1 hour
$cache->setItem('foo',1); // store an item

$cache->getOptions()->setTtl(60); // set TTL to 1 minute
$cache->setItem('bar',1); // store another item

sleep(61);

var_dump($cache->hasItem('bar'));  // false
var_dump($cache->hasItem('foo'));  // false
// *poof*   "foo" item disappeared but should live for an hour
```
### The goal

``` php
/* @var $cache StorageInterface */
$cache->setItem('foo','1', 60); // store the item for 60s
$cache->setItem('bar','2', 3600); // store the item for 1 hour

// store the item with default TTL setting from the storage
$cache->setItem('bar','2'); 

// replace multiple items and set their TTL to 60s
$cache->replaceItems(['foo'=>10,'bar'=>20], 60); 

// increment item by 5 and make it live for 1h
$cache->incrementItem('foo', 5, 3600);
```
### Solution
- [x] Add `$ttl` support to StorageInterface.
- [x] Refactor `Cache\Storage\Adapter\AbstractAdapter` to handle $ttl in write operations.
- [x] Pass on `$ttl` to listeners (plugins).
- [x] Refactor all `Adapter\*` classes to use per-item `$ttl` and pass it on to underlying backend.
- [x] Refactor `Adapter\Memory` to support per-item TTL.
- [x] Refactor `Adapter\Filesystem` to support per-item TTL.
- [x] Add TTL tests that work with ad-hoc ttl setting.
- [x] Add overlapping temporal per-item ttl tests.

### DIFF
--- a/library/Zend/Cache/Storage/Adapter/AbstractAdapter.php
+++ b/library/Zend/Cache/Storage/Adapter/AbstractAdapter.php
@@ -35,6 +35,7 @@ abstract class AbstractAdapter implements StorageInterface, EventsCapableInterfa
 
     /**
      * Event handles of this adapter
+     *
      * @var array
      */
     protected $eventHandles = array();
@@ -127,6 +128,7 @@ abstract class AbstractAdapter implements StorageInterface, EventsCapableInterfa
             $event = new Event('option', $this, new ArrayObject($options->toArray()));
             $this->getEventManager()->trigger($event);
         }
+
         return $this;
     }
 
@@ -141,6 +143,7 @@ abstract class AbstractAdapter implements StorageInterface, EventsCapableInterfa
         if (!$this->options) {
             $this->setOptions(new AdapterOptions());
         }
+
         return $this->options;
     }
 
@@ -156,10 +159,11 @@ abstract class AbstractAdapter implements StorageInterface, EventsCapableInterfa
      */
     public function setCaching($flag)
     {
-        $flag    = (bool) $flag;
+        $flag = (bool)$flag;
         $options = $this->getOptions();
         $options->setWritable($flag);
         $options->setReadable($flag);
+
         return $this;
     }
 
@@ -175,6 +179,7 @@ abstract class AbstractAdapter implements StorageInterface, EventsCapableInterfa
     public function getCaching()
     {
         $options = $this->getOptions();
+
         return ($options->getWritable() && $options->getReadable());
     }
 
@@ -190,13 +195,14 @@ abstract class AbstractAdapter implements StorageInterface, EventsCapableInterfa
         if ($this->events === null) {
             $this->events = new EventManager(array(__CLASS__, get_class($this)));
         }
+
         return $this->events;
     }
 
     /**
      * Trigger an pre event and return the event response collection
      *
-     * @param  string $eventName
+     * @param  string      $eventName
      * @param  ArrayObject $args
      * @return \Zend\EventManager\ResponseCollection All handler return values
      */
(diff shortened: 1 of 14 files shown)


elmosgot:
I haven't read the complete thread, but at least the mayor part of the discussion. I just wanted to shine a light on the adapters that don't support natively ttl.

I find it hard to believe that there is no way to simply implement a fast way of detection if the content has been expired.

For example the filesystem adapter we can't just add the ttl to the end of the key?
`$this->storage->setItem($key, $value, $ttl);`

That this results on the filesystem as:
`/tmp/$key-$ttl`

Or another idea that I got is that the key is a folder that uses the filenames for showing the metadata of the cache:
```/tmp/$key/1435792859
/tmp/$key/value```

I just wanted to put my ideas away, maybe this could resolve the issues with non-native-ttl-suporting-addapters. So we can have this feature available for all adapters without loosing performance in non-native-ttl-suporting-adapters.

marc-mabe:
@elmosgot The filesystem adapter already supports TTL but it will be calculated with `mtime + TTL <= time()`. That is the most performent way but as a consequence the TTL isn't stored within the cached item and but it will be used from current configuration on reading the item.

elmosgot:
@marc-mabe I know that you can create a generic TTL per storage adapter, but the thread is about implementing a fast solution for item specific TTL. One workaround is to have multiple storage adapters with their own TTL specified and use them where needed. But this is a more complex solution.

My ideas where to use fast ways to validate the cached data without actual accessing the file until we know if the cache is valid.

PR #5399, opened 2013-11-01
[ZF3] Clean event manager

bakura10:
Hi everyone,

This is a preliminary work for a clean-up EVM for ZF3. Event manager was pretty well done so this is mostly cleanup, optimizations, merging some interfaces, removing deprecated code.

I'd like your feedback on some choices I've made.
- The StaticEventManager was removed as it was deprecated just after ZF2 release in favour of SharedEventManager.
- The GlobalEventManager has been removed (I'm not sure about it, this class looks a bit like a StaticEventManager but has not been marked as deprecated... If needed I can make it back).
- A lot of PHPDoc fixing, and some code were updated to respect interface contract.
- Methods have been reordered to group more logically methods.
- Use callable type hint where possible.
- unsetSharedEventManager has been removed in favor of setSharedEventManager accepting null. The problem with unsetSharedEventManager is that it set it as "false" (and was considering it different from null in some tests, making it hard to follow).
- propagationIsStopped has been renamed to isPropagationStopped (and all the code that rely on it has been updated).
- trigger and triggerUntil are complicated methods. They accept a lot of parameters and the behavior changes depending on the order parameters are passed. To add to the complexity, trigger accepted a last parameter to emulate triggerUntil. To simplify the signature, no callback can be added to trigger now. You should use triggerUntil to that. It makes the code more explicit and makes the implementation simpler.
- For the same idea, the "attach" method no longer accepts a ListenerAggregate. You must use the attachAggregate method for that. This allow to avoid to test instance of against ListenerAggregate whenever we attach something. We can now also type hint for callable in the attach method which avoids another test.

All tests have been updated.

TODO:
- [x] Backport the "Filter" feature (is this needed or was it something from ZF2 that was never used? I can't find any reference about it in the doc, and it's not used anywhere in the framework). EDIT: after some discussion with @weierophinney , he told me it was something that was initially planed to be used inside the framework, but finally failed to what it was done. So I'll don't back port it until someone tell me to do so.
- [ ] Write proper documentation for event manager component
- [ ] Feedback ?


bakura10:
Replaced by #7145

Issue #5474, opened 2013-11-13
[ZF3] array() or []


bakura10:
Let the war begin.

[] : a lot of folks
array : 0

PR #5480, opened 2013-11-14
[ZF3] implement #5474 : change array() into []

samsonasik:
implement #5474


samsonasik:
rebased

Maks3w:
This will do when needed using the CS tool instead manual change or any other script.

PR #5500, opened 2013-11-19
[ZF3] [WIP] Service Manager Cleanup

Ocramius:
Here's my made-up roadmap for ZF3's ServiceManager:
- [x] Removal of aliases
- [x] Compatibility layer for ZF2 ServiceManager aliases (for those who really really need them ?!)
- [x] Removal of canonical names
- [x] Compatibility layer for ZF2 ServiceManager canonical names
- [x] Removal of `retrieveFromPeeringManagerFirst` support (no polyfill/not used in the framework itself)
- [x] Removal of peering service managers as a native functionality
- [x] Compatibility layer for ZF2 ServiceManager peering service managers
- [x] Change of interfaces to remove the `$cName` and `$rName` differentiation. Now `$rName === $cName`
- [x] Implement passing service options to the locator during lookups (currently completely broken, and broken also in mutable options interface)
- [ ] Test cases for the introduced compatibility layer
- [ ] Performance tests
- [ ] Code cleanup (currently just a PoC)
- [ ] Tests green on components depending on Zend\ServiceManager

See the changes in the tests for a good overview of what I broke.


Maks3w:
ServiceManager v3 development is done in SM repository

Martin-P:
> ServiceManager v3 development is done in SM repository

Can you provide the link to the PR which handles the issues described in this PR? Thanks.

fhein:
To all who disregard BC: 

What is the value to author a framework which will not get market acceptance? If you hit the users, they will hit you more. You can only scratch them, they can kill you. 

Switching a version number, even if it is a major one, is not enough justification to let any single user down. 

And for what reasons would you scratch them? Code gets cleaner? What do you mean? Who cares? But in particular: Do you mean, cleaner means classes have less member functions? If it is only that, buying a bigger monitor would be more appropriate than dropping functionality. Just an example.

My perception currently is, that Zend Framework is a sinking ship. One important reason is, that BC breaks are introduced, just because someone LIKES somewhat better. Hey, who cares what you like?
Open source. as any other business, is a business. If you don't have a business case for what you like, forget it or like it at home.

Please keep in mind, that businesses learn to rely on open source. Please consider, that any change requires people to define a business case to justify why to follow that change. Please deliver enough and sufficient information for those people. 

It may be ugly, but it works ok and we achieve our goals, it's perfect. That is business.

OT: I do not like this kind of monster PRs.

Ocramius:
ZF is a tool, not a product. Don't get too attached to the tool.

You are worried about market share when you are selling a product, otherwise this is "just" OSS work, and I personally only do OSS work when it is for making something better, not necessarily popular. I also say "just", but it is my passion, pride and primary source of self-improvement.

So yes, my personal work on free and open source software will keep going this way: if a BC break improves the quality, performance and reliability (reduced defect ratio), then it will be worth the effort.

If you disagree with my work, that's fine, but I really only do it to improve the final result, do not assume malice.

fhein:
@Ocramius: I appreciate your work and deeply respect your knowledge and problem solving strategies. You are also an excellent teacher. I follow your publications for a long time. I got lots of insights regarding PHP by studying your work. I come from c++ and it was your work and the work of very few others which provided me a fast track to learn about the more interesting parts of PHP.

I apologize for the tone of my statement. Don't take that personally, please. I was on my tiny pull request for the service manager. I fixed a bug. Your immediate reaction: Why not kick out the whole feature. Not nice. Then I was asked to split an ultra-tiny change into pieces. And please remember the detail of your review (redundant $config, really?). Then I came here, found this mammoth PR which currently affects 106 files or so and the task list still has open points. - This just to explain why I was not in the best mood.
Again, sorry. 

A framework - open source or not - is a platform people are invited to build upon. Building on third party requires an investment.  To invest is a matter of trust. Trust is a matter of usability, stability, reliability, sustainability, supportability, performance. Very similar to product quality requirements.
People who use ZF did invest. IMHO saving this invest and - in the best case - paying interest on that invest is important and the key to acceptance and success. A framework that nobody uses has no value, even if it is pretty well designed (because nobody is there to benefit from the excellent design).

Maybe this is a PHP specific thing, that BC is considered to be not very important. In c++ boost for
example, this is seen and handled completely different. boost libraries are developed and maintained with BC as an ultra-important aspect (because they object to break customer code). When libraries get old and concepts do not hold any longer, they are replaced by new components and new concepts.

I know you are involved in BetterReflection as an example. Why not introduce a BetterServices approach, separate from the core ZF? This could kick everything, could rethink DI for example, could rethink Symfony and Zend services handling and ... IMHO, this would be cleaner, would promote your approach to make things better and would guard the customers. And even people who fix a bug in the core.

weierophinney:
@fhein This mammoth PR....

- was never merged.
- is completely obsolete regardless, due to other changes made in zend-servicemanager for the v3 release.

This particular repository is currently only a _metapackage_; it has no code. This change happened a number of years ago, when we released version 2.5. At this time, all code is in separate repositories for each component.

Why?

By doing this, we allow the components to each release at their own pace. For some, this meant major version bumps; for others, we've only had bugfix or feature releases.

One thing to note is that _semantic versioning allows for backwards compatibility breaks when bumping the major version_. So, while I understand your arguments about BC, I would also argue that they are not relevant when we are talking about major version changes. In these specific cases, BC breaks are definitely in scope, even if they may require effort on the part of users when upgrading.

Major version bumps are not done lightly. They are executed when we have one or more of the following:

- Major bugfixes or security fixes that require a BC break to accommodate.
- Performance improvements that require a BC break to accommodate.
- Adoption of new language features, which require a BC break to accommodate (e.g., adopting `callable` or `iterable` typehints to simplify internals; adopting the splat operator to simplify internals; adopting scalar and/or return type hints to simplify internals; etc).
- Refactoring that will simplify maintenance.

While this particular patch was not applied, we did do _similar_ patches for a zend-servicemanager v3 release. Our reasons spanned each of the above: we were able to make changes that simplify maintenance by adopting new language features, to address long-standing bugs that we'd been unable to fix previously, and to get a huge performance bump.

New _major_ versions are exactly the place to make BC breaks. If you are using Composer, you should be pinning to a specific minor or major release version (e.g., `~2.7.3` or `^2.7`); adopting a new major version requires you to _opt-in_ by updating your constraints. This is exactly the idea you posit above of "replaced by new components and new concepts"; the difference is selecting a new major version.

Further, we are putting a lot of effort into making such migrations simpler. One change we have made to our processes since version 2 was first released is that we do not allow major version bumps _without migration documentation and/or tooling_. Ideally, we also provide a new _minor_ version with features that allow you to make your code _forwards compatible_ with the new major release. This also allows us to continue supporting previous major versions with critical and/or security fixes so that users are not completely left behind.

fhein:
@weierophinney : Thanks for the clarification. I understand the purpose of this PR better now. As said, I apologize for the tone of my first mail here. I tried to explain how it came to this. I hope you comprehend.

The origin of this discussion was an attempt by me to fix a bug with aliases in zend-servicemanager. This went to 'why not throw out the whole feature'. 

That made me think. Aliases are de facto a special case of a delegator. These were just introduced. You can rebuild the alias functionality completely with delegators, including the 'major pain in the butt' aspects @Ocramius mentioned. 

For me if delegators are good (and they are IMO), aliases can't be bad and evil. They are a trivial delegators which are internally handled (due to the triviality of it's operation and for the sake of performance).

Aliases provide a name mapping only, so putting them into a factory (either a new internal type or an abstract factory) would introduce significant overhead without benefit. I do not see it, at least.

As of the current zend-servicemanager master, aliases cost performance on any request. With PRs #215 and #216 I proofed that this cost can easily get reduced to almost nothing (in particular one single `empty($this->resolvedAliases)`, and that the number of boolean expression evaluations can be reduced. 

I understand and agree that in a major release BC can be broken. But IMHO this has to be considered very carefully. I read all the contributions to this thread and I got the impression that not everybody would subscribe to that. Some seem to argue as if a major release change was justification enough to break each and any API. Just my impression. I hope it's wrong.

All that said, for me aliases are not even a candidate for BC break discussions.

fhein:
Btw, I agree with @Ocramius, that [the name 'Delegator' is somewhat misleading.](https://ocramius.github.io/blog/zend-framework-2-delegator-factories-explained/)

PR #5520, opened 2013-11-21
[ZF3] Paginator refactor

snapshotpl:
I think paginator needs changes. First of all I have been removed view dependency. We can also remove most static functions and properties.

And my questions:
I don't know whether we need Paginator\AdapterAggregateInterface ?
What are you think to bring scrolling style to view helper?


GeeH:
New issue can be found at: https://github.com/zendframework/zend-paginator/issues/23

PR #5614, opened 2013-12-14
Fixed issue with Math\Rand::getInteger() on ranges close to PHP_INT_MAX

denixport:
When min - max range is close or equal to `PHP_INT_MAX`  `log()` produces number of bits
equal to machine size integer. So `$filter = (int) ((1 << $bits) - 1);` evaluates to zero.
In this case `getInteger(0, PHP_INT_MAX)` returns only zeros.

PR includes correct bit calculation and test

### DIFF
--- a/library/Zend/Math/Rand.php
+++ b/library/Zend/Math/Rand.php
@@ -143,7 +143,7 @@ abstract class Rand
         $filter = (int) ((1 << $bits) - 1);
 
         do {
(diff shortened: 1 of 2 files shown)


Maks3w:
Is it normal `$values` to be negative?

Maks3w:
Ok, I see what is happenning. $values probably has a value which produce a int overflow as result of operate with big numbers.

PR #6073, opened 2014-04-01
Reduced code duplication in FlashMessenger plugin

fp-pszczekutowicz:
- Zend/Mvc/Controller/Plugin/FlashMessenger.php


ThaDafinser:
@pszczekutowicz thanks for the change
@Ocramius thank you for merging it not before 2.4.0 -> i exactly run into the `BC break with subclasses`

PR #6108, opened 2014-04-08
Zend\Test Enable trace error by default

blanchonvincent:
I wanted to change this for a long time but I Had to wait for https://github.com/zendframework/zf2/pull/6041, but now it's ok, we can enable trace error by default, it will be better for developer.

It's not a BC because if test is broken now, it's because the developer had an exception he didn't see, so it's better :)


sylvainL:
Hey
I Just migrate from 2.3.7 to 2.4.0. My test passed before and fails now. Yes an exception is raised and this is the expected behavior for my test. 

The exception is raised and then handled by my (Zend) app during the dispatch event. The controller returns an HTTP 500 with a proper template (error/index).

So When I try to handle this with a "$this->setExpectedException(...)", it does not work because no exception is raised in test, but the trace crashes my test.

I don't want to set traceError to false everywhere, so what is the way to handle that ?

Thxs

PR #6154, opened 2014-04-17
Properly pass context to input filter

bakura10:
Hi,

I was quite surprised to see this bug. Basically, the "isValid" method of input filter did not allow to pass any context, which in hence could not be transfered to various validators, that made it impossible to use some validators that relied on context.

The problem is that input filter actually considered the whole array data as context, and pass it to various inputs.

To avoid any BC, I just checked if a context was explicitly given in the input filter. If that's the case, this context is preferred over the data as array.

Thanks!


svycka:
@bakura10 maybe similar thing should be added into forms ant fieldsets?

PR #6173, opened 2014-04-23
JOIN empty columns

buliq:
If I make a join and don't want any columns to be fetched I should be able to pass null or empty string. Now if You do this, query will look like You tried to get an column, without its name.

$select->from("B");
$select->join("A", "A.A_ID = B.B_A_ID", null);

SELECT "B".*, "A". AS "" FROM "B" INNER JOIN "A" ON "A"."A_ID" = "B"."B_A_ID"

### DIFF
--- a/library/Zend/Db/Sql/Select.php
+++ b/library/Zend/Db/Sql/Select.php
@@ -258,7 +258,10 @@ class Select extends AbstractSql implements SqlInterface, PreparableSqlInterface
                 sprintf("join() expects '%s' as an array is a single element associative array", array_shift($name))
             );
         }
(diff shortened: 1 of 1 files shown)


fernandomantoan:
I was able to reproduce this issue, but I used another code to solve it:

```php
$select->from("B");
$select->join("A", "A.A_ID = B.B_A_ID", array());
```

```sql
SELECT "B".* FROM "B" INNER JOIN "A" ON "A"."A_ID" = "B"."B_A_ID"
```

This will create the desired SQL output, maybe the method could also accept the third parameter to be `null`, I could create a PR with a test-case if needed.

PR #6239, opened 2014-05-07
ModuleManager load module childs dependents

turrsis:
Allow load modules (from `Module::init()`) after current module.
Analog of https://github.com/zendframework/zf2/pull/5651, but:
   load modules `A, B` - configs order is `A, B`.
   load modules `A -> B` - configs order is `A, B`.

### DIFF
--- a/library/Zend/ModuleManager/ModuleManager.php
+++ b/library/Zend/ModuleManager/ModuleManager.php
@@ -61,7 +61,7 @@ class ModuleManager implements ModuleManagerInterface
     /**
      * @var ModuleEvent
      */
-    protected $loadEvent = null;
+    protected $loadEvent;
 
     /**
      * True if modules have already been loaded
@@ -120,9 +120,11 @@ class ModuleManager implements ModuleManagerInterface
     public function onLoadChildsModules(ModuleEvent $e)
     {
         $parentModule = $this->loadEvent->getModuleName();
+
         if (!isset($this->childsModules[$parentModule])) {
             return;
         }
+
         foreach($this->childsModules[$parentModule] as $module) {
             $this->loadModule($module);
         }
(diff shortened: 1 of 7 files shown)


weierophinney:
Not yet reviewed by @EvanDotPro , but it _does_ need to be rebased if we do decide to merge.

turrsis:
@weierophinney rebased.

weierophinney:
@EvanDotPro Can you please review this and at the very least comment on whether you think we should merge or not?

library/Zend/ModuleManager/ModuleManager.php
```diff
@@ -52,6 +52,18 @@ class ModuleManager implements ModuleManagerInterface
...
+     * Childs dependency modules
+     *
+     * @var array
+     */
+    protected $childsModules = array();
+
+    /**
+     * @var ModuleEvent
+     */
+    protected $loadEvent;
```
EvanDotPro: I don't quite understand why we need another copy of the ModuleEvent... Is it not already set in `$this->event` ?

EvanDotPro:
@turrsis So, this might be because I haven't taken enough time to understand the approach taken, but I just don't quite understand a couple things about this approach. First, it seems to hold an extra copy or reference to the ModuleEvent? Second, could you explain the changes around the loadFinished property? Overall, I understand and appreciate what you're trying to do (a delayed module-load), I'm just hung up on the implementation details at this point. /cc @weierophinney @Ocramius

weierophinney:
I'm going to delay this, then; I think we need more discussion of the feature.

EvanDotPro:
Sounds good. Sorry for the delay, @turrsis!

turrsis:
copy of the ``ModuleEvent`` : I don't understand this too :) , fixed.
``loadFinished `` : if you increment and decrement this variable - it should be integer

GeeH:
New issue can be found at: https://github.com/zendframework/zend-modulemanager/issues/47

PR #6267, opened 2014-05-13
Support commas in names

davidwindell:
Bad things happen when an email header has a comma in the name, this fixes that.


weierophinney:
I'm assigning to 2.4. While we likely need a more robust solution, we simply need _any_ solution at this time.

weierophinney:
@davidwindell I just attempted to merge this, but have a new failing test: `ZendTest\Mail\Storage\MessageTest::testGetDecodedHeader` now fails, asserting that:

```
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'"Peter Müller" <peter-mueller@example.com>'
+'Peter Müller <peter-mueller@example.com>'
```

Could you please investigate?

davidwindell:
@weierophinney Are you happy for test to be modified to remove the speech marks around the name? That's what str_getcsv will do and was what prompted Dasprid's concerns with the PR.

weierophinney:
@davidwindell I think that would be fine; they're an optional part of the standard, and do not affect usability within messages in any meaningful way.

davidwindell:
I agree, I have updated the test so this should be ready to merge.

weierophinney:

PR #6271, opened 2014-05-13
HTML5 compliant form attributes

stldo:
According to W3C specification, ‘action’ is an optional attribute and can’t be empty in HTML5. So, if the form doctype is ‘HTML5’ or ‘XHTML5’, the ‘action’ attribute isn’t set as an empty string by default. Also, if either doctypes are detected, form ‘method’ attribute is not set to ‘get’ by default, as some applications may implement submit buttons that use ‘formmethod’ attribute, making form ‘method’ redundant or useless (according to W3C, “missing value default for the method attribute is also the GET state” — and have been since HTML 2.0).

Reference:
- http://www.w3.org/TR/html5/forms.html#attr-fs-action
- http://www.w3.org/TR/html5/forms.html#attr-fs-method

### DIFF
--- a/library/Zend/Form/View/Helper/Form.php
+++ b/library/Zend/Form/View/Helper/Form.php
@@ -81,10 +81,16 @@ class Form extends AbstractHelper
      */
     public function openTag(FormInterface $form = null)
     {
-        $attributes = array(
-            'action' => '',
-            'method' => 'get',
-        );
+        $doctype = $this->getDoctype();
+
+        if ('HTML5' !== $doctype && 'XHTML5' !== $doctype) {
+            $attributes = array(
+                'action' => '',
+                'method' => 'get',
+            );
+        } else {
+            $attributes = array();
+        }
 
         if ($form instanceof FormInterface) {
             $formAttributes = $form->getAttributes();
@@ -94,7 +100,11 @@ class Form extends AbstractHelper
             $attributes = array_merge($attributes, $formAttributes);
         }
 
-        $tag = sprintf('<form %s>', $this->createAttributesString($attributes));
+        if ($attributes) {
+            $tag = sprintf('<form %s>', $this->createAttributesString($attributes));
+        } else {
+            $tag = '<form>';
+        }
 
         return $tag;
     }
--- a/library/Zend/Form/View/Helper/Form.php
+++ b/library/Zend/Form/View/Helper/Form.php
@@ -11,6 +11,7 @@ namespace Zend\Form\View\Helper;
 
 use Zend\Form\FieldsetInterface;
 use Zend\Form\FormInterface;
+use Zend\View\Helper\Doctype;
 
 /**
  * View helper for rendering Form objects
@@ -83,7 +84,7 @@ class Form extends AbstractHelper
     {
         $doctype = $this->getDoctype();
 
-        if ('HTML5' !== $doctype && 'XHTML5' !== $doctype) {
+        if (Doctype::HTML5 !== $doctype && Doctype::XHTML5 !== $doctype) {
             $attributes = array(
                 'action' => '',
                 'method' => 'get',
--- a/tests/ZendTest/Form/View/Helper/FormTest.php
+++ b/tests/ZendTest/Form/View/Helper/FormTest.php
@@ -15,6 +15,7 @@ use ZendTest\Form\TestAsset\CityFieldset;
 
 use Zend\Form\Form;
 use Zend\Form\View\Helper\Form as FormHelper;
+use Zend\View\Helper\Doctype;
 
 class FormTest extends CommonTestCase
 {
@@ -112,11 +113,11 @@ class FormTest extends CommonTestCase
     {
         $helper = new FormHelper();
 
-        $helper->setDoctype('HTML4_LOOSE');
+        $helper->setDoctype(Doctype::HTML4_LOOSE);
         $html4Markup = $helper->openTag();
-        $helper->setDoctype('HTML5');
+        $helper->setDoctype(Doctype::HTML5);
         $html5Markup = $helper->openTag();
-        $helper->setDoctype('XHTML5');
+        $helper->setDoctype(Doctype::XHTML5);
         $xhtml5Markup = $helper->openTag();
 
         $this->assertContains('action=""', $html4Markup);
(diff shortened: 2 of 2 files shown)


fabiocarneiro:
@Ocramius This was misinterpreted i guess. It would be good to revert.

http://www.w3.org/TR/html5/forms.html#form-submission-algorithm

> If action is the __empty string__, let action be the document's address of the form document.

Angular.js has a behavior that calls preventDefault on the forms that doesn't have an action specified. It doesnt do anything if the action is empty.

https://github.com/angular/angular.js/pull/3776

My best guess is that an empty string is not considered a __specification__ of a value, because it says , if specified, it only accepts VALID NON-EMPTY URL (but a empty string would be ok if not considered a specification).

This was working perfectly fine in zf 2.3.x. and now i must specify a url in all the actions.

Martin-P:
> because it says , if specified, it only accepts VALID NON-EMPTY URL (but a empty string would be ok if not considered a specification).

Perhaps you can describe the difference between an empty url and an empty string (with examples)?

fabiocarneiro:
@Martin-P I'm not saying these are different things. I'm saying the empty string might not be considered a url specification.

Martin-P:
The specification says:
> The action and formaction content attributes, if specified, **must have** a value that is a valid non-empty URL potentially surrounded by spaces.

If an empty url and an empty string are considered the same thing, this means `action=""` is not allowed in HTML5.

What you are referring to is what HTML5 *should do* when a particular situation is encountered, it says nothing about the fact if it is allowed or not. The complaint of previous versions of HTML was that browsers made up their own rules, because the specification did not specify what action should be taken in certain (unexpected) situations. One of the goals of the HTML5 specification is to describe what *should* be done in such a situation to prevent browsers making up their own rules ones again.

fabiocarneiro:
@Martin-P You're telling me that the specification is telling what should be done by browsers if a not valid html5 declaration is found and this behavior is to threat the issue like it wasn't there by assuming the document address is the required action?

If this document is telling browsers SHOULD use the document address when they find a EMPTY STRING, the this is a valid setting, otherwise the recommended behavior was going to be DO NOTHING BECAUSE THIS IS NOT VALID.

Martin-P:
> If this document is telling browsers SHOULD use the document address when they find a EMPTY STRING, the this is a valid setting

The document describes what procedure should be followed. If that procedure describes what should be done with certain values, this does not mean that value is valid.

An example to clarify this:
When I walk in the middle of the highway and the procedure for car drivers is to avoid me, this does not mean I am supposed to walk in the middle of the highway :wink:

fabiocarneiro:
A better example would be:
You're not allowed to kill people, but if you do, remember to take a jigsaw with you and cut their members so it can fit in a bag and you can bury the corpse, so nobody can see what you're doing.

It makes no sense to allow a behavior that MUST no happen.

Martin-P:
My point is: when a situation is described it does not automatically mean that the situation is valid.

> It makes no sense to allow a behavior that MUST no happen.

It does not allow a behavior, it merely describes what should happen when such a situation is encountered. Please see: [1.10 Conformance requirements for authors](http://www.w3.org/TR/html5/introduction.html#conformance-requirements-for-authors)

> Unlike previous versions of the HTML specification, this specification defines in some detail the required processing for invalid documents as well as valid documents.

> However, even though the processing of invalid content is in most cases well-defined, conformance requirements for documents are still important: in practice, interoperability (the situation in which all implementations process particular content in a reliable and identical or equivalent way) is not the only goal of document conformance requirements. This section details some of the more common reasons for still distinguishing between a conforming document and one with errors.

This part clearly states processing for both valid and invalid documents is handled in the specification.

fabiocarneiro:
Ok, i got your point

> When a form element form is submitted from an element submitter (typically a button), optionally with a submitted from submit() method flag set, the user agent must run the following steps:

> 8. If action is the empty string, let action be the document's address of the form document.

Martin-P:
We are discussing an emtpy string as value for the `action` attribute, right? The part you quote says *the user agent must run the following steps*. One of those steps is checking whether the value of the `action` attribute is an empty string. If it is an empty string, act upon it. That's all. Nothing more, nothing less.

fabiocarneiro:
@Martin-P The specs are not even close to be clear about this. This 1.10 section is very clear, but it exemplifies the text you quoted above with problems like invalid markup, boolean attribute values, typos, etc.... Nothing even close to a empty form action.

You are assuming a __empty string__ is a __action specification__, but if i'm not wrong (and i admit i can be), a empty string is not specifying anything, but the absence of a specification, and the rule says ```if specified```. 

If your interpretation of the specs is correct, the interpretation of the folks from Angular.js isn't. 

@IgorMinar It might be of your interest.

Martin-P:
> but it exemplifies the text you quoted above with problems like invalid markup

An example of invalid markup: `action=""`

fabiocarneiro:
@Martin-P Where in the specs it is saying an empty action is invalid markup?

Martin-P:
Here it says an empty action is invalid markup: http://www.w3.org/TR/html5/forms.html#attr-fs-action

> The action and formaction content attributes, if specified, must have a value that is a valid non-empty URL potentially surrounded by spaces.

fabiocarneiro:
Its not saying that. Its saying that anything that is not considered __a valid non-empty URL, potentially surrounded by spaces__, is invalid markup.

But if you don't interpret that empty string is a specification, then it is valid markup.

Can we simply agree with disagreeing and wait for someone's else opinion? I don't see we moving forward with this. You could be less intolerant with other people opinion, quoting the same text over and over again. I guess we both understood the other side point of view.

sbtoledo:
Just test an empty value on W3C validator and it'll point that an action attribute must be non-empty.

fabiocarneiro:
@sbtoledo nice point.

Maks3w:
W3C Validator is just a "linter". Don't forget that

fabiocarneiro:
https://github.com/angular/angular.js/issues/11495

weierophinney:
Can somebody give me a tl;dr summary, please? Does anything need to change currently in ZF2? If so, what?

Martin-P:
@weierophinney:
The question is: is `action=""` allowed in HTML5 or not? Looking at the HTML5 specification this PR is correct the way it was handled and `action=""` is not allowed in HTML5. IMO opinion the real problem is in angular.js which is not following HTML5 standards.However I think @fabiocarneiro and I agree we disagree about that part :wink:

sbtoledo:
Angular extends HTML, it doesn't care—neither have to care—with W3C standards. I suggest that if you need an empty action value, just set an empty string before rendering the form.
@Maks3w I agree, but the W3C validator is also a reliable source for checking standards conformity.

Martin-P:
If you are already relying on javascript, another possibility is to let javascript check the forms and add an empty action value if none present. That way you do not have to change anything about the way ZF2 renders the form.

fabiocarneiro:
@Martin-P Yes, i'm actually doing that right now. But i guess its not necessary to prevent default submit when no action attribute is set, and this is something angular is doing. I created a issue there complaining about this. I agree with you. I'm sorry to be so insistent. zf2 is correct.

Martin-P:
@fabiocarneiro No need to be sorry for insistence, I can be insistent as well :wink: I think it is a good thing to stand behind your standpoint :+1:

PR #6359, opened 2014-06-07
Date validator problem vant validate big unix timestamps

### LINKED ISSUE #6352: Date validator problem vant validate big unix timestamps
trying like this:
```php
$dateValidator = new Date(['format' => 'U']);
$dateValidator->isValid(14018919967); // false
$dateValidator->isValid("14018919967"); // true
```

is this normal?
I am on windows64 and 32 php I think.
problem is here https://github.com/zendframework/zf2/blob/master/library/Zend/Validator/Date.php#L134-L140

becouse:
```php 
$type = gettype(14018919967); // type = double
```

svycka:
fixes https://github.com/zendframework/zf2/issues/6352

### DIFF
--- a/tests/ZendTest/Validator/DateTest.php
+++ b/tests/ZendTest/Validator/DateTest.php
@@ -70,8 +70,10 @@ class DateTest extends \PHPUnit_Framework_TestCase
             // int
             array(0,                         null,              true),
             array(1340677235,                null,              true),
-            // Commenting out, as value appears to vary based on OS
-            // array(999999999999,              null,              true),
+            // 32bit version of php will convert this to double
+            array(999999999999,              null,              true),
+            // double
+            array(12.12,                     null,              false),
             // array
             array(array('2012', '06', '25'), null,              true),
             // 0012-06-25 is a valid date, if you want 2012, use 'y' instead of 'Y'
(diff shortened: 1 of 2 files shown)


weierophinney:
My notes:

- The test was commented out because the contributor is using a 32-bit architecture, and that test failed due to int -> float conversion.
- The referenced issue essentially asks if the framework should act the same on 32- vs 64-bit architectures.
- Our own continuous integration uses 64-bit architectures.

My take is: no, the framework should _not_ handle this detail. The reason is that the value passed as a timestamp to `DateTime` **must** be an integer for reliable results.

This means we cannot pass a string (e.g., for use with bcmath) or GMP instance and expect reliable results. At the same time, we need to test sufficiently large numbers in order to verify everything works as expected.

My take would be to test using `PHP_INT_MAX`, which will vary based on architecture; we can then change expectations based on architecture (using `PHP_INT_SIZE`, which will be 4 on 32-bit and 8 on 64-bit).

tests/ZendTest/Validator/DateTest.php
```diff
@@ -70,8 +70,10 @@ public function datesDataProvider()
             // int
             array(0,                         null,              true),
             array(1340677235,                null,              true),
-            // Commenting out, as value appears to vary based on OS
-            // array(999999999999,              null,              true),
+            // 32bit version of php will convert this to double
```
weierophinney: I read this entirely wrong; this test **un**comments a test set. We're covered, and verified this passes.

weierophinney:

Issue #6406, opened 2014-06-23
Date validator bug


svycka:
Validator ignores format when integer provided as data
```php
$validator = new \Zend\Validator\Date(array('locale' => 'en', 'format' => 'Y-m-d'));
$validator->isValid('123'); // false
$validator->isValid(123); // true
```
Or maybe this is also desired behavior?

here validator ignores everything if data is integer: https://github.com/zendframework/zf2/blob/master/library/Zend/Validator/Date.php#L152-L155

adamlundrigan:
This is the intended behaviour.  Integer 123 is interpreted as UNIX epoch time and is converted like so:

```php
return date_create("@$value");
```
([Source](https://github.com/zendframework/zend-validator/blob/33f0141fce05155fb75b5d68fac5856d3a21ebd2/src/Date.php#L152-L155))

So the result is valid.  Alternatively, String 123 is interpreted as a formatted timestamp which must match the format specified (in your case `Y-m-d`)...it doesn't so it's not valid.

svycka:
@adamlundrigan if I specify date format I expect that only date in this format will be valid. for example when saving to database will be thrown exception when saving int not Y-m-d

svycka:
@adamlundrigan to be more clear imagine situation when you need to validate day of the month 'j' i would create it like this:
```php
$validator = new Validator\Date(['locale' => 'en', 'format' => 'j']);
$validator->isValid(5); // this is true, ok
$validator->isValid(99); // this is also true, this is not what I would expect, month does not have 99 days
```
So I don't get why you closed this. At least this must determine if format specified then validate it for this format not with some exceptions like if it is integer it is valid always..
when validator is used with forms yes value is always string so no problem, but often validators are used with API when data comes from json and it is not converted to string automatically.

GeeH:
This issue has been closed as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

PR #6407, opened 2014-06-23
Date validator bug issue #6406

svycka:


svycka:
This PR shows this bug: https://github.com/zendframework/zf2/issues/6406
````php
$validator = new \Zend\Validator\Date(array('locale' => 'en', 'format' => 'Y-m-d'));
$validator->isValid(123); // returns true
````
As @DASPRiD said this was intentionally desired. But I don't see why. also there was a PR https://github.com/zendframework/zf2/pull/6865 to fix it not for me to decide it was good or no.

@weierophinney what do you think?

hoppithek:
totally agree that an interger time should not validate as a formatted date string. The suggested solution in #6406 to make accepted types configurable sounds OK, but I would consider this a bug from the beginning.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-validator/issues/101

Issue #6497, opened 2014-07-30
CollectionInputFilter is never automatically instantiated by attachInputFilterDefaults


SteveTalbot:
`\Zend\Form\Form::attachInputFilterDefaults()` sets up the input filters for fieldsets, if these haven't been explicitly defined by the developer.  However it doesn't ever create a CollectionInputFIlter for a Collection fieldset.

This blog article explains why this is necessary:
http://adam.lundrigan.ca/2014/05/04/standalone-input-filter-classes-and-zf2-form-collections/

adamlundrigan:
This has since been implemented ([here](https://github.com/zendframework/zend-form/blob/767ed409dc19b127d36ed4cdc01de0318207b50b/src/Form.php#L815-L835))

PR #6498, opened 2014-07-30
Automatically instantiate CollectionInputFilter (#6497)

SteveTalbot:
This is the current workaround we're using for #6497.  It works for our use cases, but may break others.

### DIFF
--- a/tests/ZendTest/Form/FormTest.php
+++ b/tests/ZendTest/Form/FormTest.php
@@ -1253,7 +1253,11 @@ class FormTest extends TestCase
         $this->form->setInputFilter($inputFilter);
 
         $this->assertInstanceOf('Zend\InputFilter\CollectionInputFilter', $this->form->getInputFilter()->get('items'));
-        $this->assertCount(1, $this->form->getInputFilter()->get('items')->getInputs());
+
+        // #6497: No data added to form, so expect no valid inputs from collection input filter
+        $this->assertCount(0, $this->form->getInputFilter()->get('items')->getInputs());
+        $this->assertCount(0, $this->form->getInputFilter()->get('items')->getValidInput());
+        $this->assertCount(1, $this->form->getInputFilter()->get('items')->getInputFilter()->getInputs());
     }
 
     public function testFormValidationCanHandleNonConsecutiveKeysOfCollectionInData()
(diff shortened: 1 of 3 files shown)


tests/ZendTest/Mvc/Controller/Plugin/FilePostRedirectGetTest.php
```diff
@@ -448,7 +448,7 @@ public function testCorrectInputDataMerging()
                     ),
                 ),
                 1 => array(
-                    'text' => null,
+                    'text' => '',  // #6497: Input is present but empty
```
weierophinney: I'd prefer to see this as a separate case, to ensure the original `null` value case continues to pass.

tests/ZendTest/Mvc/Controller/Plugin/FilePostRedirectGetTest.php
```diff
@@ -324,7 +324,7 @@ public function testFieldsetAmountInFormEqualsFieldsetsInInputFilter()
 
         $this->assertEquals($params, $prgResult);
         $this->assertCount(count($params['links']),  $form->get('links')->getFieldsets());
-        $this->assertCount(count( $form->get('links')->getFieldsets()), $form->getInputFilter()->get('links')->getInputs());
+        $this->assertCount(count( $form->get('links')->getFieldsets()), $form->getInputFilter()->get('links')->getValidInput());  // #6497
```
weierophinney: This is a change in behavior. Can you quantify how this might affect existing applications, please?

tests/ZendTest/Form/FormTest.php
```diff
@@ -1253,7 +1253,11 @@ public function testDoNotApplyEmptyInputFiltersToSubFieldsetOfCollectionElements
         $this->form->setInputFilter($inputFilter);
 
         $this->assertInstanceOf('Zend\InputFilter\CollectionInputFilter', $this->form->getInputFilter()->get('items'));
-        $this->assertCount(1, $this->form->getInputFilter()->get('items')->getInputs());
+
+        // #6497: No data added to form, so expect no valid inputs from collection input filter
+        $this->assertCount(0, $this->form->getInputFilter()->get('items')->getInputs());
+        $this->assertCount(0, $this->form->getInputFilter()->get('items')->getValidInput());
+        $this->assertCount(1, $this->form->getInputFilter()->get('items')->getInputFilter()->getInputs());
```
weierophinney: Could the above be moved to an new test case, please?

weierophinney:
@SteveTalbot I've noted a few questions here. I can schedule for 2.4.0 if you can answer the BC concerns, but right now, I worry that this will affect existing applications.

adamlundrigan:
This has since been implemented ([here](https://github.com/zendframework/zend-form/blob/767ed409dc19b127d36ed4cdc01de0318207b50b/src/Form.php#L815-L835))

PR #6501, opened 2014-07-30
Provided failing tests for CurrencyFormat helper

demichl68:
referencing https://github.com/zendframework/zf2/pull/2318
We're on icu version 51.2 and still have the problem of wrong formatting for some locale and currency code combination

Please run this test and verify the results


weierophinney:
Question: is this a problem specifically in ZF2, or is it upstream in libicu? If the latter, I'm not sure we can do anything about this...

demichl68:
Sorry for the late response. This is not ZF2 specific problem. It's in fact the libicu version acting weird here.
This issue can be closed!

Issue #6312, opened 2014-05-22
[Zend\I18n] php.ini setting "intl.use_exceptions" on true


ThaDafinser:
When using `intl.use_exceptions` as `true` following validator will fail hard with the exception from the intl component:
* DateTime
* Float
* Int

Possible solutions:
* let it be like it is today (maybe this setting nobody used expect me?)
* try/catch block in addition to the `intl_is_failure()` check
* change php.ini setting always to false during validation and change it back after the check

Note: This feature was introduced with PHP 5.5
http://www.php.net/manual/en/class.intlexception.php

Ocramius:
Handled in #6598

PR #6598, opened 2014-08-26
add try/catch around statements that execute intl_is_failure()

gkralik:
Since PHP 5.5.0, the configuration option 'intl.use_exceptions' is available.
When set to true, IntlException's are thrown instead of just setting a error
code, so no calls to intl_is_failure() are necessary.
Without the try/catch, the Int, Float and DateTime I18n validators hard-fail
in case 'intl.use_exceptions' is set to true.

NOTE: catching Exception (super class of IntlException), as IntlException is
not available prior to PHP 5.5.0.

refs #6312

### DIFF
--- a/library/Zend/I18n/Validator/DateTime.php
+++ b/library/Zend/I18n/Validator/DateTime.php
@@ -12,6 +12,7 @@ namespace Zend\I18n\Validator;
 use Locale;
 use IntlDateFormatter;
 use Traversable;
+use IntlException;
 use Zend\I18n\Exception as I18nException;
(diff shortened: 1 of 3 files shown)


Ocramius:
Merged, thanks @gkralik! (I added the caught `IntlExceptions` as `$previousException` when re-thrown)

`master`: zendframework/zf2@19b8cbafe83b5d86663ed9311809d6fdafa842e4
`develop`: zendframework/zf2@490e5c84e3119e10842bf2928cfcd40c86d0f5b8

PR #6600, opened 2014-08-26
add label option to add css class to wrap element

lamer-x11:
I've added an option to add a css class to the `<span>` element that wraps form labels. Useful when working with bootstrap forms.


weierophinney:
Closing. This can be done in custom form view helpers, which can even be aliased to resolve with the same names as the current ones.

PR #6602, opened 2014-08-27
Fatal error when using Zend-service Twitter

botweb:
Fatal error: Call to a member function connect() on a non-object in zendframework/zend-http/Zend/Http/Client.php on line 1358


GeeH:
New issue can be found at: https://github.com/zendframework/zend-http/issues/80

PR #6619, opened 2014-09-01
Feature/log writer mail - 

Tauop:
Hi,

This patch allow the creation of an Zend\Mail\Message in Zend\Log\Writer\Mail through options passed to the writer. if $mail['mail'] is an array, we create a Zend\Mail\Message.

Here is a example of configuration, allowed by this patch

```
    'log' => array(
        'Log\App' => array(
            'writers' => array(
                'mail_log' => array(
                    'name'      => 'mail',
                    'options'   => array(
                        'mail' => array(
                            'from'      => 'system@example.com',
                            'to'        => 'admin@example.com',
                            'Encoding'  => 'UTF-8',
                        ),
                        'subject_prepend_text' => '[alert]',
                    ),
                ),
            ),
        ),
    ),
```

As this is my first pull-request in zf2, don't hesitate to give me feedback so that I can be better.

Cheers,
Patrick

### DIFF
--- a/library/Zend/Log/Writer/Mail.php
+++ b/library/Zend/Log/Writer/Mail.php
@@ -86,6 +86,9 @@ class Mail extends AbstractWriter
             }
             $transport = isset($mail['transport']) ? $mail['transport'] : null;
             $mail      = isset($mail['mail']) ? $mail['mail'] : null;
+            if (is_array($mail)) {
+                $mail = new MailMessage($mail);
+            }
         }
 
         // Ensure we have a valid mail message
--- a/library/Zend/Mail/Message.php
+++ b/library/Zend/Mail/Message.php
@@ -53,7 +53,7 @@ class Message
      * @return self
      * @throws Exception\InvalidArgumentException
      */
-    public function setOptions($options)
+    private function setOptions($options)
     {
         if (!is_array($options) && !$options instanceof Traversable) {
             throw new Exception\InvalidArgumentException(sprintf(
--- a/library/Zend/Mail/Message.php
+++ b/library/Zend/Mail/Message.php
@@ -36,6 +36,44 @@ class Message
     protected $encoding = 'ASCII';
 
     /**
+     * Constructor.
+     *
+     * @param null|array|Traversable $options (Optional) Options to set
+     * @param array|Traversable $options
+     */
+    public function __construct($options = null)
+    {
+        if (null !== $options) {
+            $this->setOptions($options);
+        }
+    }
+
+    /**
+     * @param  array|Traversable $options         Options to set
+     * @return self
+     * @throws Exception\InvalidArgumentException
+     */
+    public function setOptions($options)
+    {
+        if (!is_array($options) && !$options instanceof Traversable) {
+            throw new Exception\InvalidArgumentException(sprintf(
+                '"%s" expects an array or Traversable; received "%s"',
+                __METHOD__,
+                (is_object($options) ? get_class($options) : gettype($options))
+            ));
+        }
+
+        foreach ($options as $key => $value) {
+            $setter = 'set' . str_replace(' ', '', ucwords(str_replace('-', ' ', str_replace('_', ' ', $key))));
+            if (method_exists($this, $setter)) {
+                $this->{$setter}($value);
+            }
+        }
+
+        return $this;
+    }
+
+    /**
      * Is the message valid?
      *
      * If we don't any From addresses, we're invalid, according to RFC2822.
--- a/tests/ZendTest/Mail/MessageTest.php
+++ b/tests/ZendTest/Mail/MessageTest.php
@@ -734,7 +734,6 @@ class MessageTest extends \PHPUnit_Framework_TestCase
             $this->assertInstanceOf('Zend\Mail\AddressList', $value);
             $this->assertEquals(1, count($value));
             $this->assertTrue($value->has($options[$key]));
-
         }
     }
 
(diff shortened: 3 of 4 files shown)


library/Zend/Mail/Message.php
```diff
@@ -36,6 +36,44 @@ class Message
     protected $encoding = 'ASCII';
 
     /**
+     * Constructor.
+     *
+     * @param null|array|Traversable $options (Optional) Options to set
+     * @param array|Traversable $options
+     */
+    public function __construct($options = null)
```
Ocramius: This new API doesn't really need to exist due to the logger feature, does it?

library/Zend/Mail/Message.php
```diff
@@ -36,6 +36,44 @@ class Message
...
+            $this->setOptions($options);
+        }
+    }
+
+    /**
+     * @param  array|Traversable $options         Options to set
+     * @return self
+     * @throws Exception\InvalidArgumentException
+     */
+    public function setOptions($options)
```
Ocramius: Same as above. Also, I wouldn't make this method `private`, if it is required

tests/ZendTest/Mail/MessageTest.php
```diff
@@ -695,4 +695,87 @@ public function testPassEmptyArrayIntoSetPartsOfMimeMessageShouldReturnEmptyBody
...
+            'bcc'       => 'getBcc',
+            'reply-to'  => 'getReplyTo',
+        );
+
+        foreach ($getMethods as $key => $method) {
+            $value = $message->{$method}();
+            $this->assertInstanceOf('Zend\Mail\AddressList', $value);
+            $this->assertEquals(1, count($value));
+            $this->assertTrue($value->has($options[$key]));
+
```
malukenho: remove this blank line

library/Zend/Mail/Message.php
Tauop: I don't see the feature you reference. Can you explain me ?
In original constructor, if [$mail is an array](https://github.com/zendframework/zf2/blob/master/library/Zend/Log/Writer/Mail.php#L82), we call the parent constructor (seek for filters and formatter) and then [get $mail['mail']](https://github.com/zendframework/zf2/blob/master/library/Zend/Log/Writer/Mail.php#L88). After, the constructor check that [$mail is an Zend\Mail\Message instance](https://github.com/zendframework/zf2/blob/master/library/Zend/Log/Writer/Mail.php#L92).

When does the Zend\Mail\Message instance is created ?
I don't see any plugin manager or factory involved here.

How can i configure a log writer which sends an email to administrator ?

(Sorry for the late reply. I saw your reply few days ago only)

library/Zend/Mail/Message.php
Tauop: I don't understand your note. the method is public, not private.

tests/ZendTest/Mail/MessageTest.php
Tauop: done. thanks

library/Zend/Mail/Message.php
basz: ping @ocramius @Tauop is confused, could you help him out?

Ocramius: I'm asking for the method to be `private` unless there is any reason for it to be `public`

library/Zend/Log/Writer/Mail.php
```diff
@@ -86,6 +86,9 @@ public function __construct($mail, Transport\TransportInterface $transport = nul
             }
             $transport = isset($mail['transport']) ? $mail['transport'] : null;
             $mail      = isset($mail['mail']) ? $mail['mail'] : null;
+            if (is_array($mail)) {
```
malukenho: You don't need this verification, because you're already inside of a `is_array($mail)`

library/Zend/Mail/Message.php
Tauop: You're right. I put it private.

library/Zend/Log/Writer/Mail.php
Tauop: We need this verification.
First : check that $mail is an array.
Then : $mail = $mail['mail']

So this line checks that $mail['mail'] is an array.

library/Zend/Mail/Message.php
weierophinney: @Tauop I think what @Ocramius is getting at is that you're introducing a new API into `Zend\Mail\Message` by introducing the constructor (which did not exist before). I'm looking at it, and I'm of the opinion that it may make more sense to introduce a factory method instead, either in this class, or as a separate class. Usage would then be something like:

```php
$mail = MailFactory::getInstance($options);
```

@Ocramius I can see why the new API would be useful, though, in the particular situation that is proposed in this PR. Since the various message members are only known based on the configuration provided, and the specific members could vary based on invocation, having a factory becomes necessary, as `Zend\Mail\Message` is a value object.

weierophinney:
@Tauop Please review the comments from myself and @Ocramius above. I'm not convinced that we should be adding to `Zend\Mail\Message`, but instead should have a factory/builder for creating a message from configuration.

Tauop:
@weierophinney @Ocramius Thank for your support.
I've had lot of work recently. Sorry for the delay.
I've made a new PR

Tauop:
Here is the PR #7294

weierophinney:
Closed in favor of #7294

PR #6624, opened 2014-09-02
Feature: new URI type (for http over unix socket)

slopjong:
Unix sockets are also used for http, however, no existing URI type allows specifying an URI that takes a filepath as the authority which is a requirement in order to use the path for the actual web resource. Related to PR #6626.


weierophinney:
@slopjong — I'd love to have this for 2.4; will you be able to prepare tests in the next 2 weeks? If so, I'll add it to the milestone.

weierophinney:
@slopjong cool -- I'll keep this under the "awaiting author updates", which will make it easy for me to search for changes in the next week; once I see the tests, I'll assign the milestone.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-uri/issues/7

PR #6626, opened 2014-09-02
Feature/support unix domain socket

slopjong:
Unix sockets are also used for http, however, the current socket adapter just supports tcp/ip sockets. Related to PR #6624.


weierophinney:
@slopjong Same comment here as on #6624 — I'd like this for 2.4, but need unit tests. Can you do this in the next 2 weeks?

GeeH:
New issue can be found at: https://github.com/zendframework/zend-http/issues/79

PR #6646, opened 2014-09-07
Added password prompt to Console

richardjh:
SImple prompt that doesn't echo back (or can echo *s). Useful for reading in passwords.
If accepted I'll write up the documentation too.

### DIFF
--- a/library/Zend/Console/Prompt/Password.php
+++ b/library/Zend/Console/Prompt/Password.php
@@ -3,7 +3,7 @@
  * Zend Framework (http://framework.zend.com/)
  *
  * @link      http://github.com/zendframework/zf2 for the canonical source repository
- * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
--- a/library/Zend/Console/Prompt/Password.php
+++ b/library/Zend/Console/Prompt/Password.php
@@ -14,7 +14,7 @@ class Password extends AbstractPrompt
     /**
      * @var string
      */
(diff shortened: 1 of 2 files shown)


library/Zend/Console/Prompt/Password.php
```diff
@@ -0,0 +1,103 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2015 ;)

tests/ZendTest/Console/Prompt/PasswordTest.php
```diff
@@ -0,0 +1,63 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2015 ;)

richardjh:
Thanks @samsonasik I have updated copyright dates.

Ocramius:
This is just stuck because the console prompt is waiting forever for input by the user (during the `while` loop execution). I'm changing the tests so that they mock out the console adapter.

Ocramius:
Will need to rewrite some bits of this before pushing it out (due to security issues). Additionally, I think the `echo` functionality should just go away (security concerns).

Ocramius:
This was merged into `develop` at zendframework/zf2@306c17a2818cbd87669fa71b1ae70a6e5bb96820

I had to rewrite some bits because the prompt was actually acting really weird when entering multiple rapidly subsequent chars: figured that (for security) the best solution was to clear the line at every keystroke instead.

PR #6656, opened 2014-09-10
A getElements method has been added to DateSelect and MonthSelect which ...

MajorCaiger:
...returns an array of the Select elements created.

Unit tests have been added to cover the new method.


Ocramius:
Merged, thanks @MajorCaiger!

`develop`: zendframework/zf2@d2d7146bc7d9246ea73118144157c8857d4a5ce8

PR #6658, opened 2014-09-10
Ldap rename bug: emulate flag was always true

kristof9851:
Ldap rename bug: emulate flag was always true


weierophinney:
@kristofkotai Since this is a boolean condition, changing it could be considered a BC break unless the feature as provided is broken as-is. We cannot know that, however, without either (a) a test, and/or (b) links to specification documentation that will demonstrate that the flag needs to change.

PR #6664, opened 2014-09-11
Use file_exists to check for an uploaded file.

akrabat:
It turns out that stream_resolve_include_path doesn't work for a default
Windows install where upload_tmp_dir is set to C:\Windows\Temp. I'm
unclear if this is intended PHP behaviour or not, but the upload file
validator should not be checking for a file of the same name on the
include path regardless.


ThaDafinser:
@akrabat seems to be still a bug on windows with default path.

Not in this validator, but in `Zend\Validator\File\Size` and many others:
https://github.com/zendframework/zf2/search?q=stream_resolve_include_path&type=Code&utf8=%E2%9C%93

PR #6669, opened 2014-09-15
Support subscribing to folders in IMAP

ameir:
This brings in support for the IMAP `SUBSCRIBE` command.

There were no direct tests on Mail\Protocol, so I wasn't sure how you wanted to handle that.

Thanks!
-Ameir

### DIFF
--- a/library/Zend/Mail/Protocol/Imap.php
+++ b/library/Zend/Mail/Protocol/Imap.php
@@ -753,6 +753,17 @@ class Imap
     }
 
     /**
+     * subscribe to a folder
+     * 
+     * @param string $folder folder name
+     * @return bool success
+     */
+    public function subscribe($folder)
+    {
+        return $this->requestAndResponse('SUBSCRIBE', array($this->escapeString($folder)), true);
+    }
+    
+    /**
      * permanently remove messages
      *
      * @return bool success


weierophinney:
@ameir Can you at least link to documentation showing that this is the command used for subscriptions?

ameir:
@weierophinney, sure, this command is described in the IMAP4rev1 RFC:

https://tools.ietf.org/html/rfc3501

```
6.3.6.  SUBSCRIBE Command

   Arguments:  mailbox

   Responses:  no specific responses for this command

   Result:     OK - subscribe completed
               NO - subscribe failure: can't subscribe to that name
               BAD - command unknown or arguments invalid

      The SUBSCRIBE command adds the specified mailbox name to the
      server's set of "active" or "subscribed" mailboxes as returned by
      the LSUB command.  This command returns a tagged OK response only
      if the subscription is successful.

      A server MAY validate the mailbox argument to SUBSCRIBE to verify
      that it exists.  However, it MUST NOT unilaterally remove an
      existing mailbox name from the subscription list even if a mailbox
      by that name no longer exists.

           Note: This requirement is because a server site can
           choose to routinely remove a mailbox with a well-known
           name (e.g., "system-alerts") after its contents expire,
           with the intention of recreating it when new contents
           are appropriate.


   Example:    C: A002 SUBSCRIBE #news.comp.mail.mime
               S: A002 OK SUBSCRIBE completed
```
I hope that helps!

weierophinney:
Merged for 2.4; please provide documentation of the new feature at https://github.com/zendframework/zf2-documentation

PR #6681, opened 2014-09-22
Update FirePhpInterface for Log Writer

ChrisSchreiber:
Update /Zend/Log/Writer/FirePhp/FirePhpInterface methods to match signatures in FirePhp. This is needed because the /Zend/Log/Formatter/FirePhp changes the log message to a label when the log contains extra data. The message/label is then lost because FirePhpBridge does not pass it along to FirePhp.


ChrisSchreiber:
This was resolved in #6370

Issue #6689, opened 2014-09-23
Http Client with Curl Adapter combined with malformed return headers results in 'Error parsing body' - Exception


kasitmp:
I saw this issue on some major webpages.

In  Zend\Http\Client\Adapter\Curl there is a [line](https://github.com/zendframework/zf2/blob/master/library/Zend/Http/Client/Adapter/Curl.php#L425) where the 'Transfer-Encoding: chunked' is removed so it will not be interpreted from the Zend\Http\Response [decodeChunkedBody](https://github.com/zendframework/zf2/blob/master/library/Zend/Http/Response.php#L460) function, because curl already decodes the chunked body.

Here is the example header from buzzfeed.com:
```
HTTP/1.1 200 OK
Server: Apache
Content-Type: text/html
X-BuzzFeed: webdr08
Vary: Accept-Encoding
Content-Encoding: gzip
Cache-Control: max-age=3
Date: Tue, 23 Sep 2014 14:42:19 GMT
Transfer-Encoding:  chunked
```

As you can see there are two whitespaces between 'Transfer-Encoding:' and 'chunked' causing the ```stripos($this->response, "Transfer-Encoding: chunked\r\n")``` to become false.

As a result the decodeChunkedBody function gets called and throws an exception.

PR #6691, opened 2014-09-23
fixed removing handled header parts from response header

kasitmp:
This fix should solve two potential issues:
1.  `$this->response = str_ireplace("Transfer-Encoding: chunked\r\n", '', $this->response);` is able to remove content, also from a response body. To solve this I separate the header from the body first before manipulating the it.
2. There are pages in the wild which return headers with more than one space between _Transfer-Encoding:_ and _chunked_. This fix should solve these issues via regex

I coded this to fix my issue #6689

### DIFF
--- a/library/Zend/Http/Client/Adapter/Curl.php
+++ b/library/Zend/Http/Client/Adapter/Curl.php
@@ -421,18 +421,25 @@ class Curl implements HttpAdapter, StreamInterface
             throw new AdapterException\RuntimeException("Error in cURL request: " . curl_error($this->curl));
         }
 
+        // separating header from body because it is dangerous to accidentially replace strings in the body
+        $responseHeaderSize = curl_getinfo($this->curl, CURLINFO_HEADER_SIZE);
+        $responseHeaders = substr($this->response, 0, $responseHeaderSize);
+
         // cURL automatically decodes chunked-messages, this means we have to disallow the Zend\Http\Response to do it again
-        if (stripos($this->response, "Transfer-Encoding: chunked\r\n")) {
-            $this->response = str_ireplace("Transfer-Encoding: chunked\r\n", '', $this->response);
-        }
+        $responseHeaders = preg_replace("/Transfer-Encoding:\s*chunked\\r\\n/", "", $responseHeaders);
 
         // cURL can automatically handle content encoding; prevent double-decoding from occurring
         if (isset($this->config['curloptions'][CURLOPT_ENCODING])
             && '' == $this->config['curloptions'][CURLOPT_ENCODING]
-            && stripos($this->response, "Content-Encoding: gzip\r\n")
         ) {
-            $this->response = str_ireplace("Content-Encoding: gzip\r\n", '', $this->response);
+            $responseHeaders = preg_replace("/Content-Encoding:\s*gzip\\r\\n/", "", $responseHeaders);
         }
+
+        // cURL automatically handles Proxy rewrites, remove the "HTTP/1.0 200 Connection established" string:
+        $responseHeaders = preg_replace("/HTTP\/1.0\s*200\s*Connection\s*established\\r\\n\\r\\n/", "", $responseHeaders);
+
+        // replace old header with new, cleaned up, header
+        $this->response = substr_replace($this->response, $responseHeaders, 0, $responseHeaderSize);
 
         // Eliminate multiple HTTP responses.
         do {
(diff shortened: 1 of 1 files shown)


weierophinney:
@kasitmp As @Ocramius noted, we can modify our Travis configuration to startup a web server using the built-in PHP webserver with versions 5.4+, and then test against that (we actually have tests already that only test against a web server that runs against a document root in the test hierarchy, and we use these for integration spot-checking). Alternately, you can create mock responses to test against (we also do this in a number of places). Considering this is cURL specific, you likely need to go with the first option.

We *do* need a test, however, to validate that this works. If you think you can work with us to make that happen in the next two weeks, I'll mark this for 2.4.

kasitmp:
I'm really busy at the moment, but I will try. This is a while back so I will need some time to get back into my code, I guess :)
Count me in, I will try doing it during the next week.

library/Zend/Http/Client/Adapter/Curl.php
```diff
@@ -421,19 +421,26 @@ public function write($method, $uri, $httpVersion = 1.1, $headers = array(), $bo
             throw new AdapterException\RuntimeException("Error in cURL request: " . curl_error($this->curl));
         }
 
+        // separating header from body because it is dangerous to accidentially replace strings in the body
+        $responseHeaderSize = curl_getinfo($this->curl, CURLINFO_HEADER_SIZE);
+        $responseHeaders = substr($this->response, 0, $responseHeaderSize);
+
         // cURL automatically decodes chunked-messages, this means we have to disallow the Zend\Http\Response to do it again
```
kasitmp: I separated header from body, to be sure that i don't manipulate the body. Possible test:
Textfile with Transfer-Encoding: chunked\r\n inside AND the webserver must not make use of chunking.
In the old version the assert that the file is the same would probably fail.

library/Zend/Http/Client/Adapter/Curl.php
```diff
@@ -421,19 +421,26 @@ public function write($method, $uri, $httpVersion = 1.1, $headers = array(), $bo
...
-        if (stripos($this->response, "Transfer-Encoding: chunked\r\n")) {
-            $this->response = str_ireplace("Transfer-Encoding: chunked\r\n", '', $this->response);
-        }
+        $responseHeaders = preg_replace("/Transfer-Encoding:\s*chunked\\r\\n/", "", $responseHeaders);
```
kasitmp: In most cases the `Transfer-Encoding:*whitespace*chunked` is true, but I found some cases (at that time back than buzzfeed.com for example) where there was no whitespace between the two parts. Problem: I was not able to send a header without that whitespace. Maybe the server or php corrects the header before sending.

library/Zend/Http/Client/Adapter/Curl.php
```diff
@@ -421,19 +421,26 @@ public function write($method, $uri, $httpVersion = 1.1, $headers = array(), $bo
...
 
         // cURL can automatically handle content encoding; prevent double-decoding from occurring
         if (isset($this->config['curloptions'][CURLOPT_ENCODING])
             && '' == $this->config['curloptions'][CURLOPT_ENCODING]
-            && stripos($this->response, "Content-Encoding: gzip\r\n")
         ) {
-            $this->response = str_ireplace("Content-Encoding: gzip\r\n", '', $this->response);
+            $responseHeaders = preg_replace("/Content-Encoding:\s*gzip\\r\\n/", "", $responseHeaders);
         }
```
kasitmp: Same as above

library/Zend/Http/Client/Adapter/Curl.php
```diff
@@ -421,19 +421,26 @@ public function write($method, $uri, $httpVersion = 1.1, $headers = array(), $bo
...
 
+        // cURL automatically handles Proxy rewrites, remove the "HTTP/1.0 200 Connection established" string:
+        $responseHeaders = preg_replace("/HTTP\/1.0\s*200\s*Connection\s*established\\r\\n\\r\\n/", "", $responseHeaders);
+
```
kasitmp: Just changed the position and also use a regexp for the same reasons as above

library/Zend/Http/Client/Adapter/Curl.php
```diff
@@ -421,19 +421,26 @@ public function write($method, $uri, $httpVersion = 1.1, $headers = array(), $bo
...
+        // replace old header with new, cleaned up, header
+        $this->response = substr_replace($this->response, $responseHeaders, 0, $responseHeaderSize);
+
         // Eliminate multiple HTTP responses.
```
kasitmp: traded old header part against the new, cleaned up, version

kasitmp:
I'm sorry, but I can't find time to do the tests at the moment. But I describe what I would test, if I would know how to do it with your infrastructure. Just look at the annotations https://github.com/zendframework/zf2/pull/6691/files

weierophinney:

PR #6703, opened 2014-09-26
Failing test for I18n\Validator\Float

ThaDafinser:
Following form element example:

``` php
    public function getInputFilterSpecification()
    {
        return [
            'value' => [
                'filters' => [
                    [
                        'name' => 'Zend\I18n\Filter\NumberFormat'
                    ]
                ],

                'validators' => [
                    [
                        'name' => 'Zend\I18n\Validator\Float'
                    ],
                    [
                        'name' => 'Zend\Validator\NotEmpty',
                        'options' => [
                            'type' => NotEmpty::STRING
                        ]
                    ]
                ]
            ]
     ];
}
```

If the user provides a string, the filter is skipped so the string goes directly to the validator


ThaDafinser:
Cannot be merged here anymore

PR #6709, opened 2014-09-27
HtmlTag helper

nikolaposa:
New view helper used for rendering `<html>` tag of a web page, to which some custom attributes can be added dynamically.
## Intent

Idea is to have ability to dynamically attach some attributes to the opening HTML tag in a layout. For example:
`<html xmlns="http://www.w3.org/1999/xhtml" xmlns:og="http://ogp.me/ns#">`
## Usage

Rendering tag in a layout:

``` php
echo $this->html()->openTag(); 
//...
echo $this->html()->closeTag();
```

Adding attributes:

``` php
//single
$this->html()->setAttribute('xmlns', 'http://www.w3.org/1999/xhtml');

//multiple
$this->html()->setAttributes(array(
    'xmlns' => 'http://www.w3.org/1999/xhtml',
    'prefix' => 'og: http://ogp.me/ns#'
));

//shortcut method
$this->html(array(
    'xmlns' => 'http://www.w3.org/1999/xhtml',
    'prefix' => 'og: http://ogp.me/ns#'
));
```

There is also option to automatically pre-set appropriate namespace attributes, based on currently set DOCTYPE (`Doctype` helper):

``` php
//At some point:
$this->doctype('xhtml');

$this->html()
    ->setUseNamespaces(true) //'xmlns' attribute will be automatically set to 'http://www.w3.org/1999/xhtml'
    ->setAttributes(array(
        'prefix' => 'og: http://ogp.me/ns#'
    ));
```

### DIFF
--- a/library/Zend/View/Helper/HtmlTag.php
+++ b/library/Zend/View/Helper/HtmlTag.php
@@ -28,12 +28,12 @@ class HtmlTag extends AbstractHtmlElement
      * Retrieve object instance; optionally add attributes.
      *
      * @param array $attribs
-     * @return \Zend\View\Helper\HtmlTag
+     * @return self
      */
     public function __invoke(array $attribs = array())
     {
         if (!empty($attribs)) {
-            $this->addAttributes($attribs);
+            $this->setAttributes($attribs);
         }
 
         return $this;
@@ -44,7 +44,7 @@ class HtmlTag extends AbstractHtmlElement
      *
      * @param string $attrName
      * @param string $attrValue
-     * @return \Zend\View\Helper\HtmlTag
+     * @return self
      */
     public function setAttribute($attrName, $attrValue)
     {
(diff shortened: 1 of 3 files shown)


library/Zend/View/Helper/HtmlTag.php
```diff
@@ -0,0 +1,148 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2015 ;)

tests/ZendTest/View/Helper/HtmlTagTest.php
```diff
@@ -0,0 +1,136 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2015 ;)

library/Zend/View/Helper/HtmlTag.php
nikolaposa: Do I really have to modify that? :) I also have some other PRs planned for 2.4.0 within which new classes are added, like this one: https://github.com/zendframework/zf2/pull/6058, and I haven't got such remark from a code reviewer.

Ocramius: No, it will be done after merge. I don't feel like burdening existing PRs with these changes is needed at all ;-)

weierophinney:

PR #6712, opened 2014-09-28
Added form annotation builder factory

carnage:
Two outstanding items for this pull request:
1. What is the best method to get it into the default service locator config?
2. Should the factory support adding custom annotations and annotation listeners to the builder?


library/Zend/Form/Annotation/AnnotationBuilder.php
```diff
@@ -383,6 +389,18 @@ protected function checkForExclude($annotations)
...
+     * @return \Zend\Code\Annotation\Parser\DoctrineAnnotationParser
+     */
+    public function getAnnotationParser()
+    {
+        if (null === $this->annotationParser) {
+            $this->annotationParser = new Parser\DoctrineAnnotationParser();
+        }
+
+        return $this->annotationParser;
+    }
```
weierophinney: Not sure this abstraction is necessary, if there's no way to inject the parser otherwise.

library/Zend/Mvc/Service/ServiceListenerFactory.php
```diff
@@ -53,6 +53,7 @@ class ServiceListenerFactory implements FactoryInterface
             'DiStrictAbstractServiceFactory' => 'Zend\Mvc\Service\DiStrictAbstractServiceFactoryFactory',
             'FilterManager'                  => 'Zend\Mvc\Service\FilterManagerFactory',
             'FormElementManager'             => 'Zend\Mvc\Service\FormElementManagerFactory',
+            'Zend\Form\Annotation\FormAnnotationBuilder' => 'Zend\Mvc\Service\FormAnnotationBuilderFactory',
```
weierophinney: I think we can just call it "FormAnnotationBuilder".

prolic: :+1:

library/Zend/Form/Annotation/AnnotationBuilder.php
prolic: :+1: yes, normal getters and setters are possible. inject the parser in the factory.

carnage: If this were to be removedit would break bc -> users would have to instantiate via the factory or risk not having a parser

library/Zend/Mvc/Service/ServiceListenerFactory.php
carnage: Quoting @Ocramius on why it was changed to the fqcn:

"I'd prefer to just have the FQCN here. Some advantages

    less keys to remember
    can be statically referenced via PHP 5.5's ClassName::class syntax, allowing IDE static introspection to find non-existing classes
    less documentation required
    more predictable from a consumer perspective"

gianarb: In my opinion one class defined how FQNC and others in different way is not a good solution

library/Zend/Form/Annotation/AnnotationBuilder.php
weierophinney: @carnage I missed that it was being used in the factory as well, and, as such, is part of the public API you need to expose. We can keep it as-is.

library/Zend/Mvc/Service/ServiceListenerFactory.php
weierophinney: On merge, I'll add an alias for the FQCN; for v3, we can move to all FQCN, but considering no other framework-defined factories do this currently, we should remain consistent.

weierophinney:

PR #6713, opened 2014-09-28
Date select filter

carnage:
This pull request refactors the callback filters used in date and date time select form elements into concrete filters. This allows the filter to be reapplied when the user is using annotations and sets prefer_form_input_filter to true (as is done by default by Doctrine's form annotation builder) fixing a bug with validation.

As a bonus, the filter also contains flags to return null if the data select is empty (currently, empty selects return a string of -- or -- ::) The Date validator has also been updated with a flag to allow for this new behaviour.


carnage:
Theses filters are designed to be used in conjunction with the (Date|Month|DateTime)Select form elements: given that we plan to split the repos; should these live in a sub-namespace of Form?

The magic constants could then be referenced from the Form elements/view helpers that also use them

library/Zend/Filter/AbstractDateDropdown.php
```diff
@@ -0,0 +1,107 @@
...
+                array_reduce($value, function ($soFar, $value) { return $soFar & empty($value); }, true)
+            ) {
+                return null;
+            }
+
+            ksort($value);
+            $value = vsprintf($this->format, $value);
+        }
+
+        return $value;
```
weierophinney: Maybe create a test covering this, demonstrating a bad format and/or value, and an expected exception? That may help you or others to come up with a solution.

library/Zend/Filter/DateTimeSelect.php
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * @param mixed $value
+     * @return array|mixed|null|string
+     */
+    public function filter($value)
+    {
+        if (is_array($value)) {
+            if ($this->nullOnEmpty() && (
+                    empty($value['year']) ||
```
weierophinney: @Ocramius confirmed - these keys are all defined as part of the date input type in HTML5.

weierophinney:
@carnage I'd like to merge, but:

- We need a rebase
- There are open questions about how to handle invalid inputs

If you can resolve these in the next two weeks, I can keep it in the 2.4 release.

library/Zend/Filter/AbstractDateDropdown.php
```diff
@@ -0,0 +1,107 @@
...
+     */
+    protected $nullOnAllEmpty = false;
+
+    /**
+     * Sprintf format string to use for formatting the date, fields will be used in alphabetical order.
+     *
+     * @var string
+     */
+    protected $format = '';
+
```
coolms: A constructor should be here, otherwise there is no way setting options to filter. As an example:

```php
/**
 * Constructor
 *
 * @param array|Traversable $options OPTIONAL
 */
public function __construct($options = null)
{
    if (static::isOptions($options)) {
        $this->setOptions($options);
    }
}
```

library/Zend/Validator/Date.php
```diff
@@ -101,6 +107,24 @@ public function setFormat($format = self::FORMAT_DEFAULT)
...
+    }
+
+    /**
+     * @return boolean
+     */
+    public function allowNull()
+    {
+        return $this->allowNull;
+    }
+
```
coolms: The name of this method isn't correct. It should be getAllowNull() because of the way options are set into validator. \Zend\Validator\AbstractValidator::setOptions($options) tries to call AllowNull() method first with option's value as parameter, whereupon option will not be set.

weierophinney: @coolms I personally disagree, as I think the proposed syntax reads better in conditionals:

```php
if ($date->allowNull()) {
    // ...
}
```

Since it's a method providing a boolean response, and will be used for conditionals, the focus should not be on whether it's _getting_ a value, but on what it returns.

carnage:
Rebase messed this up. Reopened against develop

PR #6717, opened 2014-09-30
added NameNavigationFactory to create navigation from config key name

sandrokeil:
I use several navigation configs and i want to create navigation instances via config name. For example:

``` php
// navigation config e.g. in module.config.php
return array(
    // other stuff before
    'navigation' => array(
        'special' => array(
            array(
                'label' => 'Page 1',
                'uri' => 'page1.html'
            ),
            array(
                'label' => 'MVC Page',
                'route' => 'foo',
                'pages' => array(
                    array(
                        'label' => 'Sub MVC Page',
                        'route' => 'foo'
                    )
                )
            ),
            array(
                'label' => 'Page 3',
                'uri' => 'page3.html'
            )
        )
    )
);
```

So it's very easy to create new navigation instances which uses another config.

``` php
// service manager config e.g. in module.config.php
return array(
    // other stuff before
    'service_manager' => array(
        'factories' => array(
            'mymodule.navigation.special' => new \Zend\Navigation\Service\NameNavigationFactory('special'),
        ),
    ),
);
```

Maybe we could remove `\Zend\Navigation\Service\DefaultNavigationFactory` and set parameter `$name = 'default'` in `\Zend\Navigation\Service\NameNavigationFactory::__construct`, but this is a BC break.

### DIFF
--- a/library/Zend/Navigation/Service/NameNavigationFactory.php
+++ b/library/Zend/Navigation/Service/NameNavigationFactory.php
@@ -1,0 +1,45 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Navigation\Service;
+
+/**
+ * Name navigation factory.
+ *
+ * Creates navigation depending on given config key name
+ */
+class NameNavigationFactory extends AbstractNavigationFactory
+{
+    /**
+     * Config name
+     *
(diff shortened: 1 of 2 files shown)


weierophinney:
Another approach that might better fit is to use an AbstractFactory instead of a straight factory. That factory can then use the service name you use (e.g., `Nav\Special`) to determine which key to look under (e.g., `navigation -> special`). This is a pattern we've used in a number of places already (DB adapters, input filters, forms, etc.), and is designed for exactly this use case (same instantiation pattern, but requiring discrete, named instances, potentially based on configuration values).

If you're willing to refactor to use this pattern, I can schedule this for 2.4; our timeframe is two weeks from this coming Monday.

sandrokeil:
Thanks for that hint. This sounds really good. I'm starting to work on it.

sandrokeil:
This was replaced with pull request #7245.

PR #6729, opened 2014-10-03
Zend controller test contains

nbochenko:
This allow to match a partial query node value content in AbstractHttpControllerTestCase:: assertQueryContentContains, and not only an exact match.
A node with the value "foo, bar, lol" would match "foo", "foo,", "bar, lol", etc...

@blanchonvincent

### DIFF
--- a/library/Zend/Test/PHPUnit/Controller/AbstractHttpControllerTestCase.php
+++ b/library/Zend/Test/PHPUnit/Controller/AbstractHttpControllerTestCase.php
@@ -13,8 +13,8 @@ use Zend\Dom\Document;
 
 abstract class AbstractHttpControllerTestCase extends AbstractControllerTestCase
 {
-    const CONTENT_CONTAINS_EXACT_MATCH = 'content_exact_match';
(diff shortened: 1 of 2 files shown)


weierophinney:
I tend to agree with @Ocramius ; new method names would make it easier to understand intent of the test (looking at the method name is simpler than looking at the arguments to the assertion method).

If you're willing to make that change, I'm happy to schedule this for 2.4; timeframe is ~ 9 March 2015.

nbochenko:
@weierophinney @Ocramius 
Sorry for the delay.
The "***Contains" methods are actually asserting an exact match. 
What method names would you like to add, something like : 
"assertResponseHeaderPartiallyContains"
"assertXpathQueryContentPartiallyContains"

Thanks.

nbochenko:
@blanchonvincent @waltertamboer @Ocramius

waltertamboer:
I guess you meant to ping @weierophinney instead of me :)

GeeH:
New issue can be found at: https://github.com/zendframework/zend-test/issues/30

PR #6731, opened 2014-10-03
fix for FilesSize validator usage with Input

bgotink:
There's a problem with `FilesSize` when used as validator for a `FileInput`: the validator doesn't accept input in the `$_FILES` format. It expects a filename or an array of filenames.

This PR adds support for the `$_FILES` format to the `FilesSize` validator.


weierophinney:

PR #6746, opened 2014-10-10
Di runtime definition optimisation - Cache result of processing and inline the check

nickpeirson:
We noticed a long time was spent in runtime definitions on some of our requests and that calls to `processClass` could run into tens of thousands in some cases. As `hasProcessedClass()` was only called from one place, I've inlined it to reduce the method call and cached the result of the check, which has made a significant difference to requests that have significant numbers of calls to `processClass`.


weierophinney:

PR #6748, opened 2014-10-10
Di Runtime Compiler Definition

nickpeirson:
Seems that the current CompilerDefinition class has not been maintained with changes that have been introduced in the RuntimeDefinition. This results in different class definitions generated by the CompilerDefinition than those generated by the RuntimeDefinition. To resolve this I have created RuntimeCompilerDefinition which extends RuntimeDefinition and adds the functionality of CompilerDefinition. It has been designed to be used as a drop in replacement for the CompilerDefinition, and as such should be usable any where that was previously been used.

### DIFF
--- a/library/Zend/Di/Definition/CompilerDefinition.php
+++ b/library/Zend/Di/Definition/CompilerDefinition.php
@@ -3,36 +3,21 @@
  * Zend Framework (http://framework.zend.com/)
  *
  * @link      http://github.com/zendframework/zf2 for the canonical source repository
- * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
 namespace Zend\Di\Definition;
 
-use Zend\Code\Annotation\AnnotationCollection;
-use Zend\Code\Reflection;
+use ReflectionException;
 use Zend\Code\Scanner\AggregateDirectoryScanner;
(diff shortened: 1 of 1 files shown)


weierophinney:
@nickpeirson GO FOR IT!

If you can do this by 9 March 2015, we can include it in 2.4. :)

nickpeirson:
Done. I did have to make one tweak to maintain BC, which I'm not sure about.

In summary, there's a test to see if `CompilerDefinition::hasMethodParameters()` returns false when the class doesn't exist. However in `RuntimeDefinition::hasMethodParameters()`, which is where the implementation now comes from, throws a `ReflectionException` in this case. I've overridden the method in `CompilerDefinition` to catch the exception and return false, which maintains BC, but my preferred solution would be to change the test and be consistent with `RuntimeDefinition`.

Maintaining BC seemed to be the path of least resistance, but I'm happy to switch it over if there's a consensus on the small BC break.

nickpeirson:
Travis is showing failed, but it looks like the same failures as on `develop`, but shout if I missed something that I need to fix.

library/Zend/Di/Definition/CompilerDefinition.php
```diff
@@ -3,74 +3,80 @@
  * Zend Framework (http://framework.zend.com/)
  *
  * @link      http://github.com/zendframework/zf2 for the canonical source repository
- * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
malukenho: revert

GeeH:
New issue can be found at: https://github.com/zendframework/zend-di/issues/18

PR #6752, opened 2014-10-12
Added ability to specify label position as an element label option

carnage:
Does exactly what it says on the tin

### DIFF
--- a/library/Zend/Form/View/Helper/FormRow.php
+++ b/library/Zend/Form/View/Helper/FormRow.php
@@ -213,7 +213,13 @@ class FormRow extends AbstractHelper
                     $labelOpen = $labelClose = $label = '';
                 }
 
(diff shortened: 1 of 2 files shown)


library/Zend/Form/View/Helper/FormRow.php
```diff
@@ -213,7 +213,13 @@ public function render(ElementInterface $element)
                     $labelOpen = $labelClose = $label = '';
                 }
 
-                switch ($this->labelPosition) {
+                $labelPosition = $this->labelPosition;
```
weierophinney: This line caused two test failures. Essentially, it was overriding lines 128-130, which were setting the value of `$labelPosition` to the instance property if it had not been passed. As such, I removed it, as the value will be properly initialized by this point already.

weierophinney:

PR #6753, opened 2014-10-12
Input filter annotation fix

carnage:
This PR fixes an issue with the InputFilter form annotation, when applied to a form it causes the annotation builder to totally ignore any element level validation. 

This fix allows the annotation to take an array option, if it is specified as an array, element level validation is merged with the form level input filter options. 

The main purpose of this is to allow specifiying a custom input filter type that has different behaviour to the default without having to create form specific validation and filtering rules within that class.


Ocramius:
@carnage merged, thanks!

`develop`: zendframework/zf2@1933c22bf3b9a9894aee08e04349c739ec1a536a

PR #6754, opened 2014-10-12
Added preserve_defined_order flag to annotation builder

### LINKED ISSUE #6653: Form annotation builder doesn't preserve fieldset/element order
If you build a form with several elements, a fieldset and a few more elements via annotations, the fieldset is placed at the end of the form. This can be worked around be explicitly setting a priority for each element but I think it would be cleaner to preserve the order of properties defined in the class. 

This can be achieved by removing the distinction between fieldsets and elements in the annotation builder, but I'm wondering if anyone can see any potential side effects with this fix?

carnage:
This flag when set preserve the elements in the order they are defined in the class as opposed to the default of putting fieldsets after elements.

Fixes https://github.com/zendframework/zf2/issues/6653


weierophinney:

PR #6756, opened 2014-10-13
implements @todo for setters/getters at Zend\Mime\Part

samsonasik:


samsonasik:
@Ocramius  merge-able now ;) ?

PR #6758, opened 2014-10-13
add getAdapterInstance() that return adapter instance at Zend\Filter\Encrypt

samsonasik:
and add <code>getAdapterName()</code> to get name of adapter same as <code>Zend\Filter\Compress</code> class to make it consistent.


samsonasik:
@Ocramius merge-able now ? ;)

PR #6763, opened 2014-10-14
Automatic hostname route adaptation 

nikolaposa:
## Problem

In situations where Hostname route is being utilized, for example, in case of a website: _www.example.com_ which also have some pages on a separate subdomain: _blog.example.com_, routes definition might look like:

``` php
//...
'router' => array(
    'routes' => array(
        'blog' => array(
            'type' => 'hostname',
            'options' => array(
                'route' => 'blog.example.com',
                'defaults' => array(
                    'controller' => 'Application\Controller\BlogController',
                    'action' => 'index',
                ),
            ),
            'may_terminate' => true,
            'child_routes' => array(
                'post' => array(
                    'type' => 'segment',
                    'options' => array(
                        'route' => '/[:slug]',
                        'constraints' => array(
                            'slug' => '[a-zA-Z0-9_-]+'
                        ),
                        'defaults' => array(
                            'action' => 'view'
                        )
                    )
                )
            )
        ),
        'user' => array(
            'may_terminate' => false,
            'type' => 'literal',
            'options' => array(
                'route'    => '/user',
                'defaults' => array(
                    'controller' => 'Application\Controller\User',
                ),
            ),
            'child_routes' => array(
                'view' => array(
                    'type' => 'segment',
                    'options' => array(
                        'route' => '/view[/:id]',
                        'constraints' => array(
                            'id' => '\d+'
                        ),
                        'defaults' => array(
                            'action' => 'view',
                        ),
                    ),
                ),
                'edit' => array(
                    'type' => 'segment',
                    'options' => array(
                        'route' => '/edit[/:id]',
                        'constraints' => array(
                            'id' => '\d+'
                        ),
                        'defaults' => array(
                            'action' => 'edit',
                        ),
                    ),
                ),
            ),
        )
    ),
),
//...
```

The problem occurs when URL for the `user/view` route is assembled on some blog page, for example:

``` php
echo $this->url('user/view', array('id' => 1));
```

That URL will point to the _blog.example.com_, even if `force_canonical` options is supplied.
## Solution

In this situations, Router component should respond in such way that assembled URLs always point to the appropriate hostname. This can be achieved by inflecting all the other non-hostname routes into **Chain** routes, consisting of a default hostname route and the inflecting route itself.

This improvement does just that, by automating that inflection of non-hostname routes. Thereby, developer is required to supply default hostname route definition along with other `'router'` options, for example:

``` php
'router' => array(
    'default_hostname_route' => array(
        'type' => 'hostname',
        'options' => array(
            'route' => 'www.example.com',
        ),
    ),
    'routes' => array(
        //route definitions...
    ),
),
```

As a result, assembling of a `user/view` route from the above example:

``` php
echo $this->url('user/view', array('id' => 1));
```

... would always result in: _http://www.example.com/user/1_, regardless of a subdomain on which it is being assembled.

### DIFF
--- a/library/Zend/Mvc/Router/Http/TreeRouteStack.php
+++ b/library/Zend/Mvc/Router/Http/TreeRouteStack.php
@@ -62,6 +62,35 @@ class TreeRouteStack extends SimpleRouteStack
             throw new Exception\InvalidArgumentException(__METHOD__ . ' expects an array or Traversable set of options');
         }
 
+        //If some hostname route is defined, adapt the others
+        //by chaining them onto the default hostname route
+        if (isset($options['routes']) && isset($options['default_hostname_route'])) {
+            if (!isset($options['route_plugins'])) {
+                throw new Exception\InvalidArgumentException('Missing "route_plugins" in options array');
+            }
+            $routePluginManager = $options['route_plugins'];
+
+            if ($options['routes'] instanceof Traversable) {
+                $options['routes'] = ArrayUtils::iteratorToArray($options['routes']);
+            }
+
+            $defaultHostnameRoute = $options['default_hostname_route'];
+
+            foreach ($options['routes'] as $name => $route) {
+                if ($route['type'] != 'hostname') {
+                    $options['routes'][$name] = array(
+                        'type' => 'chain',
+                        'options' => array(
+                            'routes' => array($defaultHostnameRoute, $route),
+                            'route_plugins' => $routePluginManager
+                        )
+                    );
+                }
+            }
+
+            unset($options['default_hostname_route']);
+        }
+
         $instance = parent::factory($options);
 
         if (isset($options['prototypes'])) {
--- a/tests/ZendTest/Mvc/Router/Http/TreeRouteStackTest.php
+++ b/tests/ZendTest/Mvc/Router/Http/TreeRouteStackTest.php
@@ -18,6 +18,7 @@ use Zend\Uri\Http as HttpUri;
 use Zend\Mvc\Router\Http\TreeRouteStack;
 use Zend\Mvc\Router\Http\Hostname;
 use ZendTest\Mvc\Router\FactoryTester;
+use Zend\Mvc\Router\RoutePluginManager;
 
 class TreeRouteStackTest extends TestCase
 {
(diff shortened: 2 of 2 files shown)


weierophinney:
I'm going to paint a scenario here.

Let's say we have three top-level routes. One defines a hostname route of "blog.example.com"; one is a normal path-based route, no hostname provided; and another has a hostname of "api.example.com".

One thing to note is that the second route is valid _regardless of the hostname_. It will resolve if any of www.example.com, blog.example.com, or api.example.com are matched.

When you do an `assemble()` operation for the second route, the URI generated is _only the path_. The only reason it resolves to a different hostname is because the browser will fill in the authority and scheme for you; a path is considered relative to these. Because the route is valid on any of those hostnames, it will then resolve based on the current hostname.

I'd argue that if you want to resolve the route based on a given hostname, what you're calling a "default hostname" in this PR, you should make the route a child under a hostname route for that default.

```php
return [
    'router' => [ 'routes' => [ 
        'default' => [ 
            'type' => 'hostname',
            'may_terminate' => false,
            'options' => [
                'route' => 'www.example.com',
            ],
            'child_routes' => [
                /* ... user route here ... */
            ],
        ],
    ] ]
];
```

Of course, taking this approach breaks functionality when you're moving between a dev/staging/qa box and production! Except it doesn't: you can define the base hostname in one of your `config/autoload/*.local.php` files:

```php
return [
    'router' => [ 'routes' => [ 'default' => [ 'options' => [
        'route' => 'example.dev',
    ] ] ] ]
]; 
```

Going this route, we could potentially change the hostname support such that a value of '0.0.0.0' or '0' _always_ resolves, which would simplify the "local" definition.

@nikolaposa Can you try this approach, and comment?

nikolaposa:
@weierophinney One "drawback" with the approach of having that top-level "default" route is that you will be forced to assemble URLs by specifying that extra prefix in the route name, for example: `'default/user/view'` instead of `'user/view'`. Of course, in certain cases, like the one you specified, with three top-level routes in the application, this should not be considered as a drawback, because it is probably desirable to be explicit when assembling URLs.

Idea behind my solution was to avoid that additional level in the route naming in case of default, global routes and then to enable that "behind the scene magic" for preventing possible collisions with hostname-specific routes. Also, potential changes would be efficient/less painful, because if for example those "/user/*" URLs from my introductory example have to be moved to a "user.example.com" subdomain, the only thing that should be modified is the routes configuration itself.

But the more I think about it, I realize that it does not bring some greater benefits (besides those shorter route names) as such use cases can be achieved using techniques that you have demonstrated. Perhaps one more issue with this proposal is the fact that the Router will become coupled to and aware of the concrete Hostname route implementation.

If you think that this feature however want be so useful and if no one else has something to suggest, I respect such attitude, this PR should eventually be closed and would not have any problem with that.

weierophinney:
I'll leave the final call to @DASPRiD (assuming he can review before we're ready to do feature freeze on 2.4), but my vote is increasingly on limiting new features in order to allow us to better solidify and maintain the core of the framework, particularly if an approach using the existing functionality exists.

Thanks for the PR, @nikolaposa, and for trying out the approach I suggested.

DASPRiD:
For once, the assembling part can easily be solved by simply setting a baseUrl on the rotuer with your default hostname; This requires no code-change, but only works if you force canonical URIs. However, by changing the assemble() method in the router to check the current hostname (or the hostname in the resulting route, if defined) against one defined in the baseUrl, this could be solved.

The patch you supplied here assumes that one defines the hostname route in the top level, which does not have to be the case, as the hostname route can be in any level. Thus I have to decline this patch.

nikolaposa:
@DASPRiD Good point, I guess you're right. I suppose this PR should be closed. 

Once again, thanks to all the participants for the feedback.

PR #6774, opened 2014-10-17
Fixes #6769 : remove form element should remove the inputFilter

### LINKED ISSUE #6769: Removing an element from a form doesn't remove it from the input filter
Not really sure what to do about this one, however we've encountered an issue whereby removing a field from a form which was set as required causes a validation error when the form is validated: the field is removed from the form, but left in the input filter.

Any ideas on the best fix for this?

samsonasik:
Fixes #6769

### DIFF
--- a/library/Zend/Form/Form.php
+++ b/library/Zend/Form/Form.php
@@ -884,4 +884,21 @@ class Form extends Fieldset implements FormInterface
 
         return $values;
     }
+
+    /**
+     * Remove a named element or fieldset with optional to remove the inputFilter
+     *
+     * @param  string $elementOrFieldset
+     * @param  bool   $removeFromInputFilter
+     *
+     * @return self
+     */
+    public function remove($elementOrFieldset, $removeFromInputFilter = false)
+    {
+        if ($this->has($elementOrFieldset) && $removeFromInputFilter) {
+            $this->getInputFilter()->remove($elementOrFieldset);
+        }
+
+        parent::remove($elementOrFieldset);
+    }
 }
--- a/library/Zend/Form/Form.php
+++ b/library/Zend/Form/Form.php
@@ -12,6 +12,7 @@ namespace Zend\Form;
 use Traversable;
 use Zend\Form\Element\Collection;
 use Zend\InputFilter\CollectionInputFilter;
+use Zend\InputFilter\InputInterface;
 use Zend\InputFilter\InputFilter;
 use Zend\InputFilter\InputFilterAwareInterface;
 use Zend\InputFilter\InputFilterInterface;
@@ -497,6 +498,12 @@ class Form extends Fieldset implements FormInterface
             ));
         }
 
+        foreach ($filter->getInputs() as $input) {
+            if (method_exists($input, 'getName') && !$this->has($input->getName())) {
+                $filter->remove($input->getName());
+            }
+        }
+
         $filter->setData($this->data);
         $filter->setValidationGroup(InputFilterInterface::VALIDATE_ALL);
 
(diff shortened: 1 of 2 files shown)


weierophinney:
> perhaps only the validation process needs a slight change so only existing form fields are validated

I think this approach makes the most sense. Essentially, `Form::isValid()` would create a validation group based on the elements/fieldsets present, and then validate that validation group. Doing this:

- keeps the internal state of the input filter intact
- reduces coupling
- meets expectations from users with regards to what happens when an element is removed

@samsonasik Are you willing to rewrite the patch accordingly? If you can do so by 9 March 2015, I can include it in 2.4.

samsonasik:
@weierophinney i'm not sure i understand correctly. Did you mean it need check in <code>Form::isValid()</code> ?

weierophinney:
@samsonasik I'm saying that the work can be done within `Form::isValid()`. What you'd do in there is:

- Pull your input filter
- Iterate through your elements/fieldsets to build a list of items for a validation group
- Set the validation group on the input filter
- Validate the input filter

This would address the issue correctly, without introducing side effects.

samsonasik:
@weierophinney i can't reproduce it in Form::isValid() directly without rely-ing in Form::remove() ( save state ? ) because when element already removed, we can't get the element removed to match with inputfilter.

carnage:
You don't need to know if an element has been removed or not if you construct an input filter validation group from just the elements that are currently in the form, you magically exclude any elements that have been removed.

samsonasik:
@carnage ok, I've updated the code, is it ok now ?

samsonasik:
@carnage my update currently only resolve on remove() method, not on setValidationGroup(), could you send PR to my branch ? thanks...

carnage:
I've not got much time left for today - here's some psudo code:

function isValid() {
//existing code
if ($this->getValidationGroup() == self::VALIDATE_ALL) { 
$validationGroup = array();
foreach ($this->getElements() as $element) {
if ($element isAFieldset) {
//recurse into $element and add to $validationGroup
} else {
$validationGroup[] = $element->getName();
}
} else {
$validationGroup = $this->getValidationGroup();
}

$this->getFilter()->setValidationGroup($validationGroup);
return $this->getFilter()->isValid($data);
}

Does that roughtly make sense to you?

samsonasik:
seems I can't reproduce that right now.  as @weierophinney  said, last update should on 9 March 2015 , 
@carnage  you can ping me when you have time. thank you.

library/Zend/Form/Form.php
```diff
@@ -497,6 +497,12 @@ public function isValid()
             ));
         }
 
+        foreach ($filter->getInputs() as $input) {
+            if (method_exists($input, 'getName') && !$this->has($input->getName())) {
+                $filter->remove($input->getName());
```
Martin-P: There is no need to remove elements from the input filter, you only have to change the validation group. The input filter itself can remain intact.

library/Zend/Form/Form.php
```diff
@@ -497,6 +497,12 @@ public function isValid()
...
+            }
+        }
+
         $filter->setData($this->data);
         $filter->setValidationGroup(InputFilterInterface::VALIDATE_ALL);
```
Martin-P: This should be changed. Here you iterate through the form elements and if they exist you add them to the validation group.

carnage:
@samsonasik take a look at this:

https://github.com/carnage/zf2/commit/cfceeea07c34a2c812c5689b2b6dcb9def21d318

It does what needs to be done, however there are unit test failiures related to nested nested fieldsets. I can't figure out how I've created the validation group incorrectly for this scenario see what you think.

samsonasik:
@carnage I have followed https://github.com/carnage/zf2/commit/cfceeea07c34a2c812c5689b2b6dcb9def21d318 but it makes the test for removeElement() error when asserting Form::getInputFilter()::has() : 
```
2) ZendTest\Form\FormTest::testRemoveElementShouldRemoveTheInputFilter
Failed asserting that true is false.
```
and of course, the regression test on nested fieldset.

I'm closing it for now as I can't reproduce it.

PR #6784, opened 2014-10-19
Input filter should filter values before checking empty/required status

### LINKED ISSUE #6668: InputFilter: allow_empty should check filtered input
As a follow-up to my comments for #6133, I think the observed behavior for allow_empty is incorrect. The check is performed on unfiltered input, but the validator chain receives filtered input. This inconsistency can lead to erratic behavior if an attached filter alters input in a way that affects the definition of "emptiness".

For example, this happens with the StringTrim filter when raw input contains only whitespace. This is considered non-empty raw input, but the filter chain will produce an empty string. If allow_empty is TRUE, the validator chain still gets invoked which, depending on the particular validators, may cause validators to fail if these expect non-empty input. The validator chain should not be invoked in that case because empty input was explicitly declared valid.

If the StringTrim filter in the above example is followed by a Null filter, things can get really weird. Many validators (like StringLength) don't accept NULL input which is not a problem by itself, but causes the validator chain to fail when it should not have been invoked at all.

This may look like an edge case, but any filter that converts non-empty input to empty values or vice versa, or just produces a different kind of empty value (like empty string to NULL) can cause incorrect results.

Currently I have to work around the issue by reinventing the wheel: extend existing validators to re-check for empty values, or write a callback even for the most basic validations (see StringLength/Null example above).

All this hassle could be avoided by an additional emptiness check on the filter chain result - the same value that is passed to the validator chain. That would be the only behavior that works consistently and independently from the actual filter chain.

The continue_if_empty option has no effect on this behavior.

carnage:
Fixes https://github.com/zendframework/zf2/issues/6668

All the unit tests pass without any modifications, however this patch may introduce some BC breaks. Would be good to have several people review this to make sure the logic is sound.


carnage:
Superceeded by #7056

PR #6793, opened 2014-10-23
Added PROPFIND to valid methods

lkneschke:


weierophinney:
@lkneschke Still need test cases...

corneliusweiss:
with fromString($string, $allowCustomMethods = true) this change might be obsolete

GeeH:
New issue can be found at: https://github.com/zendframework/zend-http/issues/78

PR #6803, opened 2014-10-24
Fix extra types specification

fyrye:
Fix issue from Char/Varchar being copied as Text where the length was removed from Text.
Cause of issue: Specification from Char/Varchar to Text changed from 4 to 3 types
Result of issue: default would always be considered as self::TYPE_LITERAL

### DIFF
--- a/library/Zend/Db/Sql/Ddl/Column/Text.php
+++ b/library/Zend/Db/Sql/Ddl/Column/Text.php
@@ -33,7 +33,7 @@ class Text extends Column
         $spec   = $this->specification;
         $params = array();
 
-        $types    = array(self::TYPE_IDENTIFIER, self::TYPE_LITERAL);
+        $types    = array(self::TYPE_IDENTIFIER);
         $params[] = $this->name;
 
         $types[]  = self::TYPE_LITERAL;
--- a/tests/ZendTest/Db/Sql/Ddl/Column/TextTest.php
+++ b/tests/ZendTest/Db/Sql/Ddl/Column/TextTest.php
@@ -32,7 +32,7 @@ class TextTest extends \PHPUnit_Framework_TestCase
         
         $column->setDefault('default');
         $this->assertEquals(
-            array(array('%s TEXT %s %s', array('foo', 'NOT NULL', 'default'), array($column::TYPE_IDENTIFIER, $column::TYPE_LITERAL, $column::TYPE_VALUE))),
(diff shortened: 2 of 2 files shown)


fyrye:
@adamlundrigan 
As the issue description says, the Text object code is incorrect.
The specification expects 3 `$types` and `$params`, but is given 4 `$types` and 3 `$params` when `$this->default` is not null.
This is due to the extra value supplied to the `$types` array when `Text::getExpressionData()` is called.

Despite the fact that some DMS's (I believe the only exception is PostgreSQL) do not permit `TEXT` data types to have a default value.
The original issue is that `->setDefault('default');` would always be processed as a literal type rather than as a value type as the issue description also explains. 
This is due to the third value appended to the `$types` array being `self::TYPE_LITERAL`
Then when `$this->default` is not null, the fourth value appended to the `$types` array is `self::TYPE_VALUE`.

```php
$text = new Column\Text('column'); 
$text->setDefault('default');
```
The resulting SQL before would be executed as 
```SQL
ALTER TABLE `table` ADD COLUMN `column` TEXT NOT NULL DEFAULT default;
```
Instead of
```SQL
ALTER TABLE `table` ADD COLUMN `column` TEXT NOT NULL DEFAULT 'default';
```

fyrye:
Here's an example output of the code's types and params: http://ideone.com/7nFHI6

weierophinney:
I just attempted merging this to the develop branch, and there's massive conflicts... as there has much that has changed in the DDL subcomponent.

This is what I have: https://github.com/weierophinney/zf2/commit/746948fede285b2caa87e47a1354c66a44da10ac

Tests for DDL pass, but in reviewing the code present in the current develop branch, I'm half-thinking that the issue you're addressing in this PR is already addressed in whatever refactoring has occurred there already.

@fyrye can you please evaluate the current develop branch and see if you're seeing the same issues? And please comment back to indicate yes or no, so I know whether or not to continue with this merge.

fyrye:
This Master branch update shouldn't be merged into the Develop branch.

As for the issue existing in Develop, it has been resolved in the Develop branch as of 2014-11-17 by @Ocramius after I had opened this issue.
The develop branch now uses an AbstractColumnLength spec which uses a set specification, where the Master branch uses protected property specs which was out of sync for the Text object's expressionData.

weierophinney:
@fyrye Since the next version will be 2.4.0, and you indicate that this is resolved on develop, I'm going to close.

fyrye:
@weierophinney Shouldn't this still be included as a bug fix for 2.3.x as a non-BC update, since 2.4 would be a new minor version and still considered a bug in 2.3.x?
Or does ZF2 not provide bug fixes for minor versions?

If so should the pull request be against a tag instead of a branch?

PR #6827, opened 2014-11-01
Fix inability to translate to languages which don't have plural forms

hissy:
Some languages (e.g. Japanese, Chinese, etc.) does not have plural forms. With these languages, `getTranslatedMessage` function returns string value instead of array, even in `translatePlural` function. However, `translatePlural` function can not treat a string value. This issue is critical for us.


weierophinney:
Merged to master for release with 2.4.

hissy:
Thanks! :smile:

PR #6836, opened 2014-11-03
[Zend\ServiceManager] Add alternatives services proposals when service not found is thrown

blanchonvincent:
This idea comes from symfony, it would be good to add some details about services that are close (about name). This point can help developer to save time during the debug.


weierophinney:
I'm with @Ocramius on this one; let's implement it as an abstract factory. Developers can then enable/disable at will, and it doesn't require changes to the service manager itself.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-servicemanager/issues/129

PR #6849, opened 2014-11-06
Db/Sql/Predicate/Expression: Fix method argument handling

Pittiplatsch:
- Fix method argument handling of Predicate/Expression to allow single non-scalar values (e.g. null, Predicate instances) as only parameter
- Fix typo in Predicate/Predicate

### DIFF
--- a/library/Zend/Db/Sql/Predicate/Expression.php
+++ b/library/Zend/Db/Sql/Predicate/Expression.php
@@ -30,13 +30,11 @@ class Expression extends BaseExpression implements PredicateInterface
             $this->setParameters($valueParameter);
         } else {
             $argNum = func_num_args();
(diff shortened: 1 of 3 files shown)


darpla:
The changes introduced a BC break! The $valueParameter argument in the constructor has the default value null, which in the old code meant that the provided expression has no parameters. Now the new code treats this null as a query parameter and therefore the old queries fail due to this!
This means that such constructs now will fail with error 'The number of replacements in the expression does not match the number of parameters' during query preparation, because the class thinks that there is an unused parameter:

```
$select = $this->tableGateway->getSql()->select();
$select->where->expression('DATE(close_date - INTERVAL reminder_days DAY) = DATE(NOW())');
```

P.S. The Sql\Expression class handles this case still the old way - as it should, without BC break.

PR #6852, opened 2014-11-07
Fix problem of decimals always visible when using currencyFormat view helper

Gawel1:
Apply comment visible in http://php.net/manual/en/numberformatter.formatcurrency.php :

When you want to format currency's without sub units and the currency is not the one used by the given locale you need to set the currency code before as TextAttribute _BEFORE_ setting the NumberFormatter::FRACTION_DIGITS

```php
$fmt = new NumberFormatter('en_US', NumberFormatter::CURRENCY);
$fmt->setTextAttribute(NumberFormatter::CURRENCY_CODE, 'EUR');
$fmt->setAttribute(NumberFormatter::FRACTION_DIGITS, 0);
$fmt->formatCurrency(100, 'EUR');
```


GeeH:
New issue can be found at: https://github.com/zendframework/zend-i18n/issues/52

PR #6888, opened 2014-11-18
Introduce CountryCode validator (rework of #4142)

adamlundrigan:
This is an update to #4142; I rebased the original branch and updated to include feedback on the initial prototype: 
- Bump copyright year + CS fixes
- Remove unused resource bundle files
- Removed hard-coded country names
- Use `static` instead of `self` in `Zend\I18n\CountryDb`

There is one outstanding issue that I'm unsure of:  country codes are not locale-specific, so should this code (or at least the Validator class) be moved out of the I18n namespace?  If so, where should the resource class (`Zend\I18n\CountryDb`) be placed?

Thanks to @mwillbanks for the bulk of the work on this and also to those who left their feedback in the original PR (#4142)


weierophinney:
@adamlundrigan and @mwillbanks — totally agreed; less functionality for us to support in the framework is a good thing, and this is indeed one of the motivators behind a module ecosystem (extending framework functionality).

Closing! Thanks for the work and the discussions!

PR #6894, opened 2014-11-18
standard hydrator idea

ojhaujjwal:
I created this hydrator and named it as `StandardHydrator`. This is just an idea. This name can be changed if you guys like the idea.

``` php
class Album
{
    public function getId()
    {
        return 34;
    }
}

class Artist
{
    public function getName()
    {
        return 'David';
    }
}


$albumHydrator = new Zend\Stdlib\Hydrator\ClassMethods;
$artistHydrator = new Zend\Stdlib\Hydrator\ClassMethods;

$hydrators = new Zend\Stdlib\Hydrator\HydratorPluginManager;
$hydrators->setService('Album', $albumHydrator);
$hydrators->setService('Artist', $artistHydrator);

$standardHydrator = new Zend\Stdlib\Hydrator\StandardHydrator($hydrators);

$standardHydrator->extract(new Album); // returns ['id' => 34];
$standardHydrator->extract(new Artist); // returns ['name' => 'David'];
```

This detects the hydrator for each object. I think that this hydrator can be pretty useful.

### DIFF
--- a/library/Zend/Stdlib/Hydrator/MappingHydrator.php
+++ b/library/Zend/Stdlib/Hydrator/MappingHydrator.php
@@ -50,15 +50,9 @@ class MappingHydrator implements HydratorInterface
      *
      * @param  object                             $object
      * @return HydratorInterface
-     * @throws Exception\InvalidArgumentException
      */
     protected function getHydrator($object)
     {
-        $class = get_class($object);
-        if (!$this->hydrators->has($class)) {
-            throw new Exception\InvalidArgumentException(sprintf('Hydrator service does not exists for class %s', $class));
-        }
-
-        return $this->hydrators->get($class);
+        return $this->hydrators->get(get_class($object));
     }
 }
--- a/library/Zend/Stdlib/Hydrator/MappingHydrator.php
+++ b/library/Zend/Stdlib/Hydrator/MappingHydrator.php
@@ -3,7 +3,7 @@
  * Zend Framework (http://framework.zend.com/)
  *
  * @link      http://github.com/zendframework/zf2 for the canonical source repository
- * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
@@ -12,7 +12,7 @@ namespace Zend\Stdlib\Hydrator;
 use Zend\ServiceManager\ServiceLocatorInterface;
 use Zend\Stdlib\Exception;
 
-class StandardHydrator implements HydratorInterface
+class MappingHydrator implements HydratorInterface
 {
     /**
      * @var ServiceLocatorInterface
--- a/library/Zend/Stdlib/Hydrator/MappingHydratorFactory.php
+++ b/library/Zend/Stdlib/Hydrator/MappingHydratorFactory.php
@@ -7,16 +7,15 @@
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
-namespace Zend\Mvc\Service;
+namespace Zend\Stdlib\Hydrator;
 
 use Zend\ServiceManager\FactoryInterface;
 use Zend\ServiceManager\ServiceLocatorInterface;
-use Zend\Stdlib\Hydrator\MappingHydrator;
 
 class MappingHydratorFactory implements FactoryInterface
 {
     /**
-     * Creates StandardHydrator
+     * Creates MappingHydrator
      *
      * @param  ServiceLocatorInterface $serviceLocator
      * @return MappingHydrator
--- a/tests/ZendTest/Stdlib/Hydrator/MappingHydratorFactoryTest.php
+++ b/tests/ZendTest/Stdlib/Hydrator/MappingHydratorFactoryTest.php
@@ -7,9 +7,9 @@
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
 
-namespace ZendTest\Mvc\Service;
+namespace ZendTest\Stdlib\Hydrator;
 
-use Zend\Mvc\Service\MappingHydratorFactory;
+use Zend\Stdlib\Hydrator\MappingHydratorFactory;
 
 class MappingHydratorFactoryTest extends \PHPUnit_Framework_TestCase
 {
--- a/tests/ZendTest/Stdlib/Hydrator/MappingHydratorTest.php
+++ b/tests/ZendTest/Stdlib/Hydrator/MappingHydratorTest.php
@@ -9,18 +9,18 @@
 
 namespace ZendTest\Stdlib\Hydrator;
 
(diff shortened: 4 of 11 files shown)


ojhaujjwal:
Any more comments?

weierophinney:
This could actually replace the functionality we have in our `metadata_map` inside zf-hal. The key difference zf-hal proposes is that it also does some lookups based on inheritance tree, IIRC.

I, too, am not a fan of the verbiage `StandardHydrator`, and think something along the lines of `SelectingHydrator` or `MappingHydrator` would be more self-describing.

@ojhaujjwal :+1: 

- Please rebase against the current develop branch
- Rename to one of the above options

and I'll include this in 2.4 (and update zf-hal to make use of it!).

macnibblet:
@weierophinney can you please explain why this component belongs in mvc ? I don't see a reason to include it just so something can be removed from a apigility component ?

sandrokeil:
@weierophinney What's with the name DiscoveryHydrator?

ojhaujjwal:
`MappingHydrator` is better IMO!

ojhaujjwal:
@weierophinney rebasing part is done! :D

weierophinney:
> @weierophinney can you please explain why this component belongs in mvc ?

I missed that the factory was in the MVC component - it should be under the `Zend\Stdlib\Hydrator` namespace. @ojhaujjwal one more change to make...

> I don't see a reason to include it just so something can be removed from a apigility component ?

@macnibblet — It's not just Apigility; I've actually seen a number of people wanting to use the hydrator mapping aspect of zf-hal for other purposes, but it's not separated. This would provide that separation. It would also allow us to use the mapping aspect for vanilla JSON responses, for which we don't have a serialization strategy (other than JsonSerializable) in Apigility currently due to the aforementioned coupling in zf-hal — but in a way that could be put into a specialized JSON renderer.

> What's with the name DiscoveryHydrator?

Um, @sandrokeil, I don't know... you're the first to bring that up. Are you suggesting it as a name for the hydrator?

sandrokeil:
@weierophinney Yes, it's another suggestion for the hydrator name.

library/Zend/Stdlib/Hydrator/MappingHydrator.php
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @param  object                             $object
+     * @return HydratorInterface
+     * @throws Exception\InvalidArgumentException
+     */
+    protected function getHydrator($object)
+    {
+        $class = get_class($object);
+        if (!$this->hydrators->has($class)) {
+            throw new Exception\InvalidArgumentException(sprintf('Hydrator service does not exists for class %s', $class));
```
stefanotorresi: I think this check should be removed, as it can lead to unpredictable behaviour as explained in zfcampus/zf-hal#86

The plugin validation is already done by the plugin manager, calling `has` on a pluginmanager with `$autoAddInvokableClass` on by default can lead to inconsistent results.

stefanotorresi:
As a name I would personally like `DelegateHydrator`, as it follows the delegator pattern quite closely.

library/Zend/Stdlib/Hydrator/MappingHydrator.php
weierophinney: Good point; @ojhaujjwal — please update accordingly. Essentially, we can rely on the fact that the hydrator plugin manager will raise an exception here.

weierophinney:
> As a name I would personally something like DelegateHydrator, DelegatorHydrator or HydratorDelegator, as it follows the delegator pattern quite closely.

I was suggesting "map", as I was thinking of the feature with regards to the `metadata_map` of zf-hal... but that map is for more than just hydrators; in fact, hydrators are a _part_ of that map.

I'm personally wavering between "Discovery" and "Delegating"; I'll choose one during merge.

jaapio:
+1 for Delegating, since the actual hydration is deletaged to another hydrator

library/Zend/Stdlib/Hydrator/MappingHydrator.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function extract($object)
+    {
+        return $this->getHydrator($object)->extract($object);
+    }
+
+    /**
+     * Gets hydrator of an object
+     *
+     * @param  object                             $object
```
malukenho: Trim spaces

library/Zend/Stdlib/Hydrator/MappingHydratorFactory.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
malukenho: `2005-2015`

tests/ZendTest/Stdlib/Hydrator/MappingHydratorFactoryTest.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
malukenho: Same above

tests/ZendTest/Stdlib/Hydrator/MappingHydratorTest.php
```diff
@@ -0,0 +1,89 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
malukenho: Same above

weierophinney:
I'm reviewing now, and have noticed that the patch still contains additions to the MVC `ServiceListenerFactory`. I think the more appropriate route is to expose the new hydrator via the HydratorPluginManager, and feed it the parent service locator if present. I will attempt to make that adaptation during merge.

weierophinney:
Merged for release in 2.4.

Renamed the new hydrator to `DelegatingHydrator`, per consensus in the comments. Additionally, you now pull it from the `HydratorPluginManager` (the `HydratorManager` service) instead of the application ServiceManager instance.

Usage is now:

```php
$hydrators = $services->get('HydratorManager');
$delegating = $hydrators->get('DelegatingHydrator');
$albumHydrator = new Zend\Stdlib\Hydrator\ClassMethods;
$artistHydrator = new Zend\Stdlib\Hydrator\ClassMethods;

$hydrators = new Zend\Stdlib\Hydrator\HydratorPluginManager;
$hydrators->setService('Album', $albumHydrator);
$hydrators->setService('Artist', $artistHydrator);

$array = $delegating->extract(new Artist);
$artist = $delegating->hydrate($data, new Artist);
```

@ojhaujjwal Please also submit documentation to https://github.com/zendframework/zf2-documentation :smile:

PR #6906, opened 2014-11-21
Use session_status when available to better detect if the session exists

mtorromeo:
In php 5.4.0 the [session_status()](http://nl1.php.net/manual/en/function.session-status.php) function was implemented which allows to determine if the session was already started without weird checks.

This patch makes `SessionManager::sessionExists` use this function when available and falls back to the old behaviour on php<5.4

### DIFF
--- a/library/Zend/Session/SessionManager.php
+++ b/library/Zend/Session/SessionManager.php
@@ -58,9 +58,15 @@ class SessionManager extends AbstractManager
      */
     public function sessionExists()
     {
-        $sid = defined('SID') ? constant('SID') : false;
-        if ($sid !== false && $this->getId()) {
-            return true;
+        if (function_exists('session_status')) {
+            if (session_status() === PHP_SESSION_ACTIVE) {
+                return true;
+            }
+        } else {
+            $sid = defined('SID') ? constant('SID') : false;
+            if ($sid !== false && $this->getId()) {
+                return true;
+            }
         }
         if (headers_sent()) {
             return true;


library/Zend/Session/SessionManager.php
```diff
@@ -58,9 +58,15 @@ public function __construct(Config\ConfigInterface $config = null, Storage\Stora
      */
     public function sessionExists()
     {
-        $sid = defined('SID') ? constant('SID') : false;
-        if ($sid !== false && $this->getId()) {
-            return true;
+        if (function_exists('session_status')) {
```
weierophinney: We could potentially define the constant during the test, and lazy load a file that defines the function, when under PHP 5.3... but ouch. That said, the path is covered when in 5.4+ so, we're good there. The bigger issue is the new test failure. I'll comment on that separately.

weierophinney:
> One test fails with this change now

Interestingly, Travis shows **2** failing tests, both in the SessionManager: `testSessionExistsReturnsTrueWhenSessionStartedThenWritten` and `testCallingWriteCloseShouldNotAlterSessionExistsStatus`. In both cases, the _documented_ expected behavior is that `sessionExists() === true` after a `writeClose()` operation.

I observed the same failures locally.

So, I read the [session_status documentation](http://php.net/session_status). What I'm seeing there is that `session_status()` has three states:

- `PHP_SESSION_DISABLED` if sessions are disabled. In this case, `sessionExists()` should return false, always.
- `PHP_SESSION_ACTIVE` if sessions are enabled, and a session exists.
- `PHP_SESSION_NONE` if sessions are enabled, but no session exists.

The problem is that the documentation does not define what "session exists" means. I'm pretty sure it means a session identifier was found for the incoming request and a matching session was found; the question is whether the flag toggles to `PHP_SESSION_NONE` on a `session_write_close()` operation.

From my own testing, it appears that `session_status` only returns `PHP_SESSION_ACTIVE` *after* `session_start` has been called, regardless of whether or not a session ID was present in the request. Additionally, after a `session_write_close` operation, PHP *always*  toggles the flag to `PHP_SESSION_NONE`. This means that using `session_status()` in this context introduces a behavior change when going from 5.3 to 5.4+, as we currently have the method return `true` if a session ID was found at any point in the request.

As such, I'm closing this without merging, as it changes the behavior of the functionality.

Issue #4714, opened 2013-06-24
Zend\Di\Di::get() repeats the instance using differents params


sebalb:
Hello, I found something which I consider is strange in Zend\Di, I would like to know if someone could tell me if this is the expected result or not.

*(I am working with the documentation examples)*

```
<?php
namespace MyLibrary
{
    class DbAdapter
    {
        protected $username = null;
        protected $password = null;
        public function __construct($username, $password)
        {
            $this->username = $username;
            $this->password = $password;
        }
    }
}

namespace MyMovieApp
{
    class MovieFinder
    {
        protected $dbAdapter = null;
        public function __construct(\MyLibrary\DbAdapter $dbAdapter)
        {
            $this->dbAdapter = $dbAdapter;
        }
    }

    class MovieLister
    {
        protected $movieFinder = null;
        public function __construct(MovieFinder $movieFinder)
        {
            $this->movieFinder = $movieFinder;
        }
    }
}    

namespace {
	$di = new \Zend\Di\Di();
	$im = $di->instanceManager();
	
	// add alias for short naming
	$im->addAlias('movielister', 'MyMovieApp\MovieLister');
	
	// add aliases for specific instances
	$im->addAlias('dbadapter-readonly', 'MyLibrary\DbAdapter', array(
	    'username' => 'user1',
	    'password' => 'pass1',
	));
	$im->addAlias('dbadapter-readwrite', 'MyLibrary\DbAdapter', array(
	    'username' => 'user2',
	    'password' => 'pass2',
	));
	
	// set a default type to use, pointing to an alias
	$im->addTypePreference('MyLibrary\DbAdapter', 'dbadapter-readonly');
	
	$movieListerRead = $di->get('MyMovieApp\MovieLister');
	$movieListerReadWrite = $di->get('MyMovieApp\MovieLister', array('dbAdapter' => 'dbadapter-readwrite'));			
	
    var_dump($movieListerRead); //user1 / pass1
    var_dump($movieListerReadWrite); // user1 / pass1 ??
}
```

## The output:

```
object(MyMovieApp\MovieLister)[235]
  protected 'movieFinder' => 
    object(MyMovieApp\MovieFinder)[234]
      protected 'dbAdapter' => 
        object(MyLibrary\DbAdapter)[233]
          protected 'username' => string 'user1' (length=5)
          protected 'password' => string 'pass1' (length=5)

object(MyMovieApp\MovieLister)[235]
  protected 'movieFinder' => 
[...]

PR #6920, opened 2014-11-23
di/instanceManager ignores call time parameters of array

noopable:
relative #4714

InstanceManager::sharedInstancesWithParams with Hash based management ignores entries of array.This causes #4717.

### DIFF
--- a/library/Zend/Di/InstanceManager.php
+++ b/library/Zend/Di/InstanceManager.php
@@ -502,7 +502,12 @@ class InstanceManager /* implements InstanceManagerInterface */
                     $hashValue .= $param . '|';
                     break;
                 case 'array':
-                    $hashValue .= 'Array|';
+                    try {
+                        $hashValue .= serialize($param) . '|';
+                    } catch (\Exception $ex) {
+                        //not serializable
+                        $hashValue .= 'Array|';
+                    }
                     break;
                 case 'resource':
                     $hashValue .= 'resource|';
--- a/library/Zend/Di/Di.php
+++ b/library/Zend/Di/Di.php
@@ -238,7 +238,7 @@ class Di implements DependencyInjectionInterface
                 return $im->getSharedInstanceWithParameters(null, array(), $fastHash);
             }
         } elseif ($im->hasSharedInstance($name)) {
-                array_pop($this->instanceContext);
+            array_pop($this->instanceContext);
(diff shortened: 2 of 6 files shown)


library/Zend/Di/InstanceManager.php
```diff
@@ -502,7 +502,12 @@ protected function createHashForValues($classOrAlias, $paramValues)
                     $hashValue .= $param . '|';
                     break;
                 case 'array':
-                    $hashValue .= 'Array|';
+                    try {
+                        $hashValue .= serialize($param) . '|';
+                    } catch (\Exception $ex) {
```
weierophinney: @noopable `serialize()` doesn't raise an exception, it raises a warning, which you cannot catch via try/catch. `Zend\Stdlib\ErrorHandler` provides a mechanism for shimming in a context-specific error handler from which you can then recover or raise warnings/exceptions/etc. Just wrap the call with:

- [ErrorHandler::start](https://github.com/zendframework/zf2/blob/master/library/Zend/Stdlib/ErrorHandler.php#L52)
- [ErrorHandler::stop](https://github.com/zendframework/zf2/blob/master/library/Zend/Stdlib/ErrorHandler.php#L68)

noopable: @weierophinney 
really?
```php
serialize(array(function(){}));
```
I did it and I got below. 
```
Exception: Serialization of 'Closure' is not allowed in Command line code on line 1
```

Can ErrorHandler catch this ?

weierophinney: As I've observed, `serialize()` actually has two behaviors:

- If an object cannot be serialized (e.g., `__sleep()` or `serialize()` returns an invalid structure), it raises an `E_NOTICE`.
- If an object raises an exception during serialization — as `Closure` does — that exception bubbles up.

As such, you need a strategy that combines the two approaches.

noopable: @weierophinney 
Thank you.
I got it.

Even though the serialization with suppressing Exception is not perfect, I had thought it is a pragmatic solution because optimistic hash is effective in most cases.

By the way, 
 If it is important to deal with both E_NOTICE and Exception,
I think var_export is more simple to get a hash of array than serialize.
Is it ok to use var_export?

PR #6925, opened 2014-11-25
Introducing a password handler authored by @DASPRiD 

awartoft:
Whats left be done
- [ ] What key to use for the `Zend\Crypt\Password\HandlerManager` 
- [ ] Integrate the HandlerManager into the mvc service listener factory
- [ ] Documentation

### DIFF
--- a/library/Zend/Crypt/Password/Algorithm/Bcrypt.php
+++ b/library/Zend/Crypt/Password/Algorithm/Bcrypt.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Crypt\Password\Algorithm;
+
+use Zend\Crypt\Password\Bcrypt as BcryptHasher;
+use Zend\Crypt\Password\HandlerInterface;
+
+/**
+ * Password hashing handler backed by Bcrypt.
+ */
+class Bcrypt implements HandlerInterface
+{
+    /**
+     * @var BcryptHasher
+     */
+    protected $backend;
+
+    /**
+     * supports(): defined by HandlerInterface.
+     *
+     * @see    HandlerInterface::supports()
+     * @param  string $hash
+     * @return boolean
+     */
+    public function supports($hash)
+    {
+        return (bool) preg_match('(^\$2[axy]?\$\d{2}\$[./a-zA-Z0-9]{53}$)', $hash);
+    }
+
+    /**
+     * hash(): defined by HandlerInterface.
+     *
+     * @see    HandlerInterface::hash()
+     * @param  string $password
+     * @return string
+     */
+    public function hash($password)
+    {
+        return $this->getBackend()->create($password);
+    }
+
+    /**
+     * compare(): defined by HandlerInterface.
+     *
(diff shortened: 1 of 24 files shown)


library/Zend/Crypt/Password/Algorithm/Bcrypt.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        return $this->backend;
+    }
+
+    /**
+     * @param  BcryptHasher $backend
+     * @return Bcrypt
+     */
+    public function setBackend(BcryptHasher $backend)
```
ojhaujjwal: IMO Setter is not needed. Constructor will do!

ojhaujjwal:
I like the overall idea. :+1:

library/Zend/Crypt/Password/HandlerInterface.php
```diff
@@ -0,0 +1,55 @@
...
+     * @param  string $password
+     * @param  string $hash
+     * @return boolean
+     */
+    public function compare($password, $hash);
+
+    /**
+     * Determines whether the hash should be re-hashed.
+     *
+     * This is used primarily by the authentication adapter, which will verify
```
DASPRiD: This comment is pretty much tied to BaconUser; should be updated.

padraic:
Note to monitor https://github.com/zendframework/zf2/pull/7141 which is effectively proposing a new default (a minor twist on plain bcrypt to preserve entropy from >72 byte passwords).

weierophinney:
@macnibblet What remains to be done? Can this be ready for 2.4?

macnibblet:
@weierophinney Everything that is listed in the task list, if i don't manage to look into it on sunday i'm doubtful that I will have time to finish it before the 2.4 release.

weierophinney:
@macnibblet I'll give you until the end of the first week of March to get code ready. :)

malukenho:
@macnibblet just to remember /cc @weierophinney

moderndeveloperllc:
@macnibblet I'm concerned that the MCF identifiers that you are using for SHA1 and MD5 are not the standard identifiers that are used in the standard Modular Crypt Format. `$1$` is standard for MD5. SHA1 is not 'standard' (mainly because it's not strong enough), but I have seen a few implementations that use `$sha1`. At least adding these to the checks with what you have will help developers who have to integrate with other languages that use the standard C `crypt()` library.

DASPRiD:
@moderndeveloperllc do you have a reference for the $1$ marker? Apart from that I'd agree to use those instead. Also we may want to remove the "simple-" part from the classes. No clue why I had chosen those names.

moderndeveloperllc:
@DASPRiD The `$1$` is from the [`crypt()` library docs](https://www.freebsd.org/cgi/man.cgi?query=crypt(3)). It also lists `$5$` for SHA256 and `$6$` for SHA512 along with the $2$ series for bcrypt.

I might actually take back what I've said in my previous comment, because `$1$` usually refers to an md5 where you are storing the salt - kinda like: `$1$fauihdfaoihfad$caeioanceoi039rasi09`. The code in these files are looking for a plain - single round MD5 with nothing but the hashed value itself stored. In the real world, would these even have an MCF?

FYI, The best explanation/reference I found was with Python's [PassLib documentation](http://pythonhosted.org/passlib/modular_crypt_format.html). As far as I can tell, there is no "standard" in that there are no IEEE or RFC standards out there. Most links I've seen refer to the PassLib docs. For instance, in the [scrypt module I built](https://bitbucket.org/moderndeveloperllc/zf2-scrypt), I decided to use `$s0$` because the leading [Java scrypt library](https://github.com/wg/scrypt) used that.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-crypt/issues/30

PR #6931, opened 2014-11-26
Added consts for left and right outer joins

AGmakonts:
Currently you have to pass string and manually specify the join type if it's not included in constants.

### DIFF
--- a/library/Zend/Db/Sql/Select.php
+++ b/library/Zend/Db/Sql/Select.php
@@ -44,6 +44,10 @@ class Select extends AbstractSql implements SqlInterface, PreparableSqlInterface
     const JOIN_OUTER = 'outer';
     const JOIN_LEFT = 'left';
     const JOIN_RIGHT = 'right';
+    const JOIN_RIGHT_OUTER = 'outer right';
+    const JOIN_LEFT_OUTER  = 'outer left';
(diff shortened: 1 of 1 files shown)


library/Zend/Db/Sql/Select.php
```diff
@@ -44,6 +44,10 @@ class Select extends AbstractSql implements SqlInterface, PreparableSqlInterface
     const JOIN_OUTER = 'outer';
     const JOIN_LEFT = 'left';
     const JOIN_RIGHT = 'right';
+    const JOIN_RIGHT_OUTER = 'outer right';
+    const JOIN_LEFT_OUTER  = 'outer left';
+    const JOIN_RIGHT_INNER = 'inner right';
+    const JOIN_LEFT_INNER = 'inner left';
```
weierophinney: Not to be pedantic, but... considering the actual word order, shouldn't these be:

- `JOIN_OUTER_RIGHT`
- `JOIN_INNER_LEFT`
- etc.

? Right now, the constant names do not mirror the actual usage, which could be confusing, particularly if you were to use the constant value elsewhere.

AGmakonts:
Names rearranged :) also I have removed those inners mentioned by @Pittiplatsch

weierophinney:

Issue #6960, opened 2014-12-05
Whitelist Filter


gsomoza:
I'm using a whitelist filter for one of my projects and was wondering if it would be an interesting addition to `Zend\Filter`.

The way it would work is it would accept a `whitelist` option as an array of values. If the value being filtered matches one of the values in the whitelist, the filter would return that value. Otherwise it would return `null`. Therefore it could be used in combination with the `Null` filter for example.

An application for this filter is Apigility: we could use it to filter the fields extracted with the hydrator. Only certain values would be allowed through.

We could also allow the filter to behave as either a blacklist or a whitelist.

If you're interested let me know and I'll create a PR. Otherwise I'd appreciate any suggestions related to any available alternatives that I may have missed.

GeeH:
This issue has been closed as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

PR #6962, opened 2014-12-05
New Whitelist / Blacklist Filter

gsomoza:
This new filter will behave either as a blacklist or a whitelist. If the value being filtered exists in the list of known values, the filter will return the value if its a whitelist or `null` if its a blacklist. The opposite will happen if value is not in the list.
### Whitelist

``` php
$whitelist = new \Zend\Filter\Whitelist(array(
    'list' => array('allowed-1', 'allowed-2')
));
echo $whitelist->filter('allowed-2');   // => 'allowed-2'
echo $whitelist->filter('not-allowed'); // => null
```
### Blacklist

``` php
$blacklist = new \Zend\Filter\Whitelist(array(
    'type' => 'blacklist', // optional, default is 'whitelist'
    'list' => array('not-allowed-1', 'not-allowed-2')
));
echo $blacklist->filter('allowed');        // => 'allowed'
echo $blacklist->filter('not-allowed-2');  // => null
```
### Notes
1. Maybe calling the filter class `Whitelist` might be counter-intuitive if the desired behavior is that of a blacklist. But I couldn't think of a better way to name the filter, so I'm open to suggestions:
   - Something like "AllowedList" would just be a synonym of "whitelist". 
   - "List" would just be too vague.
   - Separate filter classes seemed overkill.
2. I decided to allow only one type of configuration in the constructor (by array of options) just to keep it simple. If you'd like to allow for a more dynamic constructor (like a few other filters do) let me know.
3. Not sure what commit edbc12b is doing in this PR. Looks like its in `master` but not in `develop`.

See #6960 for more info.


library/Zend/Filter/Blacklist.php
```diff
@@ -0,0 +1,90 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2015 ;)

library/Zend/Filter/Whitelist.php
```diff
@@ -0,0 +1,91 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2015 ;)

tests/ZendTest/Filter/BlacklistTest.php
```diff
@@ -0,0 +1,150 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2015 ;)

tests/ZendTest/Filter/WhitelistTest.php
```diff
@@ -0,0 +1,150 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2015 ;)

weierophinney:
Merged -- now I need @gabrielsomoza to write up docs on the new filter and send a PR to https://github.com/zendframework/zf2-documentation :)

gabrielsomoza:
Sure, will do - any chance you could give me a deadline? (its easier for me if I can schedule it :) )

Ocramius:
@gabrielsomoza should be ~2 weeks

PR #6977, opened 2014-12-08
Update Select.php

alexandretaz:
Adding the case on _get() magic methods to retrieve the Order Param, this can be useful to manipulate the order param, checking it before send a Query


GeeH:
New issue can be found at: https://github.com/zendframework/zend-db/issues/141

PR #7003, opened 2014-12-09
Fixing prepended `NotEmpty` validator

gabrielsch:
When using `InputFilter` standalone (without dev dependencies), [this](https://github.com/zendframework/zf2/blob/master/library/Zend/InputFilter/Input.php#L375) is broken because [this line](https://github.com/zendframework/zf2/blob/master/library/Zend/Validator/ValidatorChain.php#L186) calls `ValidatorPluginManager`, and this class inherits `AbstractPluginManager`, that is `suggested` at [composer.json](https://github.com/zendframework/zf2/blob/master/library/Zend/InputFilter/composer.json#L26), not required.

Then the following error is thrown:
`PHP Fatal error:  Class 'Zend\ServiceManager\AbstractPluginManager' not found in /zf2/library/Zend/Validator/ValidatorPluginManager.php on line 16
`

I just changed the `prependByName` method (that uses PluginManager) to `prependValidator`, instantiating the validator by hand.

### DIFF
--- a/library/Zend/InputFilter/Input.php
+++ b/library/Zend/InputFilter/Input.php
@@ -372,7 +372,14 @@ class Input implements InputInterface, EmptyContextInterface
             }
         }
 
-        $chain->prependByName('NotEmpty', array(), true);
         $this->notEmptyValidator = true;
+
+        if (class_exists('Zend\ServiceManager\AbstractPluginManager')) {
+            $chain->prependByName('NotEmpty', array(), true);
+
+            return;
+        }
+
+        $chain->prependValidator(new NotEmpty(), true);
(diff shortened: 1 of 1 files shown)


weierophinney:
Suggestion: test to see if `class_exists('Zend\ServiceManager\AbstractPluginManager')`, and use `prependByName()` if it's present, and `prependValidator()` if not.

gabrielsch:
ok @weierophinney, I'll finish that soon

gabrielsch:
done @weierophinney , sorry for the delay :smile:

weierophinney:

PR #7034, opened 2014-12-13
MongoDB to MongoDb

marc-mabe:
Our coding style defines names to be in CamelCase but this one was MIXEDCase.
The cache adapter wasn't released until now so this is no BC break but the session save handler and the log writer i thing. But normally PHP classes and methods are case-insensitive so this can only be an issue with some autoloaders on case-sensitive fs.

I think it's OK for the next minor release.

@Ocramius thoughts ?

### DIFF
--- a/library/Zend/Cache/Storage/Adapter/MongoDb.php
+++ b/library/Zend/Cache/Storage/Adapter/MongoDb.php
@@ -17,7 +17,7 @@ use Zend\Cache\Exception;
 use Zend\Cache\Storage\Capabilities;
 use Zend\Cache\Storage\FlushableInterface;
 
-class MongoDB extends AbstractAdapter implements FlushableInterface
+class MongoDb extends AbstractAdapter implements FlushableInterface
 {
     /**
      * Has this instance be initialized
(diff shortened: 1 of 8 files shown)


weierophinney:
@Ocramius and @marc-mabe : this is a BC break, plain and simple. While many (most?) people will be using the SM in order to setup the log, cache, and session adapters, the problem is that anybody who was manually instantiating the classes previously now has broken code, as the classes will no longer resolve — and, frankly, for no good or apparent reason. CS should have been considered _before_ releasing these as stable; once we go stable, we have to live with the names until the next major revision.

I'm thinking we should revert this, and we can re-apply for 3.0.

I'm flagging this to revert.

weierophinney:
I've reverted this in the current develop branch.

@marc-mabe Please re-submit after we break the cache component into its own repo, and we can start considering v3 patches.

Ocramius:
@weierophinney why did you revert? This stuff was not yet in a stable tag.

weierophinney:
DAMMIT! I missed that aspect. Reverting the revert...

Issue #5860, opened 2014-02-23
Scrutinizer-ci integration


Ocramius:
I have started browsing the reports recently produced by @scrutinizer-ci about ZendFramework's codebase, and there's surprisingly not a lot to fix, but still a lot for a single person.

I have a reference build at https://scrutinizer-ci.com/g/Ocramius/zf2/inspections/3a265399-8787-4fe0-a045-88d36324af99

Things to be done:
 - enable @scrutinizer-ci runs for zendframework/zf2 ( @weierophinney that depends on your repo access ) - the fact that @scrutinizer-ci requires write-access may be a blocker.
 - start cleaning the heck out of it
 - ...
 - profit!

@samsonasik fancy starting from the [suggested patches](https://scrutinizer-ci.com/g/Ocramius/zf2/inspections/3a265399-8787-4fe0-a045-88d36324af99/patches)? (not all of them are right, so it needs someone to check docblocks of properties _AND_ methods)

```sh
curl 'https://scrutinizer-ci.com/g/Ocramius/zf2/inspections/3a265399-8787-4fe0-a045-88d36324af99.diff?s=093f1e46840d44cf7d2f60b96e37eae5e28321f3' \
    | git apply -
```

Ocramius:
This was already mostly done.

PR #7039, opened 2014-12-15
Cache: better fix for #5860 introduced in #5863

marc-mabe:
Cache: better fix for #5860 introduced in #5863


Ocramius:
Merged, thanks @marc-mabe!

PR #7051, opened 2014-12-16
CS: Ordering use statements

keradus:


keradus:
Should I do sth else in this PR @Ocramius ?

Ocramius:
@keradus I'd keep this sort of CS fix as last due to the huge amount of possible conflicts with PRs still open for [2.4.0](https://github.com/zendframework/zf2/milestones/2.4.0)

keradus:
I see. Thanks for info!

weierophinney:
Closing. The size of this is huge, and I think it will be easier to do this successfully when we do the component split following the 2.4 release.

keradus:
:(

Issue #5160, opened 2013-09-23
Make "quote_identifiers" option available on all Db platforms


hschletz:
The "quote_identifiers" platform option is evaluated only by the Oracle and IbmDb2 platforms. It would be useful with other platforms too because quoting identifiers has the nasty side effect of making them case sensitive. The default behavior of unconditional quoting may cause more problems than it solves for some databases.

GeeH:
This issue has been closed as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

PR #7059, opened 2014-12-19
Move common DB adapter platform functionality to a base class

hschletz:
The Zend\Db\Adapter\Platform classes are all implemented from scratch, resulting in a lot of duplicated code and making even simple changes needlessly complicated because they would need to be propagated across all platforms. This is unnecessary because most functionality is identical across multiple platforms.

One example would be a fix for #5160. The requested functionality is simple and portable, yet only implemented for some platforms. Fixing this without copying everything all over would be desirable.

I created an abstract class "AbstractSql92BasedPlatform" which implements PlatformInterface except for getName(). It is mostly identical to the old Sql92 class. All platform classes extend this abstract class and implement only platform-specific differences. This results in a lot less code. It's also more flexible - implementing a new platform would become much simpler.

The changes are fully backward compatible - all methods are still in place and behave identically. The only difference is inheritance from the new base class.

I did not change the tests, though many tests would be redundant now. However, the existing tests are simple and document the full platform interface. I also ran the integration tests without problems, except for SqlServerIntegrationTest because I don't have a test platform available.

### DIFF
--- a/library/Zend/Db/Adapter/Platform/AbstractSql92BasedPlatform.php
+++ b/library/Zend/Db/Adapter/Platform/AbstractSql92BasedPlatform.php
@@ -9,58 +9,23 @@
 
 namespace Zend\Db\Adapter\Platform;
 
-use Zend\Db\Adapter\Driver\DriverInterface;
-use Zend\Db\Adapter\Driver\Pdo;
-use Zend\Db\Adapter\Exception;
-
-class SqlServer extends AbstractSql92BasedPlatform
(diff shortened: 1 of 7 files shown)


weierophinney:
@hschletz This definitely looks interesting. However, I need you to resubmit this against the current develop branch, as it will hit that (due to the sweeping changes). If you can do that before 9 March 2015, we can schedule it for 2.4.

hschletz:
Replaced with #7258.

PR #7062, opened 2014-12-21
remove unused variables

UFOMelkor:


library/Zend/Json/Decoder.php
```diff
@@ -267,7 +267,6 @@ protected function _decodeObject()
 
             $tok = $this->_getNextToken();
             $members[$key] = $this->_decodeValue();
-            $tok = $this->token;
```
Ocramius: This change will be reverted

library/Zend/Json/Encoder.php
```diff
@@ -359,7 +359,6 @@ private static function _encodeVariables(ReflectionClass $cls)
         $properties = $cls->getProperties();
         $propValues = get_class_vars($cls->getName());
         $result = "variables:{";
-        $cnt = 0;
```
Ocramius: This change will be reverted

PR #7070, opened 2014-12-22
Travis optimizations

marc-mabe:
- disabled GC on running unit tests and cs fixer
- allow to run more parallel processes than CPUs
- ignore special folders ('_files' and 'AllTests') to run with phpunit
- no longer keep order of parallel processes
  - this one isn't for all over performance but for humans - If you wait for a specific test case you will be informed asap the component has been ready

This should be make little performance improvements.

A much better performance improvement would be to run all unit tests and cs fixer in parallel, too.
But I currently haven't a good idea how to do that simplified.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,10 @@ before_script:
 
 script:
   # Run tests for the various components in parallel
-  - ls -d tests/ZendTest/* | parallel --gnu --keep-order 'echo "Running {} tests"; ./vendor/bin/phpunit -c tests/phpunit.xml.dist --coverage-php build/coverage/coverage-{/.}.cov {};' || exit 1
+  - ls -d tests/ZendTest/* | grep -v 'tests/ZendTest/_files' | grep -v 'tests/ZendTest/AllTests' | parallel --gnu -P 0 'echo "Running {} tests"; php -d zend.enable_gc=0 ./vendor/bin/phpunit -c tests/phpunit.xml.dist --coverage-php build/coverage/coverage-{/.}.cov {};' || exit 1
+
   # Run coding standard checks in parallel
-  - ls -d library/Zend/* tests/ZendTest/* bin | parallel --gnu --keep-order 'echo "Running {} CS checks"; ./vendor/bin/php-cs-fixer fix {} -v --dry-run --config-file=.php_cs;' || exit 1
+  - ls -d library/Zend/* tests/ZendTest/* bin | parallel --gnu -P 0 'echo "Running {} CS checks"; php -d zend.enable_gc=0 ./vendor/bin/php-cs-fixer fix {} -v --dry-run --config-file=.php_cs;' || exit 1
 
 after_script:
   # Merges the individual clover reports of each component into a single clover.xml


.travis.yml
```diff
@@ -17,9 +17,10 @@ before_script:
 
 script:
   # Run tests for the various components in parallel
-  - ls -d tests/ZendTest/* | parallel --gnu --keep-order 'echo "Running {} tests"; ./vendor/bin/phpunit -c tests/phpunit.xml.dist --coverage-php build/coverage/coverage-{/.}.cov {};' || exit 1
+  - ls -d tests/ZendTest/* | grep -v 'tests/ZendTest/_files' | grep -v 'tests/ZendTest/AllTests' | parallel --gnu -P 0 'echo "Running {} tests"; php -d zend.enable_gc=0 ./vendor/bin/phpunit -c tests/phpunit.xml.dist --coverage-php build/coverage/coverage-{/.}.cov {};' || exit 1
+
   # Run coding standard checks in parallel
-  - ls -d library/Zend/* tests/ZendTest/* bin | parallel --gnu --keep-order 'echo "Running {} CS checks"; ./vendor/bin/php-cs-fixer fix {} -v --dry-run --config-file=.php_cs;' || exit 1
+  - ls -d library/Zend/* tests/ZendTest/* bin | parallel --gnu -P 0 'echo "Running {} CS checks"; php -d zend.enable_gc=0 ./vendor/bin/php-cs-fixer fix {} -v --dry-run --config-file=.php_cs;' || exit 1
```
Ocramius: What about HHVM?

marc-mabe: @Ocramius What do you mean? The only PHP runtme difference is `zend.enable_gc=0` and it will be ignored by HHVM.

Ocramius: If it will be ignored then it's fine :)

Ocramius:
Merged, thanks @marc-mabe!

`master`: zendframework/zf2@a98e78975c6963862e77621e3f7611d3d29ed290
`develop`: zendframework/zf2@0f4e072fa481e3481f9d4ab2320f5fe0b60881b1

Ocramius:
@marc-mabe this seems to break https://travis-ci.org/zendframework/zf2/jobs/46273266#L1171 :-(

marc-mabe:
mh interesting - this error doesn't come up on my tests.
than let us remove the gc disabling

Ocramius:
It only happens on `develop`: I assume a very slow test in Zend\Math is
involved

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

marc-mabe:
It's on exporting tested results of Zend\Math and it has 312585 assertions.

Ocramius:
Probably due to RNG tests...

marc-mabe:
test coverage should not be ignored only because it's slow

Ocramius:
@marc-mabe repeatedly stress-testing the same segment of code during a statistical benchmark isn't really useful for us (for the purpose of coverage)

marc-mabe:
the problem is that this test isn't only a stress test. It would be optimal to reduce the cycles for rnd stress test to be 1 to be a normal unit test.

Ocramius:
@marc-mabe isn't `@coversNothing` on the single test that runs the stress-operations enough? Worth trying, imo

marc-mabe:
@Ocramius Good idea but even with `@coversNothing` the rnd test runs into out of memory with `zend.enable_gc=0` and coverage enabled.

Results on running `Math/RandTest.php` on my machine:
gc off + coverage on = out of memory
gc off + coverage on +  `@coversNothing` = out of memory
gc on + coverage on = 6.14 minutes, 9.75Mb
gc on + coverage on +  `@coversNothing` = 6.06 minutes, 9.75Mb
gc off + coverage off = out of memory
gc on + coverage off = 49.73 seconds, 6.75Mb

So I think gc needs to be enabled on running phpunit and the biggest slow down is code coverage generation.

Btw. the cs fixer is ~ 1-5% faster with gc off

PR #7084, opened 2014-12-29
Feature/number parse option for number form element

fabiocarneiro:
Add a NumberParse option to number form element, plus set the default locale to "en" since the html5 number form element always return in this locale.

### DIFF
--- a/library/Zend/Form/Element/Number.php
+++ b/library/Zend/Form/Element/Number.php
@@ -31,12 +31,12 @@ class Number extends Element implements InputProviderInterface
     /**
      * @var array
      */
-    protected $validators;
+    protected $validators = array();
 
     /**
      * @var array
      */
-    protected $filters;
+    protected $filters = array();
 
     /**
      * Get validator
@@ -84,7 +84,7 @@ class Number extends Element implements InputProviderInterface
         }
 
         $this->validators = $validators;
-        
+
         return $this->validators;
     }
 
--- a/library/Zend/Form/Element/Number.php
+++ b/library/Zend/Form/Element/Number.php
@@ -86,7 +86,7 @@ class Number extends Element implements InputProviderInterface
         $this->validators = $validators;
         return $this->validators;
     }
-
+    
     protected function getFilters()
     {
         if ($this->filters) {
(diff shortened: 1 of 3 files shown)


library/Zend/Form/Element/Number.php
```diff
@@ -56,32 +64,56 @@ protected function getValidators()
...
+            return $this->filters;
+        }
+
+        $this->filters = array(
+            new StringTrim()
+        );
+
+        if (isset($this->options['format'])) {
+            $this->filters[] = new NumberParse(array(
+                'locale' => 'en',
```
jaapio: this will fail when I want to use a different locale?

weierophinney:
> We are trying to decouple components, whereas this is adding more coupling between them.

True, but `Zend\Form` is essentially an integration component already, as it has dependencies on `Zend\InputFilter` (and, by extension, `Zend\Filter` and `Zend\Validator`), and, to a degree, `Zend\View` (as it provides helpers).

That said, we've explicitly not had `Zend\InputFilter` depend on `Zend\I18n`, and I'm not understanding why the `NumberParse` filter is a requirement of the `Number` element. @fabiocarneiro : is this class unusable without the filter? or just unusable for your use case?

There are other possibilities here:

- Add a `Form` subcomponent in the `Zend\I18n` namespace, and have an i18n-specific extension there.
- Offer it in a personal library/package

I'm not saying definitively that we won't merge it, but I'd really like to know the use case, and we definitely need to consider the dependency.

fabiocarneiro:
@weierophinney :D! In my use case, it made sense to me to have automatically parsed numbers, and the zf2 feature that does that is the NumberParse filter, which belongs to ```Zend\i18n```. 

Of course we could rely in php intl directly, but IMHO that's worse than relying in ```Zend\i18n```.

After this PR, i had a talk with @Ocramius on IRC and we both agreed that it would be more inteligent to move this logic to another repo, like the ```Zend\i18n``` component, which already has ```Zend\InputFilter``` features, but i didn't have more time to work on that since i'm running with company things and then this issue is stuck here.

That said, some time ago I had another issue (which i don't remember exacly what was right now) and i came back in this same dependency from Zend\Form, and then I considered it important again.

weierophinney:
I'm removing the 2.4 milestone at this time. Solutions for the short term include, as noted above, creating your own i18n-specific implementation, and wiring it into your project.

Long term, we can touch base again after the repository split, and see if this should belong in the `Zend\I18n` repo instead.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-form/issues/96

PR #7086, opened 2014-12-30
Honor returned status code for HEAD requests

chadicus:
I came across this while attempting HEAD requests with apigility.  I found that HEAD requests will always return 200 OK even when the resource is not found.

### DIFF
--- a/library/Zend/Mvc/Controller/AbstractRestfulController.php
+++ b/library/Zend/Mvc/Controller/AbstractRestfulController.php
@@ -374,8 +374,8 @@ abstract class AbstractRestfulController extends AbstractController
                     $id = null;
                 }
                 $action = 'head';
-                $this->head($id);
-                $response = $e->getResponse();
+                $headResult = $this->head($id);
+                $response = ($headResult instanceof Response) ? clone $headResult : $e->getResponse();
                 $response->setContent('');
                 $return = $response;
                 break;
--- a/tests/ZendTest/Mvc/Controller/RestfulControllerTest.php
+++ b/tests/ZendTest/Mvc/Controller/RestfulControllerTest.php
@@ -184,6 +184,21 @@ class RestfulControllerTest extends TestCase
         $this->assertEquals('patch', $this->routeMatch->getParam('action'));
     }
 
+    public function testOnDispatchHonorsStatusCodeWithHeadMethod()
+    {
+        $this->controller->headResponse = new Response();
+        $this->controller->headResponse->setStatusCode(418);
+        $this->controller->headResponse->getHeaders()->addHeaderLine('Custom-Header', 'Header Value');
+        $this->routeMatch->setParam('id', 1);
+        $this->request->setMethod('HEAD');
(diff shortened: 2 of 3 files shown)


Ocramius:
I am cherry-picking this into `master` as well (manually)

Ocramius:
This fix seems a bit weird to me, and likely an apigility bug: why isn't apigility just serving the returned response object, and is instead using the response object that is stored in the event?

Paging @weierophinney for this one, as I don't think the controller is misbehaving here

Ocramius:
Nvm, had a brain fart, I suppose. The return value of the action `head()` is not the actual return value of method `Zend\Stdlib\DispatchableInterface#dispatch()` in this case: that's what is being fixed.

library/Zend/Mvc/Controller/AbstractRestfulController.php
```diff
@@ -374,8 +375,12 @@ public function onDispatch(MvcEvent $e)
                     $id = null;
                 }
                 $action = 'head';
-                $this->head($id);
+                $headResult = $this->head($id);
                 $response = $e->getResponse();
+                if ($headResult instanceof HttpResponse) {
+                    $response->setStatusCode($headResult->getStatusCode());
```
Ocramius: After further inspection, this solution is a bit naive, as it doesn't copy headers and other information from the `$headResult` into the `$response`.

chadicus: @Ocramius I thought about that, just didn't want to make that large of a change. I'm fine with adding the headers as well. I'm also fine with another solution all together.

Ocramius: There is a better solution (probably): 
```php
$response = clone $headResult; $response->setContent(''); return $response;
```
then setting that new `$response` in the `MvcEvent`. Not sure why we'd keep re-using the `MvcEvent#getResponse()` as the blueprint where to take our data from.

chadicus: @Ocramius I've updated the code. let me know if this solution is a least a step in the right direction.

Ocramius:
@weierophinney the solution I suggested is a bit cleaner, but is a minor BC break here. Not sure if it's acceptable.

@chadicus thanks for updating the code, I'll have to wait for feedback before merging.

Ocramius:
After looking at it for some more time, this is actually how the MVC component is supposed to handle response objects (regardless of the `MvcEvent`). I will merge it.

Ocramius:
@chadicus merged, thanks!

`master`: zendframework/zf2@1b729542c6b24af767f00211396bdd0304691c85
`develop`: zendframework/zf2@bd38414cbab21324f455f72ac0112b73b6978d52

chadicus:
@Ocramius any ETA on a release for this? other than dev-master that is.

Ocramius:
@chadicus @weierophinney told me yesterday that the `2.3.4` release is due today or early next week

Issue #3758, opened 2013-02-12
partialLoop/partial View Helper can not be nested when using setObjectKey


tylkomat:
Since always the same `partial(Loop)` instance is used, setting a new object key inside a `partial(Loop)` overwrites also the current object key (because same instance). Maybe it is a better way to provide the `objectKey` as optional parameter for the `__invoke` methods of `partial` and `partialLoop`.

larsnystrom:
Since I just encountered this problem I thought I'd chip in.

As a user it's easy to get around this problem by saving the current objectKey to a temporary variable and restoring when you're done:

	<?php
		$prevKey = $this->partialLoop()->getObjectKey();
		$this->partialLoop()->setObjectKey('newkey');
		echo $this->partialLoop('partial.phtml', $object);
		$this->partialLoop()->setObjectKey($prevKey);
	?>

This could be implemented by the framework using a stack of `objectKeys` in the `PartialLoop` helper, letting `setObjectKey()` push a key to the stack and `__invoke()` pop a key from the stack when the loop is completed.

That would of course cause a BC break for those relying on the `objectKey` staying set during multiple invocations of `PartialLoop`. 

A somewhat more complex idea is to use a nesting-level bound `objectKey`, meaning if `partialLoop()` is called from within another `partialLoop()` the inner `objectKey` would be different from the outer object key, but default to the outer `objectKey`. I have created a pull request which implements this.

Ocramius:
This is handled in #7093

PR #7093, opened 2015-01-02
Scope objectKey in nested partialLoop call

larsnystrom:
This is a fix for #3758 and introduces an `objectKeyStack` which scopes the `objectKey` to a `nestingLevel`. This allows nested calls with the `PartialLoop` View Helper where each level of nesting uses it's own or a parent's `objectKey`.

It is not clear whether this is a bug fix or a feature request which introduces a BC break, but at least no tests are failing. I'd be happy to write a test case for this fix if necessary.

This is my first pull request ever, so please be gentle when providing feedback :)

### DIFF
--- a/library/Zend/View/Helper/PartialLoop.php
+++ b/library/Zend/View/Helper/PartialLoop.php
@@ -27,6 +27,22 @@ class PartialLoop extends Partial
     protected $partialCounter = 0;
 
     /**
+     * The current nesting level
+     *
+     * @var int
+     */
+    protected $nestedLevel = 0;
+
+    /**
+     * Stack with object keys for each nested level
+     *
+     * @var array
+     */
+    protected $objectKeyStack = array(
+        0 => null,
+    );
+
+    /**
      * Renders a template fragment within a variable scope distinct from the
      * calling View object.
      *
@@ -58,8 +74,12 @@ class PartialLoop extends Partial
         $content = '';
 
         foreach ($values as $item) {
+            $this->nestObjectKey();
+
             $this->partialCounter++;
             $content .= parent::__invoke($name, $item);
+
+            $this->unnestObjectKey();
         }
 
         return $content;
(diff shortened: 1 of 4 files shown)


Ocramius:
@larsnystrom can you please provide a test-case that verifies what is explained in #3758?

library/Zend/View/Helper/PartialLoop.php
```diff
@@ -74,4 +94,51 @@ public function getPartialCounter()
     {
         return $this->partialCounter;
     }
+
+    /**
+     * Set object key in this loop and any child loop
+     *
+     * @param string $key
+     */
+    public function setObjectKey($key)
```
Ocramius: Does this need to be `public`?

larsnystrom: That method overrides `setObjectKey()` in Partial, which we inherit. Maybe I should've added that to the docblock, but I'm unsure of the syntax.

samsonasik: use <code>{@inheritDoc}</code>

larsnystrom: I don't think I can use that tag for other things than classes.

larsnystrom:
I added a test for issue #3758. My commit should make it pass.

library/Zend/View/Helper/PartialLoop.php
Ocramius: You can use that for anything that is API overriding/implementing ;-)

larsnystrom: Guess my documentation skills needs to be improved then :) Pushed a fix for it now.

tests/ZendTest/View/Helper/_files/modules/application/views/scripts/partialLoopChildObject.phtml
```diff
@@ -0,0 +1,10 @@
+<?php
+	$vars = $this->vars();
+	if (empty($vars)) :
```
samsonasik: use curly braces ( { } ) instead https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#51-if-elseif-else

samsonasik: nvm, it ignored in <code>.php_cs</code> config  ;)

larsnystrom: Oh well, I've already pushed a fix :)

library/Zend/View/Helper/PartialLoop.php
```diff
@@ -74,4 +94,52 @@ public function getPartialCounter()
...
+
+        return $this;
+    }
+
+    /**
+     * Increment nestedLevel and default objectKey to parent's value
+     *
+     * @return self
+     */
+    protected function nestObjectKey()
```
Ocramius: I will make these new methods `private` upon merge, since they really are a deep implementation detail

Ocramius:
@larsnystrom merged, thanks!

`master`: zendframework/zf2@6bf0b23bdfa81dcb272e34f6098028c0ce819f4c
`develop`: zendframework/zf2@36b371473e705255e85b36c18609340ad4206530

PR #7094, opened 2015-01-03
Fixes #6366: Session validators

### LINKED ISSUE #6366: Session validators does not work
Session validators attaching to ValidatorChain before session start. On attaching validator write metadata to session storage:
```php
    public function attach($event, $callback = null, $priority = 1)
    {
        $context = null;
        if ($callback instanceof Validator) {
            $context = $callback;
        } elseif (is_array($callback)) {
            $test = array_shift($callback);
            if ($test instanceof Validator) {
                $context = $test;
            }
            array_unshift($callback, $test);
        }
        if ($context instanceof Validator) {
            $data = $context->getData();
            $name = $context->getName();
            $this->getStorage()->setMetadata('_VALID', array($name => $data));
        }

        $listener = parent::attach($event, $callback, $priority);
        return $listener;
    }
```

After, when you call SessionManager::start(), session_start() override storage ($_SESSION) and validator's metadata become lost.

At first I thought that I improperly use validators. And i should attach validator after calling SessionManager::start(). But in documentation (http://framework.zend.com/manual/2.3/en/modules/zend.session.manager.html), code of SessionManagerFactory class (https://github.com/zendframework/zf2/blob/master/library/Zend/Session/Service/SessionManagerFactory.php#L117) and test file (https://github.com/zendframework/zf2/blob/master/tests/ZendTest/Session/SessionManagerTest.php#L510)  authors use this order.

Martin-P:
Fix for issue #6366.

The session validators need the reference value at instantiation. The current implementation does _not_ insert the reference value ([`Zend\Session\Service\SessionManagerFactory` line 120](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/library/Zend/Session/Service/SessionManagerFactory.php#L120)) and therefore the default value is used and validation always returns true.

The session validators need to be set _after_ the session has started. The instantiation of the session validators can not be done in `Zend\Session\Service\SessionManagerFactory`, because the session has not started yet.

This PR fixes that issue and initializes the validators _after_ the session has started. This way the correct values are inserted into the validators.

### DIFF
--- a/library/Zend/Session/AbstractManager.php
+++ b/library/Zend/Session/AbstractManager.php
@@ -49,15 +49,25 @@ abstract class AbstractManager implements Manager
     protected $saveHandler;
 
     /**
+     * @var array
+     */
+    protected $validators;
+
+    /**
      * Constructor
      *
-     * @param  Config|null $config
-     * @param  Storage|null $storage
+     * @param  Config|null      $config
+     * @param  Storage|null     $storage
      * @param  SaveHandler|null $saveHandler
+     * @param  array            $validators
      * @throws Exception\RuntimeException
      */
-    public function __construct(Config $config = null, Storage $storage = null, SaveHandler $saveHandler = null)
-    {
+    public function __construct(
+        Config $config = null,
+        Storage $storage = null,
+        SaveHandler $saveHandler = null,
+        array $validators = array()
+    ) {
         // init config
         if ($config === null) {
             if (!class_exists($this->defaultConfigClass)) {
@@ -106,6 +116,8 @@ abstract class AbstractManager implements Manager
         if ($saveHandler !== null) {
             $this->saveHandler = $saveHandler;
         }
+
+        $this->validators = $validators;
     }
 
     /**
--- a/library/Zend/Session/Service/SessionManagerFactory.php
+++ b/library/Zend/Session/Service/SessionManagerFactory.php
@@ -65,6 +65,7 @@ class SessionManagerFactory implements FactoryInterface
         $config        = null;
         $storage       = null;
         $saveHandler   = null;
+        $validators    = array();
         $managerConfig = $this->defaultManagerConfig;
 
         if ($services->has('Zend\Session\Config\ConfigInterface')) {
@@ -103,8 +104,6 @@ class SessionManagerFactory implements FactoryInterface
             }
         }
 
-        $manager = new SessionManager($config, $storage, $saveHandler);
-
         // Get session manager configuration, if any, and merge with default configuration
         if ($services->has('Config')) {
             $configService = $services->get('Config');
@@ -113,15 +112,13 @@ class SessionManagerFactory implements FactoryInterface
             ) {
                 $managerConfig = array_merge($managerConfig, $configService['session_manager']);
             }
-            // Attach validators to session manager, if any
+
             if (isset($managerConfig['validators'])) {
-                $chain = $manager->getValidatorChain();
-                foreach ($managerConfig['validators'] as $validator) {
-                    $validator = new $validator();
-                    $chain->attach('session.validate', array($validator, 'isValid'));
-                }
+                $validators = $managerConfig['validators'];
             }
         }
+
+        $manager = new SessionManager($config, $storage, $saveHandler, $validators);
 
         // If configuration enables the session manager as the default manager for container
         // instances, do so.
--- a/library/Zend/Session/SessionManager.php
+++ b/library/Zend/Session/SessionManager.php
@@ -40,14 +40,19 @@ class SessionManager extends AbstractManager
     /**
      * Constructor
      *
-     * @param  Config\ConfigInterface|null $config
-     * @param  Storage\StorageInterface|null $storage
+     * @param  Config\ConfigInterface|null           $config
+     * @param  Storage\StorageInterface|null         $storage
      * @param  SaveHandler\SaveHandlerInterface|null $saveHandler
+     * @param  array                                 $validators
      * @throws Exception\RuntimeException
      */
-    public function __construct(Config\ConfigInterface $config = null, Storage\StorageInterface $storage = null, SaveHandler\SaveHandlerInterface $saveHandler = null)
-    {
-        parent::__construct($config, $storage, $saveHandler);
+    public function __construct(
+        Config\ConfigInterface $config = null,
+        Storage\StorageInterface $storage = null,
+        SaveHandler\SaveHandlerInterface $saveHandler = null,
+        array $validators = array()
+    ) {
+        parent::__construct($config, $storage, $saveHandler, $validators);
         register_shutdown_function(array($this, 'writeClose'));
     }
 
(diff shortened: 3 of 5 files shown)


Martin-P:
@Ocramius I think it needs to be marked for 2.4.0, because this PR changes exisiting behaviour and validators are only present in `ValidatorChain` *after* calling `Zend\Session\SessionManager::start()`.

I had to alter the existing test `ZendTest\Session\Service\SessionManagerFactoryTest::testFactoryWillAddValidatorViaConfiguration` and add `$manager->start()` to compensate for this.

Ocramius:
I'll have to keep this on hold until we decided what to do with it /cc @weierophinney @mwillbanks

mwillbanks:
@Ocramius this looks good to me from a high level, doesn't really seem to break BC since it is coming from a abstract but if someone has extended it there is a possibility of BC break but highly unlikely.

Ocramius:
While I see how this fixes the issue, it also requires a radical change in how the consumer of the API has to use the `SessionManager`, no?

Martin-P:
People do not have to change how they use `SessionManager`. They can still use it as they did before, because the new `$Validators` parameter is fully optional. This change only fixes the validators which are added via `SessionManagerFactory`.

The only issue is validators defined in *config* and added via `SessionManagerFactory` (not the regular way by using `SessionManager`) are only present in `ValidatorChain` *after* calling `Zend\Session\SessionManager::start()`. In the current implementation validators are directly available after calling `SessionManagerFactory`, even before calling `SessionManager::start()`.

When using `SessionManager` there is no difference, but when using `SessionManagerFactory` there is a slight difference in when validators are available (before vs. after calling `SessionManager::start()`).

Ocramius:
I'm concerned by the fact that standalone `SessionManager` usage still causes broken validation when the validators are attached to the validator chain by hand (before starting the session)

Martin-P:
True, that is indeed still possible and is not covered in this PR. This PR only fixes the `SessionManagerFactory`.

I am not sure if that kind of programming logic can and should be fixed at all. I think we can expect people to know you need to start a session before you can use the values stored in a session? The documentation has been updated with this info 5 months ago (https://github.com/zendframework/zf2-documentation/pull/1342).

Ocramius:
@Martin-P please contact me per email (I don't have your contact).

weierophinney:
Closing in favor of ddbf43a

Martin-P:
@weierophinney @Ocramius
Sorry to inform you both this does not fix the issue. Just add the unittests from my PR and you will see they fail:
- `SessionManagerFactoryTest::testFactoryDoesNotOverwriteValidatorStorageValues`
- `SessionManagerFactoryTest::testSessionValuesShouldNotExistBeforeCallingStart`

weierophinney:
@Martin-P From what I can tell, the reason the patch from @Ocramius fails your own unit tests is because the approach taken is different. Essentially, the approach used in the patch applied still writes to `$_SESSION`, but assigns the values of `$_SESSION` to a temporary variable before calling `session_start()`. Once that call is made, it tests to see if that temporary variable is non-empty, and, if so, merges it with the session storage. As such, I'd expect it to fail your tests, as they are testing that `$_SESSION` is empty before `session_start()` is called.

The real question is whether it solves the reported problem. From the evaluation I and others have done, I think it does. If you can prove that it doesn't, we'll re-evaluate.

Ocramius:
I just tried to patch these tests into current master and step-debugged into them.

The security flaw is fixed, but there are indeed some different approaches in the fix itself that make the tests in this specific PR not working.

Running the session validation with the existing `SessionManagerFactory` does correctly write validator data into the session being saved though.

I carefully step-debugged into everything at runtime to make sure things are correct.

Thanks for raising awareness, it's good to double check this stuff :-)

Martin-P:
> The real question is whether it solves the reported problem. From the evaluation I and others have done, I think it does.

The reported issue was never about the `SessionManager` but about the `SessionManagerFactory`. See this comment: https://github.com/zendframework/zf2/issues/6366#issuecomment-65645461

>  If you can prove that it doesn't, we'll re-evaluate.

```
There were 2 failures:

1) ZendTest\Session\Service\SessionManagerFactoryTest::testFactoryDoesNotOverwriteValidatorStorageValues
Failed asserting that null is identical to 'Foo'.

...\vendor\zf2\tests\ZendTest\Session\Service\SessionManagerFactoryTest.php:117

2) ZendTest\Session\Service\SessionManagerFactoryTest::testSessionValuesShouldNotExistBeforeCallingStart
ZF session values should not yet exist
Failed asserting that true is false.
```

So, indeed one test fails because of testing if $_SESSION is empty, but the other test (`SessionManagerFactoryTest::testFactoryDoesNotOverwriteValidatorStorageValues`) clearly indicates the issue is still present.

Ocramius:
@Martin-P ah yes, that bug is still present, but validator data is currently written/read during session start (and replaced by loaded data when the actual `session_start()` call happens), which ensures that validation will go against the previous HTTP request.

Ocramius:
Actually, that's not even a bug anymore, there is a minor behavior change though, but it's an acceptable change IMO

Martin-P:
> which ensures that validation will go against the previous HTTP request

Only when using `SessionManager` directly. When using `SessionManagerFactory` (what issue #6366 is about) the validators still don't work as they should.

Martin-P:
> Actually, that's not even a bug anymore, there is a minor behavior change though, but it's an acceptable change IMO

The test indicating the problem fails, the session data is not used, the validators do not work as they should, but it is not a bug anymore? Can you please explain why the test indicating the problem fails?

Ocramius:
@Martin-P if you look at `SessionManager#start()`, that's where the mutation of `$_SESSION` happens, but it is not possible to introspect it until after the session has actually been started (and may contain older validation data)

Martin-P:
> but it is not possible to introspect it until after the session has actually been started

If that happens adding `SessionManager::start()` to the testcase should make it pass and  the session data should be correctly present after calling `SessionManager::start()`. I added it to the testcase, but it still fails.
```php
/**
 * @group 6366
 * @runInSeparateProcess
 */
public function testFactoryDoesNotOverwriteValidatorStorageValues()
{
    $storage = new ArrayStorage();
    $storage->setMetadata('_VALID', array(
        'Zend\Session\Validator\HttpUserAgent' => 'Foo',
        'Zend\Session\Validator\RemoteAddr'    => '1.2.3.4',
    ));
    $this->services->setService('Zend\Session\Storage\StorageInterface', $storage);

    $config = array(
        'session_manager' => array(
            'validators' => array(
                'Zend\Session\Validator\HttpUserAgent',
                'Zend\Session\Validator\RemoteAddr',
            ),
        ),
    );
    $this->services->setService('Config', $config);

    // This call is needed to make sure session storage data is not overwritten by the factory
    $manager = $this->services->get('Zend\Session\ManagerInterface');
    $manager->start();

    $validatorData = $storage->getMetaData('_VALID');
    $this->assertSame('Foo'    , $validatorData['Zend\Session\Validator\HttpUserAgent']);
    $this->assertSame('1.2.3.4', $validatorData['Zend\Session\Validator\RemoteAddr']);
}
```
Testing against the latest master however results in an error:
```
There was 1 error:

1) ZendTest\Session\Service\SessionManagerFactoryTest::testFactoryDoesNotOverwriteValidatorStorageValues
Undefined variable: _SESSION
```
For the sake of argument I simply suppress the error by using `$oldSessionData = @$_SESSION;`, but the test still does not pass:
```
There was 1 error:

1) ZendTest\Session\Service\SessionManagerFactoryTest::testFactoryDoesNotOverwriteValidatorStorageValues
Zend\Session\Exception\RuntimeException: Session validation failed
```
What am I doing wrong and what should be changed to make the test pass?

Martin-P:
After taking another look the fix from @Ocramius fixes the security issue, however it also introduces some new issues:
- `Undefined variable: _SESSION` &rarr; fixed in #7127
- Session validators via config always fail using `SessionManagerFactory`. I will open a new PR for this, probably with the same suggested fix (working on it).

padraic:
Also, may I note that should there be future security concerns, do please report these as per the security policy: http://framework.zend.com/security This will ensure you get more eyeballs on the issue, and it also allows us to enact responsible disclosure.

Martin-P:
@padraic It was not clear it was a security issue untill @Ocramius asked me to contact him (https://github.com/zendframework/zf2/pull/7094#issuecomment-69208094), but thanks for the additional info.

Issue #7103, opened 2015-01-05
RFC Bytes filter


veewee:
Currently, there is no filter to convert bytes into a human readable format. 
E.g 1000 bytes = 1kB

In applications which manage files / transfer speeds / ... it can be handy to provide a standard filter for this.
My suggestion:
```php
<?php
/**
 * Zend Framework (http://framework.zend.com/)
 *
 * @link      http://github.com/zendframework/zf2 for the canonical source repository
 * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
 * @license   http://framework.zend.com/license/new-bsd New BSD License
 */

namespace Zend\Filter;

use Zend\Filter\Exception\InvalidArgumentException;

class Bytes extends AbstractFilter
{
    const MODE_DECIMAL = 'decimal';
    const MODE_BINARY = 'binary';

    const TYPE_BYTES = 'B';
    const TYPE_BITS = 'b';

    /**
     * @var array
     */
    protected static $modes = [
        self::MODE_DECIMAL,
        self::MODE_BINARY,
    ];

    /**
     * @var array
     */
    protected static $types = [
        self::TYPE_BITS,
        self::TYPE_BYTES,
    ];

    /**
     * A list of standarized binary prefix formats
     * @link https://en.wikipedia.org/wiki/Binary_prefix
     *
     * @var array
     */
    protected static $standarizedPrefixes = [
        // decimal SI units:
        self::MODE_DECIMAL => ['', 'k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'],
        // binary IEC units:
        self::MODE_BINARY => ['', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi'],
    ];

    /**
     * @var array
     */
    protected $options = array(
        'mode'         => self::MODE_BINARY,
        'type'         => self::TYPE_BYTES,
        'precision'    => 2,
        'prefixes'     => [],
    );

    /**
     * @param array $options
     */
    public function __construct($options = [])
    {
        if (!static::isOptions($options)) {
            return;
        }

        $this->setOptions($options);
    }

    /**
     * @param string $mode One of binary or decimal
     *
     * @throws InvalidArgumentException
     */
    public function setMode($mode)
    {
        $mode = strtolower($mode);
        if (!in_array($mode, self::$modes)) {
[...]

PR #7104, opened 2015-01-05
DataUnitFormatter filter

veewee:
Because most of the filter was already written in RFC #7103, I decided to add some tests and directly place it in a PR.
**Usage:**

``` php
$filter = new DataUnitFormatter($options);
$result = $filter->filter(1000);
```

**Result:**
1.00kB

**Options:**
- mode: binary / decimal
- unit: B, b, ... (mandatory)
- precision: int
- prefixes: array of custom prefixes

Please feel free to give me some feedback / improvements.


library/Zend/Filter/Bytes.php
```diff
@@ -0,0 +1,245 @@
...
+        }
+        $this->options['mode'] = $mode;
+    }
+
+    /**
+     * Get current filter mode
+     *
+     * @return string
+     */
+    public function getMode()
```
Ocramius: Defining public API that is not covered by an interface is not really a good idea when mixed with `final`, as any coupling with this method will be un-mockable. Consider removing public getters alltogether

library/Zend/Filter/Bytes.php
```diff
@@ -0,0 +1,245 @@
...
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Filter;
+
+use Zend\Filter\Exception\InvalidArgumentException;
+
+final class Bytes extends AbstractFilter
```
Ocramius: The filter name is a bit ambiguous: I was thinking of a binary string filter when I read the topic title, but now I understand that it is about measurement units. Maybe it's a generic measurement unit filter?

library/Zend/Filter/Bytes.php
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        // Parse to float and check if value is not zero
+        $amount = (float) $value;
+        if ($amount == 0) {
+            return $amount . $this->getType();
+        }
+
+        // Calculate the correct size and prefix:
+        $orderOfMagnitude = $this->isBinaryMode() ? 1024 : 1000;
```
Ocramius: These two on the right should probably not be magic constants

Ocramius:
I like the overall idea: makes it much simpler to handle file-size problems. 

Question: why don't you let the filter convert to integer (or string, depending on integer size) and have a view helper append the suffixes instead?

veewee:
@Ocramius 
I applied all proposed changes except the change of name.
What name do you suggest for the filter? I was thinking something like `Units` or `Measurement`. Which one do you think suits best?
The unit is also more configurable now. This way, it is also possible to use other units then bits and bytes.

About your question:
I don't only need this functionality in views. Also in mail services like Mandrill, which have their own build-in templating system. Also: the view wouldn't know which prefix to use if you let the filter handle the calculations and the view helper the size prefixes.

moderndeveloperllc:
FWIW, I would go with `ByteUnitFormatter` to be inline with the `DateTimeFormatter` class that is existent. That's a little more explicit with what it deals with and what it actually does.

veewee:
@moderndeveloperllc 
Maybe UnitFormatter, since it's also possible to measure other units then bytes.
Sounds good!

moderndeveloperllc:
@veewee Instead of `ByteUnitFormatter` perhaps `DataUnitFormatter` or similar? The file as written is just a "byte unit" formatter. I think that other types of unit formatting would be in other files - `TimeUnitFormatter` (for seconds to hours/days/years/etc), `WeightUnitFormatter`.  I'm afraid that `UnitFormatter` suffers the same ambiguity as `Byte`.

@Ocramius I kinda have to agree with @veewee on keeping this in the `Filter` namespace. ZF2 appears to use this namespace as the a place to put transformative functions (compress, decompress, encrypt, decrypt, DateTime formatter, etc) in addition to functions that remove unwanted data (int, digits, basename, dir). This is a transformative function for sure. Maybe ZF3 can break these apart?

Of course that also means that sometime in ZF4 we should have something like this in the `I18n\Filters` namespace - CLDR has added lots of measument units in the [last few](http://cldr.unicode.org/index/downloads/cldr-24) [releases](http://cldr.unicode.org/index/downloads/cldr-26). ICU will eventually get it, so PHP will too.

Ocramius:
What I was actually thinking is a conversion like `10Mb -> 10485760` (actually useful for filtering user input into a numeric value that can be manipulated) and then a filter and/or a view helper that does the opposite.

veewee:
I will change the name to `DataUnitFormatter` for now.
Can't imagine that I have ever manually entered something like 10MB or 10km manually in a form or in an application.  This is mostly a value in a dropdown or slider which has it's own logic.
E.g: when uploading files, the size in bytes is available.
I guess this filter is enough. If you want I can also add a view helper that implements this filter.

Will work on this feature this evening.

veewee:
The name is changed and tests should pass.
Travis failed, but that looks like another issue in phpunit:
```
Fatal error: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 268435456 bytes) in /home/travis/build/zendframework/zf2/vendor/sebastian/exporter/src/Exporter.php on line 136
```

PR #7109, opened 2015-01-08
Improved disabled inputs testing when binding values into fieldset object

sjuvonen:
It's not enough to test that 'disabled' attribute exists, should also check that its value is something else than null or false or empty.


Ocramius:
Is this just an improvement of the test or is any action required?

Ocramius:
Actually, this requires a test case as well.

weierophinney:
I've written and committed a test case to validate the change.

Issue #4960, opened 2013-08-14
DateTime form element weird DateInterval step!?


TonyBogdanov:
Can someone explain what is the reason behind setting the DateInterval step to a minute instead of a second or even a millisecond in https://github.com/zendframework/zf2/blob/master/library/Zend/Form/Element/DateTime.php#L172 ?

Better, can someone explain why the DateTime element should have a DateStep validator at all?

My problem here, is that with minimal setup, one can never submit a datetime value with seconds != 0, because the step is set to a minute. It just goes into an infinite validation loop.

kamioftea:
The problem with the validation is not infinite, but the validator starts at the base date (which is 1/1/1970 if no min attribute is set) and increments by the step until it reaches or passes the submitted datetime which takes enough time that the script times out. I've created a pull request to help improve this here: [https://github.com/zendframework/zf2/pull/7110]

To get round the step validation entirely you can set a step attribute of 'any' to prevent the DateStep validator being added.

The default step is 60 seconds for a datetime element [WHATWG Spec](https://html.spec.whatwg.org/multipage/forms.html#date-and-time-state-%28type=datetime%29) (There is no datetime element in the W3C spec anymore but time also matches this [W3C Time Spec](http://www.w3.org/TR/html5/forms.html#time-state-%28type=time%29). This suggests that the default should really by a DateStep with a `DateInterval('PT60S')`.

There is an unfortunate side affect of changing the default date interval in that it will no longer be caught by the quicker check that only applies when there is a single interval unit with value 1, so will start to use the longer validation.

Ocramius:
Handled in #7110

PR #7110, opened 2015-01-08
Fix for validation timeouts, issue #4960

kamioftea:
The fallback for iterating through all possible steps from the base step until we equal or exceed the target value can take a long time (~30 seconds for me) when using a current timestamp and the default base value, which will occur when validating a Zend\Form\Element\DateTime with step != 1.

This change calculates a lower bound for the number of steps we will need to iterate over and skips to that timestamp before iterating as before. For me this brought the tests back down to under a second.

I have also added in a return false for the special case test where we have a step of 1 hour, 1 minute as if there are non-zero seconds (or non-zero minutes for step = 1 hour) then we know it isn't a valid step. I have also fixed a bug converting hours => seconds in the same day special case.

### DIFF
--- a/library/Zend/Validator/DateStep.php
+++ b/library/Zend/Validator/DateStep.php
@@ -24,12 +24,12 @@ class DateStep extends Date
     /**
      * @var array
      */
-    protected $messageTemplates = [
+    protected $messageTemplates = array(
         self::INVALID      => "Invalid type given. String, integer, array or DateTime expected",
         self::INVALID_DATE => "The input does not appear to be a valid date",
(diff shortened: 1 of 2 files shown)


weierophinney:
Just noting that this is for #4960 (so that that issue can track the PR, and this can link to it).

Ocramius:
Currently refactoring this for merge.

Ocramius:
Refactored some bits and merged. Thanks @kamioftea!

`master`: zendframework/zf2@bf9649dbff7422d589ce083a9bd6b850301cc932
`develop`: zendframework/zf2@3f1218b72576dd01ba8c3294e69d82fd246417c7

PR #7114, opened 2015-01-09
added missing return $this to setValue method.

matwright:
added return $this to the setValue method to implement fluid interface like other setters in this class.

### DIFF
--- a/library/Zend/Form/Element/DateSelect.php
+++ b/library/Zend/Form/Element/DateSelect.php
@@ -89,7 +89,7 @@ class DateSelect extends MonthSelect
     /**
      * @param  string|array|\ArrayAccess|PhpDateTime $value
      * @throws \Zend\Form\Exception\InvalidArgumentException
-     * @return DateSelect
+     * @return self Provides a fluent interface
      */
     public function setValue($value)
     {
--- a/library/Zend/Form/Element/DateSelect.php
+++ b/library/Zend/Form/Element/DateSelect.php
@@ -112,6 +112,7 @@ class DateSelect extends MonthSelect
         $this->yearElement->setValue($value['year']);
         $this->monthElement->setValue($value['month']);
         $this->dayElement->setValue($value['day']);
+        return $this;
     }
 
     /**
--- a/library/Zend/Form/Element/DateSelect.php
+++ b/library/Zend/Form/Element/DateSelect.php
@@ -89,7 +89,7 @@ class DateSelect extends MonthSelect
     /**
      * @param  string|array|\ArrayAccess|PhpDateTime $value
      * @throws \Zend\Form\Exception\InvalidArgumentException
-     * @return void|\Zend\Form\Element
+     * @return DateSelect
      */
     public function setValue($value)
     {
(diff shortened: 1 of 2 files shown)


matwright:
ok, will do that shortly.

matwright:
DocBlock has been updated. For the regression test, as this method had no previous return value I'm not sure what to test exactly. This update cannot impact existing functionality and all current test cases are passing.

Ocramius:
@matwright the test should simply assert that the return value is the same of the object the method is called upon

library/Zend/Form/Element/DateSelect.php
```diff
@@ -89,7 +89,7 @@ public function getDayAttributes()
     /**
      * @param  string|array|\ArrayAccess|PhpDateTime $value
      * @throws \Zend\Form\Exception\InvalidArgumentException
-     * @return void|\Zend\Form\Element
+     * @return DateSelect
```
samsonasik: @return self imo for consistency.

samsonasik:
you can use : 
```php
$this->assertInstanceOf('Zend\Form\Element\DateSelect', $element->setValue('now'));
```

Ocramius:
@samsonasik should be `assertSame`

tests/ZendTest/Form/Element/DateSelectTest.php
```diff
@@ -83,4 +83,10 @@ public function testConstructAcceptsDayAttributes()
         $dayAttributes = $sut->getDayAttributes();
         $this->assertEquals('test', $dayAttributes['class']);
     }
+
+    public function testValueSetterReturnsSameObjectType()
+    {
+        $element  = new DateSelectElement();
+        $this->assertInstanceOf(get_class($element), $element->setValue('hello world'));
```
samsonasik: <code>setValue('hello world')</code> must be end up with Exception because of this https://github.com/matwright/zf2/blob/master/library/Zend/Form/Element/DateSelect.php#L97-L101

library/Zend/Form/Element/DateSelect.php
matwright: The other setters in this class use  @return DateSelect
would you like me to update this and the other setter docblocks to show return self ?

tests/ZendTest/Form/Element/DateSelectTest.php
matwright: thanks @samsonasik  just updated that to a valid date string

library/Zend/Form/Element/DateSelect.php
matwright: @samsonasik  The other setters in this class use @return DateSelect
would you like me to update this and the other setter docblocks to show return self ?

samsonasik: change only for this method i think because it is new introduced docblock. there is inconsistencies about it in other files too. i created PR in the past but can't finish to update all.

samsonasik:
:+1:

Ocramius:
Merged, thanks @matwright and @samsonasik!

`master`: zendframework/zf2@d76734a1cf4bbdaacf98f28148fe3fbad7a95b13
`develop`: zendframework/zf2@4567312b5cbe90aa7a421ea3460bd3b8d767db50

PR #7122, opened 2015-01-15
Paginator Iterator

tklever:
An OuterIterator to flatten Paginator objects.

Allows a developer to iterate over an entire paginated set as one loop. Not much of a use case with the ArrayAdapter, however, could be helpful with Paginator Adapters that fetch remote data.

### DIFF
--- a/library/Zend/Paginator/PaginatorIterator.php
+++ b/library/Zend/Paginator/PaginatorIterator.php
@@ -1,0 +1,113 @@
+<?php
+
+namespace Zend\Paginator;
+
+use OuterIterator;
+
+class PaginatorIterator implements OuterIterator
+{
+    /**
+     * Internal Paginator for iteration
+     *
+     * @var Paginator $paginator
+     */
+    protected $paginator;
+
+    /**
+     * Value for valid method
+     *
+     * @var bool $valid
+     */
+    protected $valid = true;
+
+    public function __construct(Paginator $paginator)
+    {
+        $this->paginator = $paginator;
+    }
+
+    /**
+     * Return the current element
+     * @link http://php.net/manual/en/iterator.current.php
+     * @return mixed Can return any type.
+     */
+    public function current()
+    {
+        return $this->getInnerIterator()->current();
+    }
+
+    /**
+     * Move forward to next element
+     * @link http://php.net/manual/en/iterator.next.php
+     * @return void Any returned value is ignored.
+     */
+    public function next()
+    {
+        $innerIterator = $this->getInnerIterator();
+        $innerIterator->next();
+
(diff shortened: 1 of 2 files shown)


library/Zend/Paginator/PaginatorIterator.php
```diff
@@ -0,0 +1,113 @@
+<?php
+
+namespace Zend\Paginator;
+
+use OuterIterator;
+
+class PaginatorIterator implements OuterIterator
```
keradus: it would be cool to document class as well

froschdesign: @keradus 
Please report this at the [issue tracker for documentation](https://github.com/zendframework/zf2-documentation).

keradus: To bo honest I do not see why.
I noticed that all props and methods have their own docs, even if it's just a copy-paste from php.net.
If docs for class should be done in separate repo then, imo, it should be done by author of PR.

library/Zend/Paginator/PaginatorIterator.php
```diff
@@ -0,0 +1,113 @@
+<?php
+
```
samsonasik: add license header block here : 
```
/**
 *
 * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
 * @license   http://framework.zend.com/license/new-bsd New BSD License
 */
```

tests/ZendTest/Paginator/PaginatorIteratorTest.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
```
samsonasik: here too : 
```
/**
 *
 * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
 * @license   http://framework.zend.com/license/new-bsd New BSD License
 */
```

tklever:
Updated to add license as requested

Ocramius:
@tklever what is exactly the use-case? I probably missed that...

tklever:
@Ocramius Use case (as I've used it) is flat mapping lazy-loaded data, specifically remote data. It also insulates code that needs to use / process that data from specific knowledge of Zend\Paginator.

It's also been helpful when passing paginated data to OuterIterators. If you want to want to run some kind of "MappingIterator" operation over a paginated data set, you can wrap it in this iterator first.

I realize you can set the Paginator to "-1" and get the entire data set, which makes this class somewhat redundant, but again, with remote data, often that's not ideal to grab those (potentially gigantic) data sets.

PR #7123, opened 2015-01-15
Fix issue with none well formated  "Accept" headers.

SamHecquet:
Add security to prevent none well formated param into "Accept" headers.

### DIFF
--- a/library/Zend/Http/Header/AbstractAccept.php
+++ b/library/Zend/Http/Header/AbstractAccept.php
@@ -167,6 +167,11 @@ abstract class AbstractAccept implements HeaderInterface
             }
 
             foreach ($paramsStrings as $param) {
+                // Prevents none well formated param
+                if(strstr($param, '=') === false) {
+                    continue;
+                }
+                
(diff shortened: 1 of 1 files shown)


library/Zend/Http/Header/AbstractAccept.php
```diff
@@ -167,6 +167,11 @@ protected function getParametersFromFieldValuePart($fieldValuePart)
             }
 
             foreach ($paramsStrings as $param) {
+                // Prevents none well formated param
+                if (strstr($param, '=') === false) {
```
Martin-P: http://php.net/strstr
> Note:
If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function strpos() instead.

Ocramius:
Requires test case

weierophinney:
Ping @SamHecquet — we need a test case to accept this.

Maks3w:
Duplicate https://github.com/zendframework/zend-http/pull/25

PR #7128, opened 2015-01-16
SessionManagerFactory + session validators via config makes validation always fail

Martin-P:
When using `Zend\Session\Service\SessionManagerFactory` validators can be added via config:

``` php
'session_manager' => array(
    'validators' => array(
        'Zend\Session\Validator\HttpUserAgent',
        'Zend\Session\Validator\RemoteAddr',
    ),
),
```

_Before_ security fix ddbf43a validators always returned true. _After_ the security fix validators always fail. This PR fixes that issue with `Zend\Session\Service\SessionManagerFactory` (not to be confused with standalone usage of `Zend\Session\SessionManager`).

It is the same fix as in #7094, but it got ignored because of the security issue.

### DIFF
--- a/library/Zend/Session/Service/SessionManagerFactory.php
+++ b/library/Zend/Session/Service/SessionManagerFactory.php
@@ -65,6 +65,7 @@ class SessionManagerFactory implements FactoryInterface
         $config        = null;
         $storage       = null;
         $saveHandler   = null;
+        $validators    = array();
         $managerConfig = $this->defaultManagerConfig;
 
         if ($services->has('Zend\Session\Config\ConfigInterface')) {
@@ -103,8 +104,6 @@ class SessionManagerFactory implements FactoryInterface
             }
         }
 
-        $manager = new SessionManager($config, $storage, $saveHandler);
-
         // Get session manager configuration, if any, and merge with default configuration
         if ($services->has('Config')) {
             $configService = $services->get('Config');
@@ -113,15 +112,13 @@ class SessionManagerFactory implements FactoryInterface
             ) {
                 $managerConfig = array_merge($managerConfig, $configService['session_manager']);
             }
-            // Attach validators to session manager, if any
+
             if (isset($managerConfig['validators'])) {
-                $chain = $manager->getValidatorChain();
-                foreach ($managerConfig['validators'] as $validator) {
-                    $validator = new $validator();
-                    $chain->attach('session.validate', array($validator, 'isValid'));
-                }
+                $validators = $managerConfig['validators'];
             }
         }
+
+        $manager = new SessionManager($config, $storage, $saveHandler, $validators);
 
         // If configuration enables the session manager as the default manager for container
         // instances, do so.
--- a/tests/ZendTest/Session/Service/SessionManagerFactoryTest.php
+++ b/tests/ZendTest/Session/Service/SessionManagerFactoryTest.php
@@ -75,6 +75,9 @@ class SessionManagerFactoryTest extends \PHPUnit_Framework_TestCase
         $this->assertNotSame($manager, Container::getDefaultManager());
     }
 
+    /**
+     * @runInSeparateProcess
+     */
     public function testFactoryWillAddValidatorViaConfiguration()
     {
         $config = array('session_manager' => array(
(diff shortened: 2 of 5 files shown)


library/Zend/Session/Service/SessionManagerFactory.php
```diff
@@ -113,16 +112,14 @@ public function createService(ServiceLocatorInterface $services)
             ) {
                 $managerConfig = array_merge($managerConfig, $configService['session_manager']);
             }
-            // Attach validators to session manager, if any
```
jaapio: why was this comment removed?

Martin-P: The old code attached validators at that point (`$chain->attach(...)`). That code has been removed and there are no validators attached at that point anymore, which makes the comment obsolete.

YDyachenko:
Hi guys.
This fix have one problem. He attach one validator two times (in `SessionManager::initializeValidatorChain()` and `ValidatorChain::__construct()`) 

Also i think what we should setup validators via `ValidatorChain` rather than `SessionManager` constructor.

So i created another patch for this issue: https://github.com/YDyachenko/zf2/commit/feecb0aafaceab130524c0bb70e6ad454208710c

Martin-P:
> He attach one validator two times (in `SessionManager::initializeValidatorChain()` and `ValidatorChain::__construct()`)

Can you specify where that happens? In my code nothing is attached at `ValidatorChain::__construct()`.
See [line 353 of `Zend\Session\SessionManager`](https://github.com/Martin-P/zf2/blob/95cd0c37fac4c12a3a3350ce927da373b7a80ff4/library/Zend/Session/SessionManager.php#L353):
```php
$this->setValidatorChain(new ValidatorChain($this->getStorage()));
```
Perhaps you can post a failing unittest for that? That way it is easier to check if that actually happens.

YDyachenko:
> Can you specify where that happens? In my code nothing is attached at `ValidatorChain::__construct()`.

If session contains validators, they will be attached:
[Line 36 of Zend\Session\ValidatorChain](https://github.com/Martin-P/zf2/blob/95cd0c37fac4c12a3a3350ce927da373b7a80ff4/library/Zend/Session/ValidatorChain.php#L36).

Test:
```php
    /**
     * @runInSeparateProcess
     */
    public function testFactoryDoesNotAttachValidatorTwoTimes()
    {
        $storage = new ArrayStorage();
        $storage->setMetadata('_VALID', array(
            'Zend\Session\Validator\RemoteAddr' => '127.0.0.1',
        ));
        $this->services->setService('Zend\Session\Storage\StorageInterface', $storage);

        $config = array(
            'session_manager' => array(
                'validators' => array(
                    'Zend\Session\Validator\RemoteAddr',
                ),
            ),
        );
        $this->services->setService('Config', $config);

        $manager = $this->services->get('Zend\Session\ManagerInterface');
        try {
            $manager->start();
        } catch (RuntimeException $e) {
            // pass
        }
        
        $this->assertEquals(1, $manager->getValidatorChain()->getListeners('session.validate')->count()); 
    }
```

>1) ZendTest\Session\Service\SessionManagerFactoryTest::testFactoryDoesNotAttachValidatorTwoTimes
>Failed asserting that 2 matches expected 1.

Martin-P:
Did a rebase because I encountered some merge conflicts.

@YDyachenko Thanks! I added your test to this PR and fixed the issue for that specific test.

tests/ZendTest/Session/Service/SessionManagerFactoryTest.php
```diff
@@ -85,6 +88,8 @@ public function testFactoryWillAddValidatorViaConfiguration()
         $this->services->setService('Config', $config);
         $manager = $this->services->get('Zend\Session\ManagerInterface');
 
+        $manager->start();
```
weierophinney: This changes the context for the test! The original test omitted it for a specific reason, based on the security issue we addressed.

Martin-P: Please see the previous discussion: https://github.com/zendframework/zf2/pull/7094#issuecomment-68605430

That is why I said it is a BC break, because the validators are now attached *after* the session has started. This makes sense, because *before* the session has not started there is no session data to compare with and thus the validators always fail.

In my opening post https://github.com/zendframework/zf2/pull/7128#issue-54595275 I explained the tests always passed *before* the security fix and always fail *after* the security fix. The tests added to detect the security fix still pass, which shows this PR does not compromise the security fix in any way.

This test is not related to the security issue, because:
1. If it was here to detect the security issue, it should have failed before the security issue was even addressed. It did not and it always passed the test.
2. The commit which fixed the security issue changed nothing about this test: https://github.com/zendframework/zf2/commit/ddbf43ac3fe28fe98a4104993d0cb4bffb13a026

weierophinney:

PR #7129, opened 2015-01-16
Tests for SessionManager preserveStorage flag

Martin-P:
`Zend/Session/SessionManager::start()` has an optional flag `$preserveStorage`: [`Zend/Session/SessionManager` line 84](https://github.com/zendframework/zf2/blob/037c62b140fa3dfd2b150563cd354bbb3230cd92/library/Zend/Session/SessionManager.php#L84)

> If set to true, current session storage will not be overwritten by the contents of $_SESSION.

This PR tests that functionality. Not all tests pass, because the `$preserveStorage` flag is not working as it should.

### DIFF
--- a/tests/ZendTest/Session/SessionManagerTest.php
+++ b/tests/ZendTest/Session/SessionManagerTest.php
@@ -11,6 +11,7 @@ namespace ZendTest\Session;
 
 use Zend\Session\SessionManager;
 use Zend\Session;
+use Zend\Session\Storage\ArrayStorage;
 use Zend\Session\Validator\RemoteAddr;
 
 /**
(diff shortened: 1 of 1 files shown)


Ocramius:
@Martin-P that functionality actually never worked in 2.x as far as I can see...

Martin-P:
That's possible, I only added the tests for what it should do. Not sure if it needs fixing if nobody uses this feature.

weierophinney:
@Martin-P want to take a crack at making it work? If so, I need changes by tomorrow, 19 March 2015.

Martin-P:
@weierophinney I do not think this can ever work. The security fix for session validators made me aware of this feature (https://github.com/zendframework/zf2/commit/ddbf43ac3fe28fe98a4104993d0cb4bffb13a026). The preservation of $_SESSION data is part of the security fix. Making the preservation of $_SESSION data optional would compromise the security fix.

To make this work the security fix needs a change. The tests covering the security fix are present, so theoretically it should not be a problem. I don't know if you consider it an option to change the security fix?

weierophinney:
@Martin-P We can change the security fix so long as the tests introduced by it continue to pass.

Martin-P:
I can take a look at this, but I think it will be after the 2.4.0 release however.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-session/issues/50

PR #7108, opened 2015-01-08
Fix form annotation options

carnage:
The options annotation and composed object annotation overwrite each other under certain conditions. The PR fixes this behaviour so that the specified options are merged.

PR #7131, opened 2015-01-17
Hotfix/additional form options error case

carnage:
Fixes an additional case of the previous bug where composed objects overwrite options specified in other annotations


jaapio:
please add a new testcase for this fix, rather then changing an existing one.

tests/ZendTest/Form/Annotation/AnnotationBuilderTest.php
```diff
@@ -198,7 +199,7 @@ public function testOptionsAnnotationAndComposedObjectAnnotation($childName)
 
         $child = $form->get($childName);
 
-        $target = $child->getTargetElement();
+        $target = ($child instanceof Collection)? $child->getTargetElement() : $child;
```
keradus: I would add a space before that `?`

carnage:
@jappio This fix is for another occurance of the bug that the test is for, it extends the test case via a data provider, so in this instance I don't think a separate test case is warrented.

jaapio:
Changing a test is always a bad thing. The test might have the same intension as the issue your are testing. But somehow you found a bug in the behavior, so you need te be sure you don't break anything else. Leaving existing tests as they are will help you checking this.

Ocramius:
Agree with @jaapio here: can a new test be written instead?

carnage:
ok I'll c+p the test into a new method.

Linking #7108 for reference

carnage:
Build failures seem unrelated. Is this good to merge now @weierophinney

carnage:
reopened against develop #7309

PR #7133, opened 2015-01-17
[WIP - ZF3] Refactored SM

bakura10:
Hi everyone :),

Until we have proper split of components, let me submit a refactored service manager. There are a few features that I have not completley backported (proxy and delegators). As all my other refactors, this emphasize on performance, and I've tested it carefully with BlackFire.

The main changes is a change of interfaces: now everything is a factory, and each factory is callable, so this removes all the needed check against FactoryInterface or callable.

Furthermore, there is no longer concept of invokables: invokables are implemented as a built-in factory:

``` php
return [
   'service_manager' => [
       'factories' => [
             MyInvokable::class => InvokableFactory::class
       ]
  ]
];
```

In one of my previous idea, I've introduced the concept of "abstract_factories_map". In this refacotr, I've been able to remove this additional concept while keeping the idea of being able to map multiple class names to the same factory. This is make possible because the interface of a factory now has the requested class name as second parameter.

Finally, this introduce a simplification to plugin manager, and remove the need of calling "getServiceLocator". While using ZF2 extensively, I've realized that, in a factory of a plugin maanger, I nearly never need to fetch other deendencies from this plugin manager, but rather calling dependencies from the main service locator. This refactor therefore changes the context of the factory in a plugin manager.

In ZF2, if you fetched a service from a plugin manager, you had to call "getServiceLocator" to retrieve the main service locator. In this refactor, the service locator is actually the main service locator.

Finally (bis), this removes aliases from the main service manager for performance reasons. Now that we have ::class, the only reason to use alaises is for very specific plugin managers, like view plugin managers or controller plugin managers. Therefore, I've introduced an addditional, more specialized plugin manager called "AliasesPluginMAnager", that brings back the aliases.

This make the component MUCH MUCH more efficient (tbh I think there is no other way to make it faster while preserving all features), and more coherent, especially when it comes to plugin maangers.


Ocramius:
> tbh I think there is no other way to make it faster while preserving all features

Actually, there is, and even by maintaining the previous feature-set, IMO.
Yes, I'm still thinking about codegen here.

bakura10:
Do you mean generating optimized factories that unroll the code? Yes, that's something I've thought about too ;)... And should not be too hard!

Ocramius:
Yes, I'd rather write an immutable service manager implementation that just skips all the method calls wherever possible (shared services) and inlines any non-shared service instantiation logic.

bakura10:
I've already removed nearly all mutable method calls. I'm even considering remove the configure method and being able to configure the SM only in constructor

library/Zend/ServiceManager/ServiceManager.php
```diff
@@ -9,1223 +9,216 @@
...
-     *                              Example: array('canonicalName', 'requestName')
-     * @param  bool          $checkAbstractFactories
-     * @param  bool          $usePeeringServiceManagers
-     * @return bool
+     * @param  string $name
+     * @return callable
+     * @throws InvalidFactoryException
      */
-    public function has($name, $checkAbstractFactories = true, $usePeeringServiceManagers = true)
+    protected function getFactory($name)
```
bakura10: Should we make it private?

library/Zend/ServiceManager/ServiceManager.php
```diff
@@ -9,1223 +9,216 @@
...
-     * @param  string   $cName
-     * @param  string   $rName
-     * @throws Exception\ServiceNotCreatedException
-     * @throws Exception\ServiceNotFoundException
-     * @throws Exception\CircularDependencyFoundException
+     * @param  array  $options
      * @return object
      */
-    protected function createServiceViaCallback($callable, $cName, $rName)
+    protected function createDelegatorFromFactory($name, array $options = [])
```
bakura10: Should we make it private?

Martin-P:
> This is make possible because the interface of a factory now has the requested class name as second parameter.

That is already the case in the current implementation when using an invokable class as factory :wink:

bakura10:
What do you mean? What the new implementation allows is being able to pass different service names to same factory:

```php
return [
    'service_manager' => [
        'factories' => [
               MyService1::class => MyFactory::class,
               MyService2::class => MyFactory::class
        ]
    ]
];
```

This is useful when you have hard-dependencies for your object, but that the construction is similar. Previously it needed an abstract factory, which is highly inefficient. Now, you can have a simple factory and map multiple services to it. As far as I know, it is not possible currently.

jaapio:
No tests? ;-)

Can you explain why the abstract factories shall be removed? The factory it self was not very inefficient, depending on what you did in the "canCreateServiceWithName". Is it the code that has to figure out the call to the abstract factory?

bakura10:
Tests will come :).

Regarding abstract factories, the main problem comes from the fact that the process of creating object through abstract factories comes at the end of the process, after the "fast paths". If you have only 6 or 7 abstract factories, it will need to iterate through them all. Even if that's pretty fast, you must always make sure that "canCreateServiceWithName" is highly optimized.

However, I've realized that one of the use case of abstract factories is actually create multiple services that are known in advance, but follow the same creation pattern. For instance in my code I have 7 services that are created exactly the same (I have just made 7 services to have smaller services). Currently, I'm forced to create 7 different factories, OR create an abstract factory. With this new approach, because you receive the class name into the factory, you can use this config:

```php
return [
    'service_manager' => [
         'factories' => [
              Service1::class => MyFactory::class,
              Service2::class => MyFactory::class
         ]
    ]
];
```

The factory looks like this:

```php
class MyFactory implements FactoryInterface
{
    public function __invoke(ServiceLocatorInterface $sl, $className, array $options = [])
    {
           $dependency1 = $sl->get(MyDependency::class);
           return new $className($dependency1);
    }
}
```

Abstract factories are still here, and must be reserved for creating services that you do not know in advance.

bakura10:
I have added back initializers and have fixed the delegators, that now work as expected. This new implementation now have nearly feature-parity, while being substantially faster and being 4 times smaller in term of code size :D.

jaapio:
I can remember a discussion where somebody suggested to remove the initializers since they are a performance bottleneck. initializers can be replaced by a delegator factory? So you don't have to loop over all initializers on very get call.  But that might also be up to the users of the servicemanager...

bakura10:
While I don't like initializers, there are a few places where they are useful: injecting an evm by implementing EventManagerAwereInterface and injecting a service locator using ServicelocatorAwareInterfacd.

While you can inject them yourself explicitly, this is annoying fro those ones. We should discourage using initializers in general, but some use cases are hard to replace without that.

Also, if we manage to compile the SM, performance cost of initialized will be 0

Martin-P:
> What do you mean? What the new implementation allows is being able to pass different service names to same factory

I know and that is already possible in the current version of ZF2:
```php
<?php

namespace MyNS\Factory\View\Helper;

use MyNS\Service\Foo;
use Zend\ServiceManager\ServiceLocatorInterface;

class FooFactory
{
    /**
     * @param ServiceLocatorInterface $serviceLocator
     * @param string                  $canonicalName
     * @param string                  $requestedName
     */
    public function __invoke(ServiceLocatorInterface $serviceLocator, $canonicalName, $requestedName)
    {
        // Do something
    }
}
```
Just a note that it is not a new feature, but an already existing one.

bakura10:
Hmm i didn't know that. It seems
That the service locator now give the name even for factory, but as this is not in the interface, I more consider that as a hack :p.

jaapio:
it is a hack, to be able to use closures in your config :-)

Martin-P:
> I more consider that as a hack :p.

Tell that to @weierophinney :wink:
http://blog.alejandrocelaya.com/2014/10/09/advanced-usage-of-service-manager-in-zend-framework-2/#disqus_thread

bakura10:
This PR just enforces this right into the interface :). At this point it cannot be changed in ZF2 because of BC, :)

demichl68:
$canonicalName and $requestedName are also passed to the createService Method.
The FactoryInterface just hides these params and it itsn't documented anywhere ....

```php
class Factory implements FactoryInterface
{
    public function createService(ServiceLocatorInterface $sl, $canonicalName = null, $requestedName = null)
    {
        // ...
    }
}
```

bakura10:
I see. I completely forgot/didn't know this @demichl68 . In all cases, the PR makes this clear (and also simplify it because this refactor remove the distinction between $canonicalName and $requestedName: you always receive a FQCN), and instead adds a new options array that fix a ugly hack of ZF2 (where you had to implement MutableCreationsInterface....)

library/Zend/ServiceMananger/ServiceManager.php
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Get a factory for the given service name
+     *
+     * @param  string $name
+     * @return callable
+     * @throws ServiceNotFoundException
+     */
+    protected function getFactory($name)
+    {
+        $factory = isset($this->factories[$name]) ? $this->factories[$name] : null;
```
bakura10: Should we have a second cache for factories, that would map a factory string name to an instance? I'm considering this because of the new InvokableFactory, you will have as many instantiated InvokableFactory than you have of object using it. Having said that, the cost of an object like a factory should be pretty low, and you don't have thousands of invokables, anyway.

tux-rampage:
> Yes, I'd rather write an immutable service manager implementation that just skips all the method calls wherever possible (shared services) and inlines any non-shared service instantiation logic.

I'm afraid that this would take away a lot of flexibility.

I always disliked the declare everything and create dozends of factories when there is only a need for a couple and the rest could be done via dependency injection. The current attempt for di is half hearted, and I'd rather see it integrated even better. Most services have dependencies, and writing a factory fetching other instances by class name and injecting them is a waste of time if this could be automated.

If you're really going to do this, create a service manager interface or rely on service locator interface only. Never check for the ServiceManager to allow users to replace the default SM with an implementation that might fit their use case better.

And I could imagine that there are use cases for aliases in the ServiceManager.

Additionally: Where (in terms of filesystem) and when do you want to generate the SM? This should not be at runtime for produvtion. Consider that any sane setup would prevent writing into the application directory for security reasons. Treating the whole application directory as writable is a bad attitude. Also think of apps packaged in a phar.

Take the ZendServer deployment as a good example. The whole application directory ist not writable to the php/webserver user.

Generated code is fine - but it has to be well thought, because it will create vulnerabilities (code injection). Any code you load from writable directories ist potentially compromised.

There should be a compiler for production and the generated code should be immutable. So don't hardcode data/ (besides the horrible assumption the cwd is always the place you're expecting). The application/developer should always have the chance to specify the location.

Doctrine is a good example: It allows codegen at runtime for testing/dev purposes and offers a shell command for pre-generating the code. The target directory for codegen is configurable.

Maybe the generated SM/Factories would be better only an additional option to give the developers the choice which solution fits best.

bakura10:
Hi @tux-rampage ,

As of today, the implementation does not remove many features and does not really change usage. Actually, it can even reduce the number of needed factories by explicitly allowing you to have a class name in the factory, so you can map multiple service names to the same factory (I know, it is possible now but it's not officially supported in the interface, now it will be).

I agree that most services have dependencies, so you have to write one factory per service (mostly). This means a lot of code, and a lot of boring tests to write if you want to tests your factory. HOWEVER, factories are always stupidly simple to write, and at least they allow you to exactly know what happen during the creation process, instead of doing a component doing some magic. For performance it's also quite good (and the new SM performance is really really fast contrary to ZF2).

> And I could imagine that there are use cases for aliases in the ServiceManager.

Aliases is not completely gone. I've decided to remove them from the ServiceManager though (and it was @Ocramius idea too) for several reasons:

* When we wrote ZF2, we had to support PHP 5.3. At the beginning, we were unsure about what to use as service key. ZF2 is using a mixed between FQCN and short names (like "Router", "ViewManager"...). However, the fact of using the FQCN emerged, as it is always simpler and allows to immediately know which object is created, instead of an opaque "Router". As ZF3 will be PHP 5.5+, we have access to ::class, and this is awesome because it provides you a quick syntax and also powerful refactoring tools in IDE (which was not the case when using strings). The fact that I removed aliases should encourage the paradigm of using ::class keyword, and make everyone's code more coherent.

* Aliases introduced an additional overheads: you have to do an additional while loop for each service you create. While not terribly expensive, it can rapidly adds-up for hundreds of objects. And for a critical component like SM, the performance must be tuned and watched. As I said, for most cases, the FQCN is a better choice than aliases. However, there is one case where aliases is useful: some plugin managers like view helper, controller plugin manager. You definitly want to do $this->identity() in your view rather than $this->{'Zend\View\Helper\IdentityHelper'}() (not even sure this syntax is valid).
To that extent, I have introduced a specialized plugin manager that is aware of aliases. You can get best of both worlds: best performance for main service locator and plugin managers that don't need aliases, and flexibility for others.


For the compiler thing, obviously it won't be a runtime compiler. It will work exactly as how Doctrine generation work, and it will be saved to an optimized file that will be used. However, I have tried this and it is far from easy, so we'll likely need some crazy @Ocramius skills to do that.


Finally, for the fact that you need to write factories, I actually tried to use some kind of automatism, through Reflection. I thought that because you had access to the various types through reflection, I could create objects without having to write factories. However, it's not that easy for a lot of reasons:

1. This only works for objects, not interfaces. As I inject interfaces a lot, this is a case that cannot be handled automatically.
2. You can't do anything with scalar options.
3. Plugin managers make things very hard, because any dependencies may actually not come from the main service locator, but from a plugin manager. And knowing which plugin manager the object come from is not that easy. We could argue that plugin managers could be removed in favour of a unique service manager, but plugin managers are really nice and I'd like to keep them.

At the end, there are so many cases where "magic" cannot work, that you end up writing factories 80% of the time anyway. The performance overhead due to the Reflection is also not really acceptable for such a component, tbh.

While it seems appealing, in practice it is not. Speaking about, other people came to the same conclusion. Aura.Di used to do auto-resolution, and they admitted that it was a not-so good idea too: https://github.com/auraphp/Aura.Di#a-note-about-auto-resolution


To conclude, yes, factories are annoying to write, but they are not complex and easy to test, and the most efficient way (except compiling the service manager, but this is another story).

tux-rampage:
Well I've got your point. But I have to disagree about the 80% of writing factories.
I coupled di and the service manager for zf2 and endet up in writing factories in only 10% of the cases. It's a matter of design.

And I personally like the approach of using and encouraging `::class`. It has so much benefits, you named them. 
But we should keep the possibility to define services which are no real or invalid (i.e. containing a dot) class names. This still seems to be working. Code should be reusable and subtyping a class just for retrieving a different parameterized shared instance (i.e. another DocumentManager or db adapter) is awful.

The main question from my side is: How do you handle existing classes that are not explicitly defined?

ZF2 had abstract factories.

The easiest (and fastest) approach would simply be to instanciate them with `new` or fall back to a generic factory. 

But why not giving developers a choice here to couple it with Zend\Di (A component I like to have some refactoring for as well) which kicks in when a class is not defined? I think of it in a way where the service manager could be exchanged with a different implementation. That would require changes of typehints to the SM class to an interface. If you're breaking BC it's a good time to clean up a bit more.

So purists can completely ignore di and have max performance and the lazy folks like me can keep things simple and benefit from di by exchanging the servicemanager. I see no reason for preventing possibilities.

About the compiler thingey: Well this would need some specification as well. Maybe it would even be a waste of time and resources if the performance win is not overwhelming. Generated code has cons (it's hard to debug for example). And at the moment I don't see a point where generating code makes the SM more efficient but less maintainable. Maybe I just didn't get the idea here. Please enlighten me.

tux-rampage:
I just checked your implementation and my concerns about servive names and abstract factories seem to be off the table.

Di could be coupled via an abstract factory.
If I find some time I might offer some refactoring to the di component which will be able to do so.

bakura10:
To be honest I cannot really talk about Di because I have never used this component (except at the very beginning of ZF2 development when we didn't have the Service Manager).

Actually, regarding the key, I want to encourage people to use ::class, but it's just a key. You can still use "doctrine.foo" if you want, as soon as it resolves to a factory that is able to create it. The only use case is for invokables.

In this refactor, everything is a factory (it simplifies implementation and avoid many checks during creation). 

Before:

```php
return [
        'invokables' => [
		MyInvokable::class => MyInvokable::class
	],
	'factories' => [
		MyObject::class => MyObjectFactory::class
	]
]; 
```

After:

```php
return [
	'factories' => [
		MyInvokable::class => InvokableFactory::class,
		MyObject::class => MyObjectFactory::class
	]
];
```

InvokableFactory is a built-in factory, and it expects that the service key is actually the FQCN of the class.

> Code should be reusable and subtyping a class just for retrieving a different parameterized shared instance (i.e. another DocumentManager or db adapter) is awful.

Actually, this refactor comes with a solution to that. The `get` method always accept an optional second parameter for options. Contrary to ZF2 where it was only transferred to invokable, in this refactor it is transferred everywhere (even for delegators). This means that your factory is aware of the creation (it is the third parameter of the factory interface), and you can therefore parameterized creation:

```php
$instance1 = $this->serviceLocator->get(MyObject::class, ['foo' => 'bar']);
```

The only limitation for now is that the service must be marked as non-shared (otherwise, the second time you call the get method, it will return the same instance, without going back to the factory. Actually, I think that we could be smarter here, and assume that if we have options, *there are chances* that it will be called with another set of options, and hence make this as a criteria to not cache the service. Just an idea, though.

bakura10:
Finally, regarding the compiler, I'm not convinced too. We are using ZF2 for several years, and this refactor is way faster (when I say way faster, it's WAY FASTER !!!), so the performance boost brought by this refactor should be already more than enough ;).

tux-rampage:
Looks good so far. 

Well as said I'd love to support DI here much better, but that's not the business of the ServiceManager, but the Di component. I'd rather see Di as (optional) addon to the ServiceManager.

Plus your implementation already offers a way to do that.

About the second $options parameter, I'm not sure if this would be an overkill here. The Di instance manager offers distinction between parameterized calls and it complicates things a lot.
Of course this is doable, but do we really need it?

Thanks for the effort so far, looks promising. :+1:

library/Zend/ServiceMananger/ServiceManager.php
tux-rampage: Maybe not a bad idea :+1:

library/Zend/ServiceMananger/Proxy/LazyServiceFactory.php
```diff
@@ -22,17 +21,17 @@
  *
  * @link https://github.com/Ocramius/ProxyManager/blob/master/docs/lazy-loading-value-holder.md
  */
-class LazyServiceFactory implements DelegatorFactoryInterface
+final class LazyServiceFactory implements DelegatorFactoryInterface
```
tux-rampage: Why are you making this final and properties private? Is there really a reason why this class must never be extended and (if it wasn't final) derived classes must never manipulate the properties?

bakura10: This is a factory, I do not think you actually need to extend it. And if you need, you should use a delegator.

As ProxyManager may evolve, declaring it final may actually allows us to do breaking changes inside the code without having to release a new major release.

Ocramius: @tux-rampage there's no real need to use inheritance in this case

tux-rampage: Thanks for the clarification.

library/Zend/ServiceMananger/Exception/ExceptionInterface.php
```diff
@@ -9,6 +9,9 @@
...
 namespace Zend\ServiceManager\Exception;
 
-interface ExceptionInterface
+/**
+ * Base exception
+ */
+interface ExceptionInterface 
 {
-}
+}
```
DASPRiD: You are missing newlines at end of file in multiple files.

library/Zend/ServiceMananger/composer.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "zendframework/zend-servicemanager",
+  "type": "library",
+  "license": "MIT",
```
DASPRiD: ZF is licensed under BSE-3-clause ;)

library/Zend/ServiceMananger/Exception/ExceptionInterface.php
bakura10: Grrr I should configure my IDE correcty :). Thanks.

bakura10:
I'm closing in favor of the component PR. Please make comments there: https://github.com/zendframework/zend-servicemanager/pull/2

PR #7137, opened 2015-01-19
Request from string

mathiasverraes:
``` php
<?php
$request = \Zend\Http\Request::fromString(
    "GET /?foo=bar HTTP/1.1\r\n\r\n"
);

var_dump($request->getQuery()->get('foo'));
// Expected "bar", got null
```

### DIFF
--- a/library/Zend/Http/Request.php
+++ b/library/Zend/Http/Request.php
@@ -93,6 +93,13 @@ class Request extends AbstractMessage implements RequestInterface
         $request->setMethod($matches['method']);
         $request->setUri($matches['uri']);
 
+        $parsedUri = parse_url($matches['uri']);
+        if(array_key_exists('query', $parsedUri)) {
+            $parsedQuery = [];
+            parse_str($parsedUri['query'], $parsedQuery);
+            $request->setQuery(new Parameters($parsedQuery));
+        }
+
         if (isset($matches['version'])) {
             $request->setVersion($matches['version']);
         }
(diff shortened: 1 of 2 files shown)


mathiasverraes:
I've no idea why travis fails :-(

anthonysterling:
It looks like [Zend\Http\Request::setUri](https://github.com/zendframework/zf2/blob/master/library/Zend/Http/Request.php#L94) delegates the parsing of Query to [Zend/Uri/Uri](https://github.com/zendframework/zf2/blob/master/library/Zend/Uri/Uri.php#L332).

I suspect it's getting passed `/?foo=bar` and not `?foo=bar` so this [pattern](https://github.com/zendframework/zf2/blob/master/library/Zend/Uri/Uri.php#L332) never matches.

Maybe.

mathiasverraes:
Did you find a place in the travis builds that fails?

texdc:
Not sure, but it I'd guess that there's an issue with Travis, not the code or build.

Ocramius:
I'll check the CS fixes locally: we handled this in `develop`, but not in `master`, so it's indeed a bit annoying to not be able to see the failures.

Ocramius:
Merged, thanks @mathiasverraes!

`master`: zendframework/zf2@7e704a7ef336804171020844e310516d7a2ef38c
`develop`: zendframework/zf2@bef1fc181b6e16a8155f34fa3cac098d56df88af

mathiasverraes:
YW, thanks for getting this in fast :-)

PR #7138, opened 2015-01-19
Email validator hostname check for not empty

freax:
Sometimes there is a warning because of empty `hostname` parameter in `checkdnsrr` function.

### DIFF
--- a/library/Zend/Validator/EmailAddress.php
+++ b/library/Zend/Validator/EmailAddress.php
@@ -408,6 +408,7 @@ class EmailAddress extends AbstractValidator
             }
 
             if (!$res
+                && !empty($hostname)
                 && (checkdnsrr($hostname, "A")
                 || checkdnsrr($hostname, "AAAA")
                 || checkdnsrr($hostname, "A6"))


Ocramius:
@freax please rebase and provide a test case for the hostname affected by this bug.

DASPRiD:
@Ocramius adding a hostname to the test suit which could fail any time? Sounds like a bad idea.

Ocramius:
I mean that data-providers need to be adjusted ;-)

freax:
I don't think this PR needs a test case, because it's nothing to break actually.
This condition fails if `hostname` will be empty anyway, but this PR provides no more thipple warning in this case.
I got this warnings twise and I can't reproduce it.
Actually, I think it's a pretty rare case when script can't get MX-record for hostname.

Ocramius:
> I don't think this PR needs a test case, because it's nothing to break actually.

code changes affecting branching of execution paths **always require** testing, regardless of the impact they have.

 > This condition `fails` if hostname will be empty anyway, but this PR provides no more thipple warning in this case.

And your test should actually try to reproduce the warning instead: reaching the point where you can reproduce the warning will also give you a better understanding the problem.

 > Actually, I think it's a pretty rare case when script can't get MX-record for hostname.

You can inject a fake hostname validator to get to that.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-validator/issues/99

PR #7141, opened 2015-01-20
Implement BcryptSha Hashing & Harden Verification Checks

padraic:
This PR, as discussed, implements several security improvements:
1. Offers a new Zend\Crypt\Password\BcryptSha class which is functionally identical to the Bcrypt class but will pre-hash incoming passwords using SHA256. This small step allows for passwords exceeding 72 bytes to have all 72 bytes utilised in a starting hash. At present, the use of Bcrypt will truncate any input to 72 bytes and effectively discard the remainder of the string. This is perfectly normal and valid behaviour in the bcrypt algorithm, but obviously not what users might expect when allowed to use long passwords.
2. It replaces the current set of identical comparisons with the pre-existing fixed time compare functionality to offset any risk of timing leaks about the hashes being compared.

Noting PR is against master at present.

### DIFF
--- a/library/Zend/Crypt/Password/BcryptSha.php
+++ b/library/Zend/Crypt/Password/BcryptSha.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Crypt\Password;
+
+use Traversable;
+use Zend\Math\Rand;
+use Zend\Stdlib\ArrayUtils;
+use Zend\Crypt\Hash;
+
+/**
+ * Bcrypt algorithm using crypt() function of PHP with password
+ * hashed using SHA2 to allow for passwords >72 characters.
+ */
+class BcryptSha extends Bcrypt
+{
+
+    /**
+     * BcryptSha
+     *
+     * @param  string $password
+     * @throws Exception\RuntimeException
+     * @return string
+     */
+    public function create($password)
+    {
+        $hashedPassword = Hash::compute('sha256', $password);
+        return parent::create($hashedPassword);
+    }
+
+    /**
+     * Verify if a password is correct against a hash value
+     *
+     * @param  string $password
+     * @param  string $hash
+     * @throws Exception\RuntimeException when the hash is unable to be processed
+     * @return bool
+     */
+    public function verify($password, $hash)
+    {
+        $hashedPassword = Hash::compute('sha256', $password);
+        return parent::verify($hashedPassword, $hash);
+    }
+
+}
--- a/library/Zend/Crypt/Password/BcryptSha.php
+++ b/library/Zend/Crypt/Password/BcryptSha.php
@@ -3,7 +3,7 @@
  * Zend Framework (http://framework.zend.com/)
  *
  * @link      http://github.com/zendframework/zf2 for the canonical source repository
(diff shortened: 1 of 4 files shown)


library/Zend/Crypt/Password/BcryptSha.php
```diff
@@ -0,0 +1,51 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
```
samsonasik: 2015

library/Zend/Crypt/Password/BcryptSha.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * BcryptSha
+     *
+     * @param  string $password
+     * @throws Exception\RuntimeException
+     * @return string
+     */
+    public function create($password)
+    {
+        $hashedPassword = Hash::compute('sha256', $password);
```
Ocramius: No need for this assignment

library/Zend/Crypt/Password/BcryptSha.php
```diff
@@ -0,0 +1,51 @@
...
+     * Verify if a password is correct against a hash value
+     *
+     * @param  string $password
+     * @param  string $hash
+     * @throws Exception\RuntimeException when the hash is unable to be processed
+     * @return bool
+     */
+    public function verify($password, $hash)
+    {
+        $hashedPassword = Hash::compute('sha256', $password);
```
Ocramius: No need for this assignment

tests/ZendTest/Crypt/Password/BcryptShaTest.php
```diff
@@ -0,0 +1,127 @@
...
+    public $bcryptPassword;
+    /** @var string */
+    public $password;
+
+    public function setUp()
+    {
+        $this->bcrypt   = new BcryptSha();
+        $this->salt     = '1234567890123456';
+        $this->password = 'test';
+        $this->prefix = '$2y$';
```
Ocramius: Alignment

tests/ZendTest/Crypt/Password/BcryptShaTest.php
```diff
@@ -0,0 +1,127 @@
...
+use Zend\Config\Config;
+use Zend\Crypt\Password\Exception;
+
+/**
+ * @group      Zend_Crypt
+ */
+class BcryptShaTest extends \PHPUnit_Framework_TestCase
+{
+    /** @var Bcrypt */
+    public $bcrypt;
```
Ocramius: These should be `private`

tests/ZendTest/Crypt/Password/BcryptShaTest.php
```diff
@@ -0,0 +1,127 @@
...
+
+namespace ZendTest\Crypt\Password;
+
+use Zend\Crypt\Password\Bcrypt;
+use Zend\Crypt\Password\BcryptSha;
+use Zend\Config\Config;
+use Zend\Crypt\Password\Exception;
+
+/**
+ * @group      Zend_Crypt
```
Ocramius: Please add coverage annotations

weierophinney:
For the inevitable commenters who have read: http://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html — this patch follows the practices outlined in Anthony's post already. `Hash::compute()` takes up to three arguments, the last of which is `$output`, which is a constant indicating whether the returned hash should be in string or binary format. By default, string is used, which is equivalent to passing a boolean `false` to the `hash()` function's third argument, making the returned strings safe for re-hashing.

weierophinney:
Merged to develop for release with 2.4; @padraic — please provide documentation in the zf2-documentation repo. :)

Issue #7143, opened 2015-01-21
[ZF3][kind-of RFC] Event Manager, attempt n°4


bakura10:
# Event Manager

Hi everyone,

This is a new thread that superseeds #5693. A lot of things will be duplicate, but I want to give a few more options and bring more experience based on my last project that heavily relied on event manager.

## The shared event manager

Zend Framework 2 is the only framework that contains a shared event manager, and what makes it so powerful. However, it comes with a few caveats:

* People don't understand why there are two ways to attach listeners (to the shared event manager and the normal event manager).
* It comes with a heavy overhead in performance.
* The shared event manager uses a concept of identifiers, that is both obscure and can lead to hard-to debug bugs (more on that later).
* It forces to instantiate all the listeners on each requests, which introduce a large overhead if you have listeners with a lot of dependencies.

Basically, the shared event manager as it acts as a kind of "singleton", and we wanted to avoid the `StaticEventManager` that is still present in ZF2.

But I realized two things:

* I always use the shared event manager. I've take a look at my code, and all the listeners I attach are attached to the shared event manager.
* I nearly never use the concept of identifiers, but I am forced to indicate one (or using a wildcard).

In this app, I came accross a subtle bug that took me hours to find, and that was made possible only because of the fundamental concept of shared event managers. Basically, I had a service, that I call "Service1", that triggered an event:

```php
class Service1 implements EventManagerAwareInterface
{
	use EventManagerAwareTrait;
	
	public function doSomething()
	{
		// ...
		$this->getEventManager()->trigger('foo', $event);
	}
}
```

Then, I have a listener aggregate that attaches a callable to this event:

```php
class MyListener extends AbstractListenerAggregate
{
	public function attach(EventManagerInterface $evm)
	{
		$sem = $evm->getSharedManager();
		$sem->attach(Service1::class, 'foo', [$this, 'doSomething']);
	}
}
```

However, my Service1 became too complex, so I decided to split it into two services, and the event manager got triggered in the new "Service2". Consequence: because the listener was attached to the identifier "Service1", it was no longer executed. No error, nearly impossible to test in unit test (you need complex functional tests), and can be even completely silent for ages if what the listener do is sending an email, for instance.

[...]

PR #7145, opened 2015-01-22
[WIP] Event Manager Refactor

bakura10:
Hi everyone,

This is a very quick prototype for #7143. It removes completely the shared event manager, and add the idea of having lazy listener. Actually, I've simplified a lot the interfaces, to remove completely the "laziness": it's now just a detail of implementation for the EventManager, but it's not in the interface anymore.

Basically, the default implementation of EventManager accepts a ListenerPluginManager. The ListenerAggregateInterface now has a simple "attachAggregate" that is static, where you can attach event using a class name:

``` php
class UserListenerAggregate implements ListenerAggregateInterface
{
    static public function attachAggregate(EventManagerInterface $eventManager)
    {
        $eventManager->attach('user.create', [static::class, 'sendRegistrationEmail'], 15);
        $eventManager->attach('user.delete', [static::class, 'removeFromNewsletter'], 100);
    }

    // Your methods
}
```

The listener can have any hard dependencies, but contrary to now, it will be created lazily. Actually, if the event maanger detect a string name, it will mark the listener as "lazy", to indicate that it must fetch it from the plugin manager.

Additionally, the event manager uses generators and listeners are created only when ask, so for instance, if propagation is stopped, any other listener that may need to be created won't be created at all from the plugin manager.

This is still rough and I suspect there are a lot of potential optimizations, but the implementation is simple, remove any reference ot the laziness and is way simpler to undesrtand than ZF2 one.


library/Zend/EventManager/EventInterface.php
```diff
@@ -1,96 +1,39 @@
...
 namespace Zend\EventManager;
 
-use ArrayAccess;
-
 /**
- * Representation of an event
+ * EventInterface
  */
-interface EventInterface
+interface EventInterface
```
keradus: extra space

bakura10:
I've upgraded the implementation to my Service Manager refactor, and I have also implemented the detach method :).

library/Zend/EventManager/Event.php
```diff
@@ -117,93 +50,34 @@ public function getParams()
...
     {
-        if (is_array($this->params) || $this->params instanceof ArrayAccess) {
-            // Arrays or objects implementing array access
-            $this->params[$name] = $value;
-        } else {
-            // Objects
-            $this->params->{$name} = $value;
-        }
-        return $this;
+        return isset($this->params[$key]) ? $this->params[$key] : $defaultValue;
```
marc-mabe: This requires `array_key_exists()` as a param could be set to `null` explicit.

mwillbanks:
What is the performance of this vs. the previous refactor?  Is there still going to be a far quicker and faster EM ?  :)

bakura10:
Yes it's even faster because it has removed the shared manager and introduced lazy listener. So I expect to to be even faster than faster.

Envoyé de mon iPhone

bakura10:
@mwillbanks , I've done a small benchmark with old and new benchmark, on PHP 5.6 and PHP 7. The benchmark is attaching 20 listeners, detaching one listener, triggering 5 times different events. Please remind also that the main advantage of this refactor is to also have a lazy listeners, which will be HUGE in memory and costly object creation, and will allow to attach much more listeners without impacting performance.

For ZF2 EVM :

PHP 5.6

    Bench\EventManagerBenchmark
    Method Name         Iterations    Average Time      Ops/second
    -----------------  ------------  --------------    -------------
    attach20Listeners: [1,000     ] [0.0012481315136] [801.19762]

PHP 7

    Bench\EventManagerBenchmark
    Method Name         Iterations    Average Time      Ops/second
    -----------------  ------------  --------------    -------------
    attach20Listeners: [1,000     ] [0.0001788964272] [5,589.82656]




ZF3 Refactor : 

PHP 5.6

    Method Name         Iterations    Average Time      Ops/second
    -----------------  ------------  --------------    -------------
    attach20Listeners: [1,000     ] [0.0002542481422] [3,933.16542]

PHP 7

    Method Name         Iterations    Average Time      Ops/second
    -----------------  ------------  --------------    -------------
    attach20Listeners: [1,000     ] [0.0000408854485] [24,458.57971]

macnibblet:
@bakura10 bloody hell, that's a really nice performance boost :+1:

bakura10:
Being able to attach many listeners without having to have the cost on each request is even better, actually. In my app, I didn't do a proper benchmark, but it incurs at least 5-10 ms overhead on each request, while all the listeners are only executed like on 0,1% of each request :/. With this refactor, no issue :D.

bakura10:
@Ocramius wanted me to separate attach and trigger. Here are the results:

ZF2 EVM

PHP 5.6

    Bench\EventManagerBenchmark
    Method Name          Iterations    Average Time      Ops/second
    ------------------  ------------  --------------    -------------
    attach20Listeners : [1,000     ] [0.0005693485737] [1,756.39326]
    trigger20Listeners: [1,000     ] [0.0010452888012] [956.67341]


PHP 7

    Bench\EventManagerBenchmark
    Method Name          Iterations    Average Time      Ops/second
    ------------------  ------------  --------------    -------------
    attach20Listeners : [1,000     ] [0.0000875046253] [11,427.96734]
    trigger20Listeners: [1,000     ] [0.0001518340111] [6,586.13965]



ZF3 EVM

PHP 5.6

    Bench\EventManagerBenchmark
    Method Name          Iterations    Average Time      Ops/second
    ------------------  ------------  --------------    -------------
    attach20Listeners : [1,000     ] [0.0001252155304] [7,986.22980]
    trigger20Listeners: [1,000     ] [0.0002773256302] [3,605.86939]

PHP 7

    Bench\EventManagerBenchmark
    Method Name          Iterations    Average Time      Ops/second
    ------------------  ------------  --------------    -------------
    attach20Listeners : [1,000     ] [0.0000128707886] [77,695.31713]
    trigger20Listeners: [1,000     ] [0.0000429670811] [23,273.63123]

bakura10:
@Ocramius expressed the need on identifiers. I'm personally strongly against identifiers considering the hard life it caused to me debugging nasty things (as I said in the issue, attaching for a given service identifier, then moving the code that triggers into another service, and forgetting to change the identifier in the attach... identifiers are way to fragile).

However I could understand the need to scope even more which listeners are called, but I want this to be just optional (use it if you need it, but for most people they won't use it because it's overkill for most cases). The idea is to have an optional parameter when attaching and triggering:

```php
$evm->attach('foo', $callable, $priority, ['identifier1', 'identifier2']);
$evm->trigger('foo', $event, ['identifier3']); // won't execute any listener
```

I'm just want to make sure that this does not add too much performance overhead because of having to check the listeners.

What do you think?

Ocramius:
> identifiers are way to fragile

Identifiers are routing: wrong routing is still wrong routing :-)

Ocramius:
>  I could understand the need to scope even more which listeners are called

My problem is not to scope event listeners, but to still aggregate in the same way the `SharedEventManager` does it.

The typical use-case is:

```php
$shippingProvider1 = new UpsShippingProvider($evm1);
$shippingProvider2 = new DhlShippingProvider($evm2);

// triggering shipping provider 1 payment logic and product inventory 1 availability update
$shippingProvider1->ship($order1); 
// triggering shipping provider 2 payment logic and product inventory 1 availability update
$shippingProvider2->ship($order2);
```

As you can see, there is a shared listener involved, but the event triggering is context-specific, which means that `$evm1 !== $evm2`.

Nothing to do with identifiers.

ltouro:
@Ocramius 

What if you have two or maybe three different events? It makes sense? It would be up to the provider to trigger the correct events.
```
// $evm1 has listeners attached to ship, ship.dhl and ship.ups
$shippingProvider1 = new UpsShippingProvider($evm1);
$shippingProvider2 = new DhlShippingProvider($evm1);

// triggers: ship and ship.ups events
$shippingProvider1->ship($order1); 
// triggers: ship and ship.dhl events
$shippingProvider2->ship($order2);
```

Ocramius:
> What if you have two or maybe three different events?

The events are the same: that's where your decoupling/abstraction actually lives.

ltouro:
I see Ocramius, it makes sense.Thanks! :]

bakura10:
First, I personally find this is a wrong usage of event manager. Updating an availability is something critical. What if a listener stop propagation before it reaches the availability service? I really think that for that kind of use case, you should have a coordinator service that explicitly propagate the various steps (attaching to ship, updating stock...), only non critical things like sending confirmation emails should be handled using the EVM, as they are the consequence of the full process. I tend to remove having a lot of related logic using only events. So much can happen that I find it quite fragile.

I can see the use case though, but still, I don't think that the event manager should have so many complexity by default. This is a rather extreme use case. I REALLY REALLY want to have something that is intuitive for most cases. In my code, 100% of the time (and I know I'm not the only one), I'm only using the shared event manager because I don't have any other way, but what I'm only interesting is triggering an event and have all the listeners do something. Period. And from what I've read, it seems a lot of people are doing like me.

The Shared Event manager with identifiers is the most confusing stuff. Why can't we remove the shared event manager and offer a simpler pattern by default, if what I'm doing is ALWAYS retrieving the shared event manager?

To achieve your use case, we could add some kind of ComposableEventManager, where the event maanger would compose another (main) event manager, and have its own set of listeners, so you could achieve kind of the same result (but with the advantage of keeping the exact same interface as event manager, and don't introduce the shared event manager:

```php
class ComposableEventManager extends EventManager
{
    private $eventManager;

    public function __construct(EventManager $eventManager)
    {
         $this->eventManager = $eventManager;
    }

    public function getListeners($eventName)
    {
        // would merge the local listeners and the one from the composed event manager
    }
}
```

But once again, I'm 100% sure that typical use case does not need all this complexity. What you want is typically: attach listeners, trigger, make sure all listeners execute.


Regarding the identifiers, I'm sorry, but this is not really "routing". Really, this IS REALLY FRAGILE. Let's take an example: a third-party module triggers an event in ServiceA. This service therefore adds the "ServiceA" identifier to its event manager. The ServiceA could even be final.

You listen to the service manager with ServiceA identifier. Then this module moves the trigger logic to ServiceB (it would not even be a BC break from the module point of view, it has only moved a part of the logic, that could even reside in a private method, to another service). No BC break in the module, and boom... your app break. Not only does it break, but it breaks completely silently: the listener will never be executed because of the identifier. If this is non critical piece like emails, it could even break silently for months without you noticing it. And no unit test could find this. And exactly what happen recently to me (except that it was in a code of mine instead of a module).

This is why identifiers MUST be removed. They caused bugs that are nearly impossible to track.

bakura10:
What you want can be achieved with the "identifiers" (or more exactly, "tagged"). For instance, Ups will trigger:

```php
$this->eventManager->trigger('product.ship', $event, ['ups']);
$this->eventManager->trigger('inventory.update', $event);
$this->eventManager->trigger('product_sent.email', $event);
```

For inventory.update and product_sent.email, you will have listeners to it, for product.ship, you may have a listener like this:

```php
$this->eventManager->attach('product.ship', $callable, $priority, ['ups']);
```

The difference in "those" identifiers is that they are part of the trigger and attaching signature. If properly unit-tested, changing the identifier here IS a BC break and can be spotted. Even if you move the trigger code from one service to another, it's still triggering for the given identifier in the parameter (this is not the case for identifiers tied to an event maanger.

stefanotorresi:
probably a stupid question but, since the event name is namespaced, how is:

```php
$this->eventManager->trigger('product.ship', $event, ['ups']);
[...]
$this->eventManager->attach('product.ship', $callable, $priority, ['ups']);
```

different from

```php
$this->eventManager->trigger('ups.product.ship', $event);
[...]
$this->eventManager->attach('ups.product.ship', $callable, $priority);
```

?

Both the trigger the listener have to be aware of the context so, as far as I understand, the main difference with the current identifier system is that the identifiers are defined at `EventManager` instantiation, and then specified by the listeners; the trigger on the other hand has no such context awareness. 
But then again, sometimes the `Event::$target`, which should represent the context of the trigger, is changed. Sometimes the whole event instance is instantiated from scratch for an event, while sometimes the same event instance is used throughout different triggers, and it's state is changed instead. What I would like is a stricter convention about all this.
Someone suggested immutable events in the other thread (#5693), should this option be just disregarded? FWIW, I think there is value in it.

Anyway, shouldn't at least `Event::$target` be immutable? Shouldn't it be the only entry point to the event context? isn't the event itself just a bag that carries a context throughout the cycle of a set of listeners?

Honourable example: using two different `DISPATCH` in `Zend\Mvc`, changing `Event::$target` from `Application` the the controller instance. Such context change drove me mental a number of times. These are basically two different event cycles, but using the same manager and event instances, and using the results in a different way (i.e. using both the results queue and `Event::$result`, with no clear clue for the listener on what is used first).
Furthermore, why should I be able to grab an event instance outside a trigger cycle and fiddle with it?
Am I missing some compelling point for this behaviour?

The way I see it is that a new event instance should created each time a `trigger` is invoked.

To instantiate and run listeners only in some contexts, an assertion callback could be run on the event target before even instantiating the listener, i.e.:

```php
$userListener = function($event)
{
    // do something related to UserController only, referenced in $event->getTarget()
}

$targetAssertion = function($target) {
    return $target instanceof UserController;
}
$evm->attach('dispatch', $userListener, $targetAssertion);
```

does this make any sense?

codeliner:
@bakura10 The static ListenerAggregate::attachAggregate method still has performance impact. The listener aggregate needs to be autoloaded. 
I use an event bus implementation (comparable with an application wide EM) which uses a configurable event routing map, something like:

```php
//config/autoload/event_bus.config.php
return [
  'event_bus' => [
    'event_routing_map' => [
      My\Event::class => [
        My\Listener1::class,
        My\Listener2::class,
      ] 
    ]
  ]
];
```

Would it make sense to support something similar with the refactored EM?

bakura10:
I honestly don't think that the auto loading will have many impact on performance, or so negligible you'll need hundreds of listeners to start having an impact.

I'm not against the idea but the event manager should stay simple. But that is an interesting idea (except that it is even more config arrays...).

I'm not sure how your thing would work with wildcard listeners though.

Envoyé de mon iPhone

codeliner:
In a CQRS + ES architecture  (the origin of the idea) you have a lot of event listeners :-), but that can not be directly mapped to the usage of the EM.

> that it is even more config arrays...).
I consider this a good thing :-). That's the way to set up things in ZF2, so why not use it for listeners too? 

> I'm not sure how your thing would work with wildcard listeners though.
I have a special regex event router to handle wildcards, but normally I favour explicit event routing to keep things simple :-)

bakura10:
I'm definitely not against introducing some event bus/CQRS into it if it makes sense or if this is a better way to use event manager.

We just need to see how it could work internally in the framework.

Anyway I'll soon move this PR to the event manager component so things can be discussed.

Envoyé de mon iPhone

texdc:
Please keep in mind that an event-sourced architecture is geared towards multiple disparate processes reacting to domain events, not necessarily on the same box or application.  Framework events, on the other hand, occur and are reacted to in the same process.  So, there must be flexibility for both implementations.  i.e. `UserHasSubscribed` is not the same as `ModuleLoaded`.

codeliner:
@texdc :+1: I totally agree and would never say something different. But I think configurable listeners based on a routing map can work for the EM too and align the set up with the rest of the framework.

Btw. support for immutable events (like mentioned by @stefanotorresi)  - maybe by use either EventInterface or ImmutableEventInterface (like with \DateTime and \DateTimeImmutable) - would allow to use the EM as a DomainEventDispatcher ... But I'm just thinking loud. Maybe it's better to keep both concepts separate.

bakura10:
Please discuss about this there now: https://github.com/zendframework/zend-eventmanager/pull/2

I'm closing :).

Issue #7148, opened 2015-01-22
ModuleLoader not capable of loading modules contained within an application.phar


basz:
Although the ModuleLoader contains a loadModuleFromPhar method, i've noticed it does not work for modules that are included within a packaged application.

resulting in 'Module (Application) could not be initialized.' after packaging

Ocramius:
Handled in #7151

PR #7151, opened 2015-01-24
Applications packaged as not capable of loading modules contained within the archive

basz:
The ModuleAutoloader fails to load modules when the modules are part of the phar that contains the application.

A realpath check on phar://path/to/phar.phar/modules/Application/Module.php will return false. There no such thing as a realpath on the real filesystem for these kind of files.

This fix uses the Phar::current(true) method to find out the correct basepath of the application. Then regular loadModuleFromDir is after minor adjustment capable to load the module.php 'fromDir'. 

Note existing tests still pass, but I did not write a new one for this addition. I wouldn't know how to do that for such a case.

original #7148


Ocramius:
Requiring a test case...

library/Zend/Loader/ModuleAutoloader.php
```diff
@@ -232,12 +242,19 @@ public function autoload($class)
      */
     protected function loadModuleFromDir($dirPath, $class)
     {
-        $file = new SplFileInfo($dirPath . '/Module.php');
-        if ($file->isReadable() && $file->isFile()) {
+        $modulePath = $dirPath . '/Module.php';
+        if (substr($modulePath, 0, 7) == 'phar://') {
```
keradus: please use strict comparision

basz:
@Ocramius : haz test

Ocramius:
@basz thanks, looks much better!

basz:
@Ocramius out of curiosity, if this lands, will it land in master when reviews are done or are you waiting with merging moments before 2.3.5 is released?

Ocramius:
@basz I'm waiting for a legendary magical beast called "free time" ;-)

I merge when I have a bit of dedicated time to do so. Yes, merge will happen before 2.3.5/2.4

basz:
fat

hook your magic method __sleep to __clone and time is multiplied.

thanks

Ocramius:
@basz merged, thanks!

`master`: zendframework/zf2@d035b6ad93721497ae9314c168c9910c113a72a2
`develop`: zendframework/zf2@5aed8a9d77a8901d3abe9ee5f229dc1c26e817e7

PR #7152, opened 2015-01-24
Fixes #4936

### LINKED ISSUE #4936: Zend\Mvc\Router\Http\Chain::assemble calls key on PriorityList
On line 159 of Zend/Mvc/Router/Http/Chain.php:
```php
$lastRouteKey = key($this->routes);
```

This returns something like: "\u0000*\u0000sorted" on my system (PHP 5.4)

This fails the test on line 168:

```php
$chainOptions['has_child'] = ($hasChild || $key !== $lastRouteKey);
```

As key is a numeral value.

A fix would be to replace it with:

```php
$routes = ArrayUtils::iteratorToArray($this->routes);
end($routes);
$lastRouteKey = key($routes);
```

kusmierz:
It will fix #4936 issue.

### DIFF
--- a/tests/ZendTest/Mvc/Router/PriorityListTest.php
+++ b/tests/ZendTest/Mvc/Router/PriorityListTest.php
@@ -9,9 +9,8 @@
 
 namespace ZendTest\Mvc\Router;
 
-use PHPUnit_Framework_TestCase as TestCase;
 use Zend\Mvc\Router\PriorityList;
-use Zend\Stdlib\ArrayUtils;
+use PHPUnit_Framework_TestCase as TestCase;
 
 /**
  * @group      Zend_Router
--- a/tests/ZendTest/Mvc/Router/PriorityListTest.php
+++ b/tests/ZendTest/Mvc/Router/PriorityListTest.php
@@ -9,7 +9,6 @@
 
 namespace ZendTest\Mvc\Router;
 
-use Zend\Mvc\Router\PriorityList;
 use PHPUnit_Framework_TestCase as TestCase;
 use Zend\Stdlib\ArrayUtils;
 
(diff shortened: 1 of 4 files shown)


Ocramius:
Requires test case

kusmierz:
@Ocramius I've added. Dunno how it write it better.

Ocramius:
@kusmierz the test should be related to `Chain`, not to the stdlib component: you changed code only in the `Mvc` component after all ;-)

kusmierz:
Indeed, you've right. I hope there is no more usages of "key()" function on PriorityList in code (I've checked it quickly).

Is it correct now?

tests/ZendTest/Mvc/Router/PriorityListTest.php
```diff
@@ -11,6 +11,7 @@
 
 use Zend\Mvc\Router\PriorityList;
 use PHPUnit_Framework_TestCase as TestCase;
+use Zend\Stdlib\ArrayUtils;
```
Ocramius: This test should be on [`ChainTest`](https://github.com/zendframework/zf2/blob/1690c8824d0966e564a204bfea1a17d95de6d6d9/tests/ZendTest/Mvc/Router/Http/ChainTest.php), as that's where the code changes are.

tests/ZendTest/Mvc/Router/PriorityListTest.php
```diff
@@ -137,4 +138,15 @@ public function testPriorityWithNegativesAndNull()
 
         $this->assertEquals(array('bar', 'foo', 'baz'), $orders);
     }
+
+    public function testKey()
```
Ocramius: The test should only interarct with a `Chain` instance, not with the internal implementation details of it.

You should simply use `Chain#assemble()` with parameters that show a failure.

kusmierz:
I've finally found time to write it. @Ocramius could you check it if it's ok now? Thanks in advance.

Without my fix, the new test will fail on:

```phpunit
Failed asserting that two strings are equal.
Expected :/foo
Actual   :/foo/bar
 <Click to see difference>

zf2/tests/ZendTest/Mvc/Router/Http/ChainTest.php:180
 phar:///usr/share/php/phpunit.phar/phpunit/TextUI/Command.php:151
 ```

kusmierz:
@Ocramius, have you added this to v2.3.5?

Ocramius:
@kusmierz no, as you can see it's scheduled for 2.4.0

Issue #7155, opened 2015-01-25
AbstractOptions no longer camel casing methods names.


iamtommetcalfe:
When checking if methods exists from array in AbstractOptions.php, ucwords is no longer applied resulting in BadMethodCallException which does actually use ucwords.

e.g. https://github.com/zendframework/zf2/blob/b1e9f3bb3ce9cdc925139d98672af40f6b0f33d7/library/Zend/Stdlib/AbstractOptions.php#L101

This is what it used to be - $setter = 'set' . str_replace(' ', '', ucwords(str_replace('_', ' ', $key)));

jaapio:
might be related to the changes in #7006.
But since the build for master is still passing on this point and methods are case insensitive it is probably an other issue.

Could you provide some more info? Like your OS and PHP version. And if possible your options class and the way you are calling it.

Ocramius:
This is invalid: you most probably have a bug somewhere else: http://3v4l.org/94eT9


```php
<?php

class Foo { function BAR() {}}

var_dump(method_exists(new Foo, 'bar'));
var_dump(method_exists(new Foo, 'BAR'));
```

Please open a different issue once you figured out the exact problem.

iamtommetcalfe:
I am passing an array of options to setFromArray -
e.g 
[
"name" => "Price Range",
"range" => "0.00-10.00"
];

Because __set no longer uses ucwords, it is looking for "getname" rather than "getName". As method_exists is case sensitive this now throws an exception.

iamtommetcalfe:
Apologies for this.
Turns out the array I was passing through had a space in one of the array keys.

It would have been helpful if the exception uses the $setter variable defined previously instead of displaying a differently formatted setter. Because the exception indicated that it was looking for "getName" I went down a different route of debugging. 

Ideally it should be

throw new Exception\BadMethodCallException(sprintf(
                'The option "%s" does not have a matching "%s" setter method which must be defined',
                $key,
                $setter
            ));

Martin-P:
It is indeed strange the exception uses a completely different setter which is not even used in the check in [`Zend\Stdlib\AbstractOptions` line 103](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/library/Zend/Stdlib/AbstractOptions.php#L103).
```php
$setter = 'set' . str_replace('_', '', $key);
```
versus
```php
'set' . str_replace(' ', '', ucwords(str_replace('_', ' ', $key)))
```

[...]

PR #7157, opened 2015-01-26
Exception message in Zend\Stdlib\AbstractOptions should use the actual used setter

Martin-P:
The current exception message uses some fabricated setter which is not even used in the actual test in [`Zend\Stdlib\AbstractOptions` line 103](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/library/Zend/Stdlib/AbstractOptions.php#L103).

This PR changes the exception message and makes it use the actual used setter and thus creates a more accurate message. See: https://github.com/zendframework/zf2/issues/7155#issuecomment-71451833

### DIFF
--- a/library/Zend/Stdlib/AbstractOptions.php
+++ b/library/Zend/Stdlib/AbstractOptions.php
@@ -108,9 +108,10 @@ abstract class AbstractOptions implements ParameterObjectInterface
 
         if ($this->__strictMode__) {
             throw new Exception\BadMethodCallException(sprintf(
-                'The option "%s" does not have a matching "%s" setter method which must be defined',
+                'The option "%s" does not have a matching "%s" ("%s") setter method which must be defined',
                 $key,
-                'set' . str_replace(' ', '', ucwords(str_replace('_', ' ', $key)))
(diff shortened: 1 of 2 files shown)


Martin-P:
Was thinking about changing the value of `$setter` itself, but it seems the usage of an actual different `$setter` value was a minor performance tweak added in https://github.com/zendframework/zf2/commit/f07104142881de54c1131ebb5fabca781f532563.

Ocramius:
I know I'm pedantic here, but can haz test?

Martin-P:
I added a test for the exception message, but I may hope nobody uses the exception message itself as a way to test which exception is thrown.

Ocramius:
>  may hope nobody uses the exception message itself as a way to test which exception is thrown

Yeah, it's not about that, it's mainly about the usual "cast to string" and such errors ;-)

Ocramius:
@Martin-P merged, thanks!

Had to manually apply changes to the `develop` rebase due to conflicts, no big deal tho.

`master`: zendframework/zf2@49f3986a5e450f758dc5d337f8e08904c8c3613b
`develop`: zendframework/zf2@5d6570fda809c68c829e49ca56b599f6972e81bf

PR #7158, opened 2015-01-26
Add Fix for CollectionsFilter Attachment

Mezzle:
Collections don't seem to have CollectionFilters automatically
added to them.  This change makes it so we automatically connect
the CollectionFilter to them, and attach the defaults for it's
child fieldsets


Ocramius:
Requires tests for the newly introduced paths.

Mezzle:
And fixes for old paths.

weierophinney:
@Mezzle as noted, we need unit tests to accept this.

Issue #7147, opened 2015-01-22
AbstractUnicode filters use $this->options['encoding'] instead of getEncoding()


27cm:
Filters extending AbstractUnicode class use $this->options['encoding'] instead of getEncoding() method, e. g. https://github.com/zendframework/zf2/blob/master/library/Zend/Filter/StringToLower.php#L56

Martin-P:
Fixed in #7159.

PR #7159, opened 2015-01-26
Use getEncoding() instead of accessing options array directly

Martin-P:
Fix for issue #7147. Both `Zend\Filter\StringToLower` and `Zend\Filter\StringToUpper` extend `Zend\Filter\AbstractUnicode`.

No new tests, but that is because I don't see a way to test this change. If anyone got suggestions on how to test it, I will add the tests.

### DIFF
--- a/library/Zend/Filter/StringToLower.php
+++ b/library/Zend/Filter/StringToLower.php
@@ -53,7 +53,7 @@ class StringToLower extends AbstractUnicode
         }
         $value = (string) $value;
 
-        if ($this->options['encoding'] !== null) {
+        if (null !== $this->getEncoding()) {
             return mb_strtolower($value, $this->options['encoding']);
         }
 
--- a/library/Zend/Filter/StringToUpper.php
+++ b/library/Zend/Filter/StringToUpper.php
@@ -53,7 +53,7 @@ class StringToUpper extends AbstractUnicode
         }
         $value = (string) $value;
 
-        if ($this->options['encoding'] !== null) {
+        if (null !== $this->getEncoding()) {
             return mb_strtoupper($value, $this->options['encoding']);
         }
 
--- a/tests/ZendTest/Filter/StringToLowerTest.php
+++ b/tests/ZendTest/Filter/StringToLowerTest.php
@@ -183,7 +183,7 @@ class StringToLowerTest extends \PHPUnit_Framework_TestCase
      */
     public function testFilterUsesGetEncodingMethod()
     {
-        $filterMock = $this->getMock('\Zend\Filter\StringToLower', array('getEncoding'));
+        $filterMock = $this->getMock('Zend\Filter\StringToLower', array('getEncoding'));
         $filterMock->expects($this->once())
                    ->method('getEncoding')
                    ->with();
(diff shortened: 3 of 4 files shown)


library/Zend/Filter/StringToLower.php
```diff
@@ -53,7 +53,7 @@ public function filter($value)
         }
         $value = (string) $value;
 
-        if ($this->options['encoding'] !== null) {
+        if (null !== $this->getEncoding()) {
```
samsonasik: no yoda condition for consistencies

library/Zend/Filter/StringToUpper.php
```diff
@@ -53,7 +53,7 @@ public function filter($value)
         }
         $value = (string) $value;
 
-        if ($this->options['encoding'] !== null) {
+        if (null !== $this->getEncoding()) {
```
samsonasik: no yoda condition for consistencies

library/Zend/Filter/StringToLower.php
Martin-P: There is not really a consistency when I do a search in my IDE:
- Found 403 matches of **!== null** in 205 files
- Found 265 matches of **null !==** in 129 files

There are even PR's where conditions are being changed to yoda conditions. To mention a few: #3520, #3521, #6750.

samsonasik: the #3521 and #3520 , the yoda condition is not accepted. anyway, the current purpose of this pr is changing <code>$this->options['encoding']</code> to <code>$this->getEncoding()</code>, so changing right to left is a cosmetic imo

Martin-P: The yoda condition is cosmetic indeed, but so is PSR2 :wink: Personally I prefer yoda conditions, because I think defensive programming is good practice.

library/Zend/Filter/StringToUpper.php
Ocramius: @samsonasik tbh, I like the slow migration to yodaism :P

Ocramius:
@Martin-P this requires testing, IMO. Mocking the `getEncoding` method is also "meh", but required for the test to work :-\

Martin-P:
How can I mock the method of the same class? This does not work:
```php
public function testFilterUsesGetEncoding()
{
    $filterMock = $this->getMock('StringToLowerFilter');
    $filterMock->expects($this->once())
               ->method('getEncoding')
               ->with();
    $filterMock->filter('foo');
}
```

>  Fatal error: Call to undefined method Mock_StringToLowerFilter_0ae23a04::filter()

Adding the `filter()` method to the mocked object does not help either, because then `getEncoding()` is never called.

Ocramius:
@Martin-P add `$this->getMock('StringToLowerFilter', array('getEncoding'))`. The class of the mock must be the FQCN.

tests/ZendTest/Filter/StringToLowerTest.php
```diff
@@ -177,4 +177,16 @@ public function testReturnUnfiltered($input)
     {
         $this->assertEquals($input, $this->_filter->filter($input));
     }
+
+    /**
+     * @group 7147
+     */
+    public function testFilterUsesGetEncodingMethod()
+    {
+        $filterMock = $this->getMock('\Zend\Filter\StringToLower', array('getEncoding'));
```
Ocramius: Note that in strings you always deal with an absolute namespace, therefore the leading `\` is not needed and should be avoided.

Martin-P: Ah, okay, I have removed them.

weierophinney:

PR #7164, opened 2015-01-27
.php_cs - add few rules

keradus:
Add few rules to PHP CS Fixer configuration that are already followed.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -45,6 +45,7 @@ $config->fixers(
         'parenthesis',
         'multiple_use',
         'method_argument_space',
+        'object_operator',
         'php_closing_tag',
         'psr0',
         'remove_lines_between_uses',
--- a/.php_cs
+++ b/.php_cs
@@ -38,6 +38,7 @@ $config->fixers(
         'function_call_space',
         'function_declaration',
         'indentation',
+        'join_function',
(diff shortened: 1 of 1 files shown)


.php_cs
```diff
@@ -37,12 +38,14 @@ $config->fixers(
         'function_call_space',
         'function_declaration',
         'indentation',
+        'join_function',
```
Ocramius: Maybe we should bump the `composer.json` version of the fixer, to avoid regressions?

gianarb:
@Ocramius with other CR team developers have you thinking a set of common roles for cs-fix? (thoughts after split :P )

.php_cs
keradus: To be honest I don't get what's on your mind.
In composer there is latest stable version of PHP CS Fixer.
And what regression?

keradus:
@gianarb that would be great. But currently we do not know which rules are followed in actual codebase - they are followed in most cases, but to turn then on they usually need little fixing to made Travis happy about them.

gianarb:
Np for this PR.. It's a general question :+1: 
Sorry for this intrusion :smile:

Ocramius:
@keradus thanks, merged after some further thoughts about it

PR #7170, opened 2015-01-30
Fixed wrong phpdoc type hint.

hver:
`$pages` variable can only has the type `stdClass`, because it is created with the function `_createPages()` (line 890), and this function returns `stdClass`

### DIFF
--- a/library/Zend/Paginator/Paginator.php
+++ b/library/Zend/Paginator/Paginator.php
@@ -140,7 +140,7 @@ class Paginator implements Countable, IteratorAggregate
     /**
      * Pages
      *
-     * @var array
+     * @var \stdClass
      */
     protected $pages = null;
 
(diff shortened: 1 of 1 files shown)


Ocramius:
A little bit of me just died by reading this :-\ Sad `stdClass` is sad.

hver:
As git blame shows, this is a rather old class :calendar: 
 There are some more type hint errors in this class, but they are not so trivial:
on line 372: `$cacheIterator = static::$cache->getIterator();`
`self::$cache` is hinted as `Zend\Cache\Storage\StorageInterface` but `StorageInterface` doesn't has the method `getIterator()`, only `Zend\Cache\Storage\IterableInterface` has it. I searched through adapters and only some of them implements `IterableInterface`. So basically `Paginator` class is using `$cache` as an instance of concrete `Adapter`, not as a `StorageInterface` implementator.

Ocramius:
@hver consider opening an issue about that. I'm merging this PR meanwhile.

Ocramius:
@hver merged, thanks!

`master`: zendframework/zf2@6bf112d64225cd2429234d19e88909be1b724c4d
`develop`: zendframework/zf2@71f781356a5f51c1182441af9d78afb62950ceb2

PR #7171, opened 2015-01-31
Do not render password value

### LINKED ISSUE #7166: Factory does not create Password Element if only Password Attribute is present
Issue #2602 added desirable security functionality to the Password Element including:

 - If an Edit form is built from a database object, the password will automatically be excluded from the data sent to the browser

It was later noted that the password attribute (frequently used to identify a password field) is not considered by the Factory:

    array(
        'name' => 'something',
        'attributes' => array(
            'type' => 'password',
        ),
        // ...
    )

This results in a `\Zend\Form\Element` instead of a `\Zend\Form\Element\Password`.  To get a password element, the user must explicitly identify the password type in the parent array:

    array(
        'name' => 'something',
        'type' => 'Password',
        // ...
    )

To be fair, the distinction is real and could theoretically be a desired behavior.  

 - At minimum, I would recommend adding a special warning in the documentation, possibly under [Factory-Backed Form Extension](http://framework.zend.com/manual/current/en/modules/zend.form.quick-start.html#factory-backed-form-extension).
 - However, I also believe this is an obvious spot to "nudge" users by making the most secure assumption (password attribute intends to create a password element).  Obviously, a user could override this setting by explicitly asking for a `type => text`.

Martin-P:
This PR fixes #7166. The original issue #2602 was partially solved in #2613. I used the same approach as in #2613.

### DIFF
--- a/library/Zend/Form/View/Helper/FormInput.php
+++ b/library/Zend/Form/View/Helper/FormInput.php
@@ -120,8 +120,12 @@ class FormInput extends AbstractHelper
 
         $attributes          = $element->getAttributes();
         $attributes['name']  = $name;
-        $attributes['type']  = $this->getType($element);
+        $type                = $this->getType($element);
+        $attributes['type']  = $type;
         $attributes['value'] = $element->getValue();
+        if ('password' == $type) {
+            $attributes['value'] = '';
+        }
 
         return sprintf(
             '<input %s%s',
--- a/tests/ZendTest/Form/View/Helper/FormInputTest.php
+++ b/tests/ZendTest/Form/View/Helper/FormInputTest.php
@@ -499,4 +499,16 @@ class FormInputTest extends CommonTestCase
 
         $this->assertContains('title="translated&#x20;string"', $markup);
     }
+
+    /**
+     * @group 7166
+     */
+    public function testPasswordValueShouldNotBeRendered()
+    {
+        $element = new Element('foo');
+        $element->setAttribute('type', 'password');
+
+        $markup  = $this->helper->__invoke($element);
+        $this->assertContains('value=""', $markup);
+    }
 }
--- a/tests/ZendTest/Form/View/Helper/FormPasswordTest.php
+++ b/tests/ZendTest/Form/View/Helper/FormPasswordTest.php
@@ -126,7 +126,8 @@ class FormPasswordTest extends CommonTestCase
         $markup  = $this->helper->render($element);
         switch ($attribute) {
             case 'value':
-                $expect  = sprintf('%s="%s"', $attribute, $element->getValue());
+                // Password value should never be rendered, see 7166
+                $expect  = sprintf('%s=""', $attribute);
                 break;
             default:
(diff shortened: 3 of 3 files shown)


library/Zend/Form/Element.php
```diff
@@ -83,6 +86,20 @@ public function init()
...
+     * Remove the password before rendering if the form fails in order to avoid any security issue
+     *
+     * @param  FormInterface $form
+     * @return mixed
+     * @see    \Zend\Form\Element\Password::prepareElement
+     */
+    public function prepareElement(FormInterface $form)
+    {
+        if ('password' == $this->getAttribute('type')) {
+            $this->setValue('');
```
Ocramius: This seems wrong to me: you don't know when `prepareElement` is going to be called. Instead, it would be better to filter out the value before rendering

Martin-P: It is always called at `$form->prepare()` and because `$form->prepare()` is needed before rendering a form this will work just fine.

I agree ideally you should alter the view helper, but because an `<input type="password" />` can be an instance of `Zend\Form\Element` *or* `Zend\Form\Element\Password` and thus can be rendered by `Zend\Form\View\Helper\FormInput` *or* `Zend\Form\View\Helper\FormPassword` it needs fixing on two separate places. I am opening a new issue about this, because it affects all form elements.

Martin-P: Please see #7172 for a more thorough explanation.

Ocramius: @Martin-P I know when it is called, but that's still a big mess if anything modifies the form after `prepare()`. We rely on that initializer approach too much, and that makes the component highly unstable.

Martin-P: What would be your suggestion? Changing the view helpers? Can do that, no problem, but placing password specific code in the generic `FormInput` view helper does not seem the right way to do this...

[Line 124 `Zend\Form\View\Helper\FormInput`](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/library/Zend/Form/View/Helper/FormInput.php#L124)
Change this:
```php
$attributes['type']  = $this->getType($element);
$attributes['value'] = $element->getValue();
```
Into this?
```php
$type                = $this->getType($element);
$attributes['type']  = $type;
$attributes['value'] = $element->getValue();
if ('password' == $type) {
    $attributes['value'] = '';
}
```

Martin-P: Ping @Ocramius Im adding a tag `awaiting organisation updates` :wink:

weierophinney: @Martin-P Actually, yes, I think that _would_ make more sense. There are cases where you want to keep the value of the password in the element, even if you do not want to _display_ it. As such, doing this in the view layer makes more sense, as display issues are view issues.

Martin-P:
@weierophinney I updated the view helper so it does not render the password value value anymore and also added tests for both angles:
-  `Zend\Form\View\Helper\FormInput`
- `Zend\Form\View\Helper\FormPassword`

This PR makes PR #2613 obsolete, so theoretically that patch can be removed.

weierophinney:

PR #7175, opened 2015-02-02
Reduce iteration when attempt to authenticate user

ccasciotti:
In case of (many) ambiguous identities we can reduce iteration on _authenticate()_ method by setting ORDER clause to SELECT query

### DIFF
--- a/library/Zend/Authentication/Adapter/DbTable/CredentialTreatmentAdapter.php
+++ b/library/Zend/Authentication/Adapter/DbTable/CredentialTreatmentAdapter.php
@@ -95,6 +95,12 @@ class CredentialTreatmentAdapter extends AbstractAdapter
             ->columns(array('*', $credentialExpression))
             ->where(new SqlOp($this->identityColumn, '=', $this->identity));
 
+        if ($this->getAmbiguityIdentity()) {
+            $dbSelect->order(array(
+                'zend_auth_credential_match' => 'DESC'
+            ));
+        }
+
         return $dbSelect;
     }
 


Ocramius:
@teknoman I'm missing the point of this PR: what exactly is being solved? Can you provide a functional test case?

teknoman:
@Ocramius assume that you have records in the table users with same credentials but with different roles:

id | email | password | role
------------ | ------------- | -------------- | --------------
1 | mymail@domain.tld | mycryptedpwd | guest
2 | mymail@domain.tld | mycryptedpwd | user
3 | mymail@domain.tld | mycryptedpwd | pro

When the user with role *pro* tries to login, the resultant query built in *authenticateCreateSelect()* method in Zend\Authentication\Adapter\DbTable\CredentialTreatmentAdapter.php (row 79) is:

SELECT users.*, (CASE WHEN PASSWORD = 'mycryptedpwd' AND role = 'pro' THEN 1 ELSE 0 END) AS zend_auth_credential_match FROM users WHERE EMAIL = 'mymail@domain.tld'; as shown below:

id | email | password | role | zend_auth_credential_match
------------ | ------------- | -------------- | -------------- | -------------
1 | mymail@domain.tld | mycryptedpwd | guest | 0
2 | mymail@domain.tld | mycryptedpwd | user | 0
3 | mymail@domain.tld | mycryptedpwd | pro | 1

if we add a ORDER clause (ORDER BY zend_auth_credential_match DESC, the code in this PR), in case of matched credentials, the number of iterations in method *authenticate()* (row 245 of AbstractAdapter.php) can be reduced to 1, saving time and memory.

Ocramius:
Why would you ever match against the credential anyway?
The DB should never be queried with the hashed password, as it indicates that your hashing algo is messed up/wrong.

teknoman:
Hi Marco,
Sorry but that was just an example. The query is built through the SqlExpr class and is matching against the credentialTreatment attribute. I haven't touched that code though, I just added an order by clause right below.
As you can see from the "Files changed" report I added the following code:

    if ($this->getAmbiguityIdentity()) {
        $dbSelect->order(array(
            'zend_auth_credential_match' => 'DESC'
        ));
    }

This because the ``authenticateCreateSelect()`` method generates and returns a query with a ``zend_auth_credential_match`` column which is used by the ``authenticateValidateResult()`` method to check the user identity.
I noticed that AbstractAdapter class iterates over all the result set and validates against each record using the ``authenticateValidateResult()`` method. This happens here: https://github.com/zendframework/zf2/blob/master/library/Zend/Authentication/Adapter/DbTable/AbstractAdapter.php#L242

Now, I believe that the ORDER BY I'm adding with this PR will let us save a couple iterations and match on the first iteration (given that a valid identity has been provided).
Does it make sense to you?

PR #7177, opened 2015-02-02
Oracle quoting fix for values with single quotes

marcelto:
Single quotes need to be handled differently

$str = 'foo \'bar';
echo 'select '.$oracleAdapterPlatform->quoteTrustedValue($str).'from dual';//sql string won't run in Oracle(11.2.0.4.0 in this case)

outputs: select 'foo \'bar' from dual

needs to output: select 'foo ''bar' from dual

### DIFF
--- a/library/Zend/Db/Adapter/Platform/Oracle.php
+++ b/library/Zend/Db/Adapter/Platform/Oracle.php
@@ -104,7 +104,7 @@ class Oracle implements PlatformInterface
             'Attempting to quote a value in ' . __CLASS__ . ' without extension/driver support '
                 . 'can introduce security vulnerabilities in a production environment.'
         );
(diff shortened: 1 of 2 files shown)


macnibblet:
Can you add a test case ?

MarcelTO:
Sorry I'm still a little new to pull requests.  Do you mean an assertion in the Oracle unit test?

macnibblet:
You need to add a test case in https://github.com/zendframework/zf2/blob/master/tests/ZendTest/Db/Adapter/Platform/OracleTest.php so we don't have any regressions.

MarcelTO:
Ah yes, I see.

MarcelTO:
Priority issue at work, will fix delete assertions in unit test asap

macnibblet:
:+1:

weierophinney:

PR #7178, opened 2015-02-02
Fixes #7136 : add ext-ldap as required extension in Ldap\composer.json

### LINKED ISSUE #7136: Zend Ldap Composer Dependency
Zend Ldap, composer dependency in composer.json does not enforce requirement for php's LDAP extension, thus causing exception when used on servers without the exception.

To replicate, install on server without php_ldap, then try to instantiate using `new Zend\Ldap\Ldap($options);`

samsonasik:
Fixes #7136

### DIFF
--- a/library/Zend/Ldap/composer.json
+++ b/library/Zend/Ldap/composer.json
@@ -21,7 +21,8 @@
         "zendframework/zend-eventmanager": "self.version"
     },
     "suggest": {
-        "zendframework/zend-eventmanager": "Zend\\EventManager component"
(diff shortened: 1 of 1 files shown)


weierophinney:
@samsonasik Shouldn't this be in the `require` section? Essentially, the component is unusable if the extension is not present...

samsonasik:
@weierophinney ok, done. I've moved ext-ldap to require in composer.json

weierophinney:

PR #7179, opened 2015-02-02
Fixes #6419 : add missing dependencies at Feed\composer.json

### LINKED ISSUE #6419: missing dependency for zendframework/zend-feed stand-alone
When I use _zendframework/zend-feed_ standalone;

```
{
    "zendframework/zend-feed": "~2.3"
}
```

I get the following error;

```
PHP Fatal error:  Class 'Zend\ServiceManager\AbstractPluginManager' not found in...
```

Seems there is missing a dependency for the feed component.

samsonasik:
Fixes #6419

### DIFF
--- a/library/Zend/Feed/composer.json
+++ b/library/Zend/Feed/composer.json
@@ -16,20 +16,18 @@
     "require": {
         "php": ">=5.3.23",
         "zendframework/zend-escaper": "self.version",
-        "zendframework/zend-stdlib": "self.version"
+        "zendframework/zend-stdlib": "self.version",
+        "zendframework/zend-http": "self.version",
+        "zendframework/zend-servicemanager": "self.version"
     },
     "require-dev": {
         "zendframework/zend-db": "self.version",
         "zendframework/zend-cache": "self.version",
-        "zendframework/zend-http": "self.version",
-        "zendframework/zend-servicemanager": "self.version",
         "zendframework/zend-validator": "self.version"
     },
     "suggest": {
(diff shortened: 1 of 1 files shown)


Ocramius:
Ping @padraic

weierophinney:
This is not an appropriate fix.

The dependencies you are adding were explicitly removed to allow the component to be used standalone. The problem people are reporting is a documentation issue: if you use it standalone, you need to create your own implementation of `Zend\Feed\Reader\ExtensionPluginInterface` that will allow resolving the various feed and entry extensions plugins.

Right now, the default implementation of that is a `Zend\ServiceManager\AbstractPluginManager` extension, which is why users are reporting that the dependency is missing. Ideally, we should likely provide a default implementation that does not have dependencies, and only have the current `ExtensionManager` implementation for those who want to leverage the capabilities provided by our ServiceManager component.

gianarb:
@weierophinney in my opinion this problem exist in a lot of components and this solution is little shortcut..
This "optional dependencies" produce confusion, Composer manage three chapters:

* `require` satisfy the requirements of component's feature
* `require-dev` satisfy the tests requirements
* `suggest` is another history :)

My question is why (for example) [ExtensionPluginManager](https://github.com/zendframework/Component_ZendFeed/blob/43ee6d93e806a33fea326bc610bd7f90d472045b/Reader/ExtensionPluginManager.php) is in this component and not in another place?  
This is a component's feature, it should satisfy your requirement

Thanks! :)

PR #7180, opened 2015-02-02
Fix: Inline variable

localheinz:
This PR
- [x] returns the newly created instance of a module class directly, instead of assigning it to a variable first and then returning it


Ocramius:
Merged, thanks @localheinz!

`master`: zendframework/zf2@de98f631ac64b53ccc00bcaf667040fa35962993
`develop`: zendframework/zf2@be0b3496b73a61a255d05e945b75f6fdf0995c31

Issue #7174, opened 2015-02-02
Cache in Paginator module supports only `Memory` and `Filesystem` adapters


hver:
There are problems with cache in Paginator module `Zend/Paginator/Paginator.php`.
The `Paginator` class uses for caching an `Adapter` object, which implements `Zend\Cache\Storage\StorageInterface`. But this class uses methods from `Adapter` which are not declared by `StorageInterface`:
* Lines 372, 710: `Zend\Cache\Storage\IterableInterface::getIterator()`
* Lines 375, 713, `Zend\Cache\Storage\TaggableInterface::getTags()`
* Line 620, `Zend\Cache\Storage\TaggableInterface::setTags()`

This is a problem, because only `Memory` and `Filesystem` cache adapters implement `IterableInterface` and `TaggableInterface`. Other adapters will throw a runtime error i.e. `Apc`, `Dba`, `Memcached`, `Redis`, `Wincache`, `XCache`, `ZendServerDisk`, `ZendServerShm`.

I think, the only possibility is to rewrite the cache handling in `Paginator` module, so it doesn't require additional interfaces. Otherwise, we should explicitly mention that only an instance of `Memory` or `Filesystem` adapter is allowed to be used as cache in Paginator.

Ocramius:
Absolute links to code references:

 - https://github.com/zendframework/zf2/blob/9ee2075758265876b294a14287c135c63ff241f7/library/Zend/Paginator/Paginator.php#L372
 - https://github.com/zendframework/zf2/blob/9ee2075758265876b294a14287c135c63ff241f7/library/Zend/Paginator/Paginator.php#L710
 - https://github.com/zendframework/zf2/blob/9ee2075758265876b294a14287c135c63ff241f7/library/Zend/Paginator/Paginator.php#L375
 - https://github.com/zendframework/zf2/blob/9ee2075758265876b294a14287c135c63ff241f7/library/Zend/Paginator/Paginator.php#L713
 - https://github.com/zendframework/zf2/blob/9ee2075758265876b294a14287c135c63ff241f7/library/Zend/Paginator/Paginator.php#L620

hver:
These additional interfaces are only used by two methods:
* `getPageItemCache()` https://github.com/zendframework/zf2/blob/9ee2075758265876b294a14287c135c63ff241f7/library/Zend/Paginator/Paginator.php#L705
[...]

PR #7181, opened 2015-02-03
Remove hidden TaggableInterface dependencies from paginator

hver:
This is an example for discussion on issue #7174 Cache in Paginator module supports only `Memory` and `Filesystem` adapters

### DIFF
--- a/library/Zend/Paginator/Paginator.php
+++ b/library/Zend/Paginator/Paginator.php
@@ -372,11 +372,8 @@ class Paginator implements Countable, IteratorAggregate
             $cacheIterator = static::$cache->getIterator();
             $cacheIterator->setMode(CacheIterator::CURRENT_AS_KEY);
             foreach ($cacheIterator as $key) {
-                $tags = static::$cache->getTags($key);
-                if ($tags && in_array($this->_getCacheInternalId(), $tags)) {
-                    if (substr($key, 0, $prefixLength) == self::CACHE_TAG_PREFIX) {
-                        static::$cache->removeItem($this->_getCacheId((int)substr($key, $prefixLength)));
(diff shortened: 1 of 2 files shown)


library/Zend/Paginator/Paginator.php
```diff
@@ -372,11 +372,8 @@ public function clearPageItemCache($pageNumber = null)
             $cacheIterator = static::$cache->getIterator();
             $cacheIterator->setMode(CacheIterator::CURRENT_AS_KEY);
             foreach ($cacheIterator as $key) {
-                $tags = static::$cache->getTags($key);
-                if ($tags && in_array($this->_getCacheInternalId(), $tags)) {
-                    if (substr($key, 0, $prefixLength) == self::CACHE_TAG_PREFIX) {
-                        static::$cache->removeItem($this->_getCacheId((int)substr($key, $prefixLength)));
```
hver: It works also with `static::$cache->removeItem($key);`. Is there a reason to do it this way?

hver:
I'm not sure, why the Travis build on PHP 5.4 fails on `ZendTest\Cache\Storage\Adapter\FilesystemTest::testClearExpired`. I have modified only `Paginator` and `PaginatorTest`, these changes should not affect `Filesystem` adapter.

Ocramius:
Those tests are sadly time-bound, and tend to fail when the travis
infrastructure is stressed out.

library/Zend/Paginator/Paginator.php
weierophinney: From reading the code, this ensures that the page number is properly injected into the key.

weierophinney:

PR #7182, opened 2015-02-04
update copyright year to 2015 that still uses 2013 in "develop"

samsonasik:


samsonasik:
@weierophinney this should can land in 2.4.0 ;)

PR #7183, opened 2015-02-04
update copyright year to 2015 that still uses 2013 in "master"

samsonasik:


weierophinney:

PR #7184, opened 2015-02-05
FormButton $buttonContent parameter is now translated

Victory:
In master if you pass in a `$buttonContent` parameter to `Zend\Form\View\Helper\FormButton` the `$buttonContent` is not translated. 

 `Zend\Form\View\Helper\FormButton->_invoke($element, "translate me")`

This pull requests creates a test to show the error and a fix to this issue.

### DIFF
--- a/library/Zend/Form/View/Helper/FormButton.php
+++ b/library/Zend/Form/View/Helper/FormButton.php
@@ -87,20 +87,14 @@ class FormButton extends FormInput
                     )
                 );
             }
-
-            if (null !== ($translator = $this->getTranslator())) {
-                $buttonContent = $translator->translate(
-                    $buttonContent,
-                    $this->getTranslatorTextDomain()
-                );
-            }
-        } else {
-            if (null !== ($translator = $this->getTranslator())) {
-                $buttonContent = $translator->translate(
-                    $buttonContent, $this->getTranslatorTextDomain()
-                );
-            }
         }
+        
+        if (null !== ($translator = $this->getTranslator())) {
+            $buttonContent = $translator->translate(
+                $buttonContent, $this->getTranslatorTextDomain()
+            );
+        }
+        
 
         if (! $element instanceof LabelAwareInterface || ! $element->getLabelOption('disable_html_escape')) {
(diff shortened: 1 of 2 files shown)


library/Zend/Form/View/Helper/FormButton.php
```diff
@@ -91,6 +91,12 @@ public function render(ElementInterface $element, $buttonContent = null)
                     $buttonContent, $this->getTranslatorTextDomain()
                 );
             }
+        } else {
```
Martin-P: No need for the extra `else`. You can simply move the first `if`:
```php
/**
 * Render a form <button> element from the provided $element,
 * using content from $buttonContent or the element's "label" attribute
 *
 * @param  ElementInterface $element
 * @param  null|string $buttonContent
 * @throws Exception\DomainException
 * @return string
 */
public function render(ElementInterface $element, $buttonContent = null)
{
    $openTag = $this->openTag($element);

    if (null === $buttonContent) {
        $buttonContent = $element->getLabel();
        if (null === $buttonContent) {
            throw new Exception\DomainException(sprintf(
                '%s expects either button content as the second argument, ' .
                    'or that the element provided has a label value; neither found',
                __METHOD__
            ));
        }
    }

    if (null !== ($translator = $this->getTranslator())) {
        $buttonContent = $translator->translate(
            $buttonContent, $this->getTranslatorTextDomain()
        );
    }

    if (! $element instanceof LabelAwareInterface || ! $element->getLabelOption('disable_html_escape')) {
        $escapeHtmlHelper = $this->getEscapeHtmlHelper();
        $buttonContent = $escapeHtmlHelper($buttonContent);
    }

    return $openTag . $buttonContent . $this->closeTag();
}
```

Victory: Good suggestion, i have added the changes and pushed.

Victory:
There seems to be some unrelated failing tests in "ZendTest\Cache\Storage\Adapter\DbaInifileTest" on travis. The test in this pull request passes.

weierophinney:

Victory:
thanks as always @weierophinney !

PR #7194, opened 2015-02-08
fix some of the documentation errors

jaapio:


library/Zend/Code/Reflection/ClassReflection.php
```diff
@@ -230,11 +230,17 @@ public function getProperties($filter = -1)
...
+     */
     public function toString()
     {
         return parent::__toString();
     }
 
+    /**
+     * @return string
+     */
     public function __toString()
```
Martin-P: If this method only calls `parent::__toString()`, it has no use and can be removed.

library/Zend/Code/Reflection/MethodReflection.php
```diff
@@ -460,11 +461,17 @@ protected function isValidFunction($haystack, $position, $functionName = null)
...
+     */
     public function toString()
     {
         return parent::__toString();
     }
 
+    /**
+     * @return string
+     */
     public function __toString()
```
Martin-P: If this method only calls `parent::__toString()`, it has no use and can be removed.

library/Zend/Code/Reflection/ParameterReflection.php
```diff
@@ -98,11 +98,17 @@ public function getType()
...
+     */
     public function toString()
     {
         return parent::__toString();
     }
 
+    /**
+     * @return string
+     */
     public function __toString()
```
Martin-P: If this method only calls `parent::__toString()`, it has no use and can be removed.

jaapio:
@Martin-P tnx for your comments. I 'm working in an other branch to clean this up. There are several methods in the zf-code component that where not covered by tests and methods like these tostring that are useless. I didn't want to mix those changes, so this one only contains docblock changes.

jaapio:
The to string methods are possibly added for the same reason why this was done in the [FunctionReflection class](https://github.com/zendframework/zf2/blob/master/library/Zend/Code/Reflection/FunctionReflection.php#L258) There are no tests for this method and I could reproduce it by adding a test?

Martin-P:
> Required due to bug in php

@ralphschindler You wrote [this comment](https://github.com/zendframework/zf2/commit/17e89d2c481e1f49ead94d543ac316c4bf55f44f#diff-6a0bcab8b73d78a50697255be67ba297R133). Can you elaborate on this and provide a link to the bug you were referring to?

Issue #6285, opened 2014-05-15
[ZF3] [RFC] Removal of `Zend\Loader`


Ocramius:
## Composer as primary autoloader

Just a suggestion, but I think that [composer](http://getcomposer.org/) pretty much took over the autoloading-related problems in all of PHP.

This is a good time to dismiss `Zend\Loader`.

**EDIT**: the `git submodule update --init` approach could also finally die :-)

## Compatibility

Almost every module that is on github currently provides a `composer.json` file. The BC break would be only for most in-house modules that still rely on `Module#getAutoloaderConfig()`.

We can get rid of `getAutoloaderConfig()` or provide a small compat layer that just merges the fetched configuration into the already existing [`Composer\Autoload\ClassLoader`](https://github.com/composer/composer/blob/0c343f925ad578a9da4d9a4cfff4e91b500d206c/src/Composer/Autoload/ClassLoader.php).

## Advantages

One less component to maintain, nuff said.

## Problems

The only issue with composer-first autoloading is security, but I think that can be dealt with until this issue becomes a PR.

Additionally, composer can be used without packagist and as a pure autoloader (`composer.phar dump-autoload`).

internalsystemerror:
In case its relevant, I only ended up learning composer because I wanted to try out ZF2 and I'm forever greatful for it. First thing I ended up doing was abandoning the use of the zf2 autoloader. Since I'm using packages that require composer now I'd rather use 1 autoloader than 2.

One thing that struck my attention was the mention of enterprise systems that move a lot slower. I happen to be working for one of said enterprises where over the past few months I've been introducing the idea of composer and have started to use packages from it. Things do move very slowly indeed although the problems have been reluctance more than anything technical.

In short, those enterprise companies move so slow that by the time they end up wanting to use ZF3, will be when you guys are thinking about ZF4, so maybe isnt as much of a concern as you might think.

Just my two cents, hope it helps.

freemany:
why reinvent the wheel, composer is good<img class="emoji" title=":+1:" alt=":+1:" src="https://assets-cdn.github.com/images/icons/emoji/unicode/1f44d.png" height="20" width="20" align="absmiddle">

GeeH:
This issue has been closed as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

PR #7195, opened 2015-02-08
Composer autoloader

nikolaposa:
## Problem

Most of the ZF2-based applications relies on Composer autoloading mechanism for loading ZF2 classes itself, as well as classes of all the other vendor dependencies. On the other hand, ZF2 internally uses its own autoloading implementation, which ultimately registers appropriate autoloader strategies (`StandardAutoloader`, `ClassMapAutoloader`, etc.) with the `spl_autoload_register()`.

Composer pretty much handles all the autoloading-related aspects, thus it is kinda unnecessary to have all those custom autoloading solutions registered within the SPL __autoload queue.
## Solution

Autoloader implementation - `Zend\Loader\ComposerAutoloader`, which adapts Composer's autoloader so that it can be used for class loading in the application.
### Usage

Autoloading initialization (somewhere in the `index.php`)

``` php
$loader = require 'vendor/autoload.php';
Zend\Loader\AutoloaderFactory::factory(array(
    //Composer class loader instance is a mandatory dependency
    'Zend\Loader\ComposerAutoloader' => $loader 
));
```

Sample module:

``` php
namespace Application;

use Zend\Mvc\ModuleRouteListener;
use Zend\Mvc\MvcEvent;

class Module
{
    public function getConfig()
    {
        return include __DIR__ . '/config/module.config.php';
    }

    public function getAutoloaderConfig()
    {
        return array(
            'Zend\Loader\ComposerAutoloader' => array(
                'classmap' => include __DIR__ . '/autoload_classmap.php',
                'namespaces' => array(
                    __NAMESPACE__ => __DIR__ . '/src',
                ),
                //PSR-4 namespaces/paths are also supported
                'psr4' => array(
                    __NAMESPACE__ => __DIR__ . '/src',
                ),
            ),
        );
    }
}
```

### DIFF
--- a/library/Zend/Loader/ComposerAutoloader.php
+++ b/library/Zend/Loader/ComposerAutoloader.php
@@ -1,0 +1,132 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Loader;
+
+use Composer\Autoload\ClassLoader;
+
+// Grab SplAutoloader interface
+require_once __DIR__ . '/SplAutoloader.php';
+
+/**
+ * Adapts Composer's autoloader so that it is used for class loading
+ * in the application.
+ *
+ * @author Nikola Posa <posa.nikola@gmail.com>
+ */
+class ComposerAutoloader implements SplAutoloader
+{
+    const LOAD_NS         = 'namespaces';
+    const LOAD_PSR4       = 'psr4';
+    const LOAD_CLASS_MAP  = 'classmap';
+
+    /**
+     * @var ClassLoader
+     */
+    protected $composerAutoloader;
+
+    /**
+     * Constructor
+     *
+     * @param  null|array|ClassLoader $options
+     */
+    public function __construct($options = null)
+    {
+        if (null !== $options) {
+            if (is_array($options) && isset($options['composer_autoloader'])) {
+                $options = $options['composer_autoloader'];
+            }
+
+            if ($options instanceof ClassLoader) {
+                $this->composerAutoloader = $options;
+                return;
+            }
+        }
+
+        throw new Exception\RuntimeException('Composer autloader instance must be supplied');
+    }
+
+    /**
+     * Configure autoloader
+     *
+     * Allows specifying both "namespace" and "prefix" pairs, using the
+     * following structure:
+     * <code>
+     * array(
+     *     'namespaces' => array(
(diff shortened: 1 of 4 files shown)


Ocramius:
I don't see much of a use-case for `Zend\Loader` in any case.

I'd rather suggest just coding all autoloading mappings in `composer.json`

library/Zend/Loader/ComposerAutoloader.php
```diff
@@ -0,0 +1,132 @@
...
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Loader;
+
+use Composer\Autoload\ClassLoader;
```
Ocramius: This couples `Zend\Loader` with the composer autoloader: it's much easier to just use the composer loader directly instead (and avoid coupling just for adapter purposes)

nikolaposa:
@Ocramius I generally agree that `Zend\Loader` is becoming redundant. I saw your ZF3 RFC on the topic of [Zend\Loader removal](https://github.com/zendframework/zf2/issues/6285) and I sincerely hope that your proposal will be accepted, because we share the same opinion.

I thought that it might be beneficial for ZF2 to have this ability, which is the reason why I created this PR. 

As for coupling with Composer's class loader, maybe it can be resolved using different approach. Maybe `ClassLoader` instance can somehow be resolved from the `__autoload` queue, through the `spl_autoload_functions()` call. That would solve that constructor dependency of `ComposerAutoloader` class and there would be no need for that snippet in `index.php`. But yes, these two classes would still be coupled, as `ComposerAutoloader` still has to work with `ClassLoader` interface.

Ocramius:
@nikolaposa what I miss here is the actual use-case scenario.

Consider that I already use `composer.json` to autoload app-specific modules: what is the advantage here, compared to that?

nikolaposa:
@Ocramius Maybe for systems where app-specific modules are dinamically loaded and when there is a mechanism for enabling/disabling modules, i.e. some CMS. Thus it would not be modular and efficient to list all those modules in `composer.json`, if some of them may not be used.

Also, if there is no indication that `Zend\Loader` will be removed from ZF2, that component should probably have these "composer-aware" capabilities.

But ok, I understand your point, your arguments are valid. I guess that this PR should then be closed...

Ocramius:
> modules are dinamically loaded

Usually a bad idea

 > i.e. some CMS

Modules' functionality be disabled, not the module itself

 > there is no indication that `Zend\Loader` will be removed from ZF2

When split happens, `Zend\Loader` will be less than useless

nikolaposa:
Ok, probably no one else will join us in this discussion, so as for me, this PR can be closed.

Ocramius:
@nikolaposa thanks for following up on this!

PR #7203, opened 2015-02-10
Cache: fix mongo options

marc-mabe:
This PR addresses the configuration of the mongo db storage adapter in the following form:
- removed option `lib_options` -> this was copied from moncached but the mongo db hasn't such an option
- moved `server`, `database`, `collection` into own configuration options (before part of `lib_options`)
- removed default server so the default will be auto detected by ext/mongo using INI options
- added own configuration options for `connection_options` and `driver_options`
- removed default `connection_options` so the default will be defined by ext/mongo.

As of the mongo db cache storage adapter is new in 2.4 this will be no BC break until 2.4 is out!!

ping @Ocramius Hopefully soon enough for 2.4
ping @alexdenvir please review as author of this adapter

### DIFF
--- a/tests/ZendTest/Cache/Storage/Adapter/MongoDbResourceManagerTest.php
+++ b/tests/ZendTest/Cache/Storage/Adapter/MongoDbResourceManagerTest.php
@@ -46,63 +46,19 @@ class MongoDbResourceManagerTest extends \PHPUnit_Framework_TestCase
 
         $this->object->setResource($id, $resource);
 
-        $expected = array($id => array('initialized' => true, 'resource' => $resource));
-
-        $this->assertAttributeSame($expected, 'resources', $this->object);
+        $this->assertSame($resource, $this->object->getResource($id));
     }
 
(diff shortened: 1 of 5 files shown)


alexdenvir:
:+1:
Looks good to me.

tests/ZendTest/Cache/Storage/Adapter/MongoDbResourceManagerTest.php
```diff
@@ -110,14 +66,8 @@ public function testSetResourceThrowsException()
         $id = 'foo';
```
weierophinney: This test fails currently, indicating that a `Zend\Cache\Exception\InvalidArgumentException` was not thrown. When I look at `MongoDbResourceManager::setResource()` there is no validation of the `$resource` argument. The typehint specifies array or `MongoCollection`, but does not test for anything other than `MongoCollection` (in which case it branches logic). I'll update it on merge to raise the exception if `$resource` is neither of the two types.

PR #7204, opened 2015-02-11
Error in FirePhp log writer

### LINKED ISSUE #7202: Possible bug in Zend\Log\Writer\FirePhp::__construct
It seems to me that [this](https://github.com/zendframework/zf2/blob/master/library/Zend/Log/Writer/FirePhp.php#L45) condition should be reversed?

chadicus:
The use statement @ line 17 is trying to load a class with a relative namespace path. This is not valid in PHP.  if you use the full namespace the unit tests will fail. 

 I have removed the `use` statement and just used the namespace `FirePhp\FirePhpInterface` in the `instanceof` statement and reversed the condition.

Fixes #7202

### DIFF
--- a/library/Zend/Log/Writer/FirePhp.php
+++ b/library/Zend/Log/Writer/FirePhp.php
@@ -14,7 +14,6 @@ use FirePHP as FirePHPService;
 use Zend\Log\Exception;
 use Zend\Log\Formatter\FirePhp as FirePhpFormatter;
 use Zend\Log\Logger;
-use FirePhp\FirePhpInterface;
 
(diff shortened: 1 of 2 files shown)


Martin-P:
> Contains temp commit which demonstrates the error.

I think it is better to add a unittest.

chadicus:
@Martin-P the first commit is now a new unit test

Martin-P:
Nice :+1: BTW, shouldn't this PR be merged against `master`? It is not a new feature, it is a bug fix.

chadicus:
Closing this based on @Martin-P recommendation. New PR is #7213

Issue #6381, opened 2014-06-16
Zend\Cache\Service\StorageCacheFactory: Order of Redis adapter options may cause Exception


eiwaen:
Depending on the order of the adapter options, the following Exception may be thrown:
`Zend\Cache\Exception\InvalidArgumentException' with message 'Missing required server host'`

This **works**:
```php
'caches' => array(
    'redis' => array(
        'adapter' => array(
            'name' => 'Zend\Cache\Storage\Adapter\Redis',
            'options' => array(
                'resource_id' => 'myresource',
                'server' => 'hostname',
                'password' => 'password',
            ),
        ),
    ),
),
```

This **doesn't work**:
```php
'caches' => array(
    'redis' => array(
        'adapter' => array(
            'name' => 'Zend\Cache\Storage\Adapter\Redis',
            'options' => array(
                'server' => 'hostname', // Defined under resource "default"
                'resource_id' => 'myresource',
                'password' => 'password', // Would be defined under resource "myresource", but exception is thrown (missing hostname when setting password)
            ),
        ),
    ),
),
```

lucian303:
This is obviously a bug. What are we supposed to do guess what order an unordered array of options should go in? At the very least, add a big note to the documentation so that people know the order matters and don't waste hours trying to figure out what's wrong.

marc-mabe:
@eiwaen @lucian303 Should be "fixed" by https://github.com/zendframework/zend-cache/issues/91
Please test it

marc-mabe:
@Ocramius @weierophinney This one is fixed by zendframework/zend-cache#91 and can be closed

PR #7205, opened 2015-02-11
Prototype fixing #6381: Order of resource based adapter options may cause Exception

marc-mabe:
This is a prototype!!!!

This would be duplicated in all Options classes requiring some ordering of options.

@Ocramius @weierophinney In my opinion the issue #6381 should be marked as won't fixed as the options will be applied in the order they are defined. This behavior should be documented but not changed into a adapter defined ordering like this prototype. Additionally this prototype shows it's more error prone than the current behavior and would be needed to be document3ed for each adapter what the defined order would resulting in other WTF moments.

### DIFF
--- a/library/Zend/Cache/Storage/Adapter/MemcacheOptions.php
+++ b/library/Zend/Cache/Storage/Adapter/MemcacheOptions.php
@@ -10,6 +10,7 @@
 namespace Zend\Cache\Storage\Adapter;
 
 use Zend\Cache\Exception;
+use Zend\Stdlib\AbstractOptions;
 
 /**
  * These are options specific to the Memcache adapter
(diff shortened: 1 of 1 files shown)


library/Zend/Cache/Storage/Adapter/MemcacheOptions.php
```diff
@@ -44,6 +45,58 @@ class MemcacheOptions extends AdapterOptions
...
+                )
+            );
+        }
+
+        if ($options instanceof Traversable) {
+            $options = iterator_to_array($options);
+        }
+
+        $options  = array_change_key_case($options, CASE_LOWER);
+        $prioOpts = array();
```
weierophinney: What does `$prioOpts` stand for? `prior options`? or `priority options`? or...? Please be explicit. :)

GeeH:
New issue can be found at: https://github.com/zendframework/zend-cache/issues/106

PR #7208, opened 2015-02-12
Fix: Email Validator vs IDN

svyatoslav-kubakh:
Non-latin emails validation

### DIFF
--- a/library/Zend/Validator/EmailAddress.php
+++ b/library/Zend/Validator/EmailAddress.php
@@ -334,7 +334,7 @@ class EmailAddress extends AbstractValidator
         // atext: ALPHA / DIGIT / and "!", "#", "$", "%", "&", "'", "*",
         //        "+", "-", "/", "=", "?", "^", "_", "`", "{", "|", "}", "~"
         $atext = 'a-zA-Z0-9\x21\x23\x24\x25\x26\x27\x2a\x2b\x2d\x2f\x3d\x3f\x5e\x5f\x60\x7b\x7c\x7d\x7e';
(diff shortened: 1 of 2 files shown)


moderndeveloperllc:
I don't think you can use `idn_to_ascii()` in this class as Zend\Validator does not require the `intl` extension to be present. If you want to validate non-Latin emails, it might be better to write an Zend\I18n validator that uses that function. Thoughts @DASPRiD?

weierophinney:
Two potential options:

- Have it use `idn_to_ascii`, but only if ext/intl is installed. This could be achieved via a method call:
```php
$this->getLocalizedLocalPart();

private function getLocalizedLocalPart()
{
    if (extension_loaded('intl')) {
        return idn_to_ascii($this->localPart);
    }
    return $this->localPart;
}
```
- **or** move it into the `Zend\I18n\Validator` namespace.

Your pick. If you can do this by 9 March 2015, I can include it in 2.4.

moderndeveloperllc:
@svyatoslav-kubakh I can do the change and move if you don't want to deal with it.

svyatoslav-kubakh:
Hello, everyone. I apologize for my absence. Thank you for your help and advice. I will make a second attempt to fix within 1-2 days. If the patch is rejected again, you give in turn. :-)

moderndeveloperllc:
@svyatoslav-kubakh We're your fixes made on another PR, or have you decided not to fix this?

svyatoslav-kubakh:
That's correct. https://github.com/zendframework/zf2/pull/7281

PR #7215, opened 2015-02-13
replace is_a() with instanceof

samsonasik:


sasezaki:
is class autoloading of is_a() not need ??

samsonasik:
no, imo

weierophinney:

PR #7224, opened 2015-02-17
Fix: Argument is missing from doc-block

localheinz:
This PR
- [x] adds missing argument to doc-block of `View\Helper\FlashMessenger::renderMessages()`

Spotted while looking at https://github.com/zendframework/modules.zendframework.com/pull/384#discussion_r24778520


weierophinney:
Already fixed in develop, which is where I was going to merge this.

Issue #7219, opened 2015-02-14
ZendTest\I18n\Validator\FloatTest::testValidationFailures fails with icu 54


remicollet:
Since the update from icu 53 to 54, this test fails.

    There was 1 failure:
    1) ZendTest\I18n\Validator\FloatTest::testValidationFailures with data set #14 ('2 000', false, 'fr-CH')
    Failed expecting 2 000 being false (locale:fr-CH)
    Failed asserting that true matches expected false.

Detected in Fedora QA, see 
http://koschei.cloud.fedoraproject.org/package/php-ZendFramework2

moderndeveloperllc:
Yep. Looks like CLDR 26 (part of ICU 54) [made that change](http://unicode.org/cldr/trac/ticket/7238). I just so happened to pick a wrong example for a validation failure! I'll put in a PR this week to remove it. Functionality still works, just a test break.

PR #7225, opened 2015-02-17
Fix for #7219 - Validation failure fr-CH in testValidationFailures

moderndeveloperllc:
Due to [change in CLDR 26](http://unicode.org/cldr/trac/ticket/7238). Just removed that one instance that is actually valid now.


weierophinney:

PR #7226, opened 2015-02-18
remove deprecated --dev in composer install

samsonasik:
Dev packages are installed by default now


weierophinney:
Change already exists on develop, where I was going to merge this.

PR #7227, opened 2015-02-18
remove required dependencies

ThaDafinser:
Since only the https://github.com/zendframework/zf2/blob/master/library/Zend/Code/Annotation part uses the EventManager, please let us remove the dependency.

If the EventManager gets included, also the complete StdLib comes with the composer resolving....

I just build something similar to: https://github.com/EvanDotPro/EdpSuperluminal/blob/master/Module.php#L5-L7

And from Zend only those classes were actually autoloaded:

```
* AUTOLOAD Zend\Code\Reflection\ClassReflection
* AUTOLOAD Zend\Code\Reflection\ReflectionInterface
* AUTOLOAD Zend\Code\Scanner\FileScanner
* AUTOLOAD Zend\Code\Scanner\TokenArrayScanner
* AUTOLOAD Zend\Code\Scanner\ScannerInterface
* AUTOLOAD Zend\Code\Reflection\FileReflection
* AUTOLOAD Zend\Code\Scanner\CachingFileScanner
```

### DIFF
--- a/library/Zend/Code/composer.json
+++ b/library/Zend/Code/composer.json
@@ -14,12 +14,12 @@
     },
     "target-dir": "Zend/Code",
     "require": {
-        "php": ">=5.3.23",
-        "zendframework/zend-eventmanager": "self.version"
+        "php": ">=5.3.23"
     },
     "require-dev": {
         "doctrine/common": ">=2.1",
-        "zendframework/zend-stdlib": "self.version"
+        "zendframework/zend-stdlib": "self.version",
+        "zendframework/zend-eventmanager": "self.version"
     },
     "suggest": {
         "doctrine/common": "Doctrine\\Common >=2.1 for annotation features",


Ocramius:
I wouldn't remove this dependency: if we do that, I'd rather do it after the split, when every component manager has control over the codebase, and we may be able to completely get rid of the EventManager in the code component.

Removing it now is a problem for anyone using Zend\Code just for the annotations.

gianarb:
why do you remove this dependency from composer.json? This lib use the EventManager in your PR

ThaDafinser:
@gianarb i dont understand u correctly i think.

The EventManager is include in require-dev, since for testing it would be required.
It could be also added to suggest section

gianarb:
Hi! @ThaDafinser for me Zend\EventManager in this moment is a needful requirement to work with annotation, it is not used only during tests, why do you move it into require-dev section?

ThaDafinser:
@gianarb sure its a requirement for your needs, but not for all.
Take Zend\Mvc as an example...it required only some Zend\* componenets: https://github.com/zendframework/zf2/blob/master/library/Zend/Mvc/composer.json#L16-L22
But suggest many of them: https://github.com/zendframework/zf2/blob/master/library/Zend/Mvc/composer.json#L42-L63

Only because you may need a part, you dont take to include that for all.

Just in my case i need exactly 7 classes from Zend\Code, but need to download additionally Zend\EventManager and Zend\StdLib...

gianarb:
Not for me in production, but in my idea of dependencies :) I don't use `Zend\Code` in my projects..
`Zend\Mvc` has the same problem, exists a % of dependency? Below it the requirement becomes require-dev?

For me a dependency is require-dev if it uses by the library for running tests for example `phpunit`, `phpspec` not because is used by few classes

ThaDafinser:
I think its okay to define them in require-dev and maybe suggest.

Iin my case i'm not using annotations, so i dont want to download it.
But when i dont to run `phpunit` i need all dependencies, so require-dev ...

@gianarb  but this seems to be a nice topic to discuss about in general....`to require or not to required`

 :smile:

gianarb:
eheh yes! :) 
The modularity has a price, If you don't use Annotation the solution (maybe ideally) is the split of library..
I think that to move dependencies in dev/suggest is only a good shortcut :yum:

Ocramius:
I strongly disagree on optional dependencies. While it is nice to have one dependency less, we have just hidden a problem and swept it under the carpet.

Dependency should stay, and removing it means that all classes should still work without it, so the PR is either invalid or incomplete right now.

harikt:
I agree with optional dependency argument of @Ocramius . Either it needs a dependency or the component can stand as a different bridge. 

I have suffered a bit configuring a few components for at run time only I was seeing it needs the suggested package. So each time I was forced to call composer require `component` .

ThaDafinser:
I will close it for now then, since earliest date will be 3.0

PR #7230, opened 2015-02-18
DBA inifile is not supported for caching

weierophinney:
We recently noticed failures occuring for the DbaInifile cache storage tests.
The errors reported indicated an inability to perform `dba_replace()`
operations, with no indication of why this was.

On digging through the various DBA documentation on php.net, I discovered that
the `dba_replace` and `dba_insert()` functions are only reliable on GDBM and
QDBM handlers; any other handler may result in unexpected or unintended
behavior.

I have tested against flatfile, GDBM, DB4, and GDBM, and none of these show any
issues; however, inifile reliably has issues at this point. As such, we can no
longer support that version.

This patch raises an exception if the inifile handler is provided as a DBA
option, and the tests have been modified to ensure that the exception is raised.

### DIFF
--- a/library/Zend/Cache/Storage/Adapter/DbaOptions.php
+++ b/library/Zend/Cache/Storage/Adapter/DbaOptions.php
@@ -117,6 +117,12 @@ class DbaOptions extends AdapterOptions
             throw new Exception\ExtensionNotLoadedException("DBA-Handler '{$handler}' not supported");
         }
 
+        if ($handler === 'inifile') {
+            throw new Exception\ExtensionNotLoadedException(
+                "DBA-Handler 'inifile' does not reliably support write operations"
+            );
+        }
+
         $this->triggerOptionEvent('handler', $handler);
         $this->handler = $handler;
         return $this;
--- a/tests/ZendTest/Cache/Storage/Adapter/DbaInifileTest.php
+++ b/tests/ZendTest/Cache/Storage/Adapter/DbaInifileTest.php
@@ -9,11 +9,20 @@
 
 namespace ZendTest\Cache\Storage\Adapter;
 
+use PHPUnit_Framework_TestCase as TestCase;
+use Zend\Cache\Storage\Adapter\Dba;
 
 /**
  * @group      Zend_Cache
  */
-class DbaInifileTest extends AbstractDbaTest
+class DbaInifileTest extends TestCase
 {
-    protected $handler = 'inifile';
+    public function testSpecifyingInifileHandlerRaisesException()
+    {
+        $this->setExpectedException('Zend\Cache\Exception\ExtensionNotLoadedException', 'inifile');
+        $cache = new Dba(array(
+            'pathname' => sys_get_temp_dir() . DIRECTORY_SEPARATOR . uniqid('zfcache_dba_') . '.inifile',
+            'handler'  => 'inifile',
+        ));
+    }
 }


weierophinney:
While this is technically a BC break, it's also a recognition that the support likely was broken already, and should not have been included.

weierophinney:
Some more notes: @Ocramius thinks this is likely due to recent upgrades in the Travis-CI PHP environment. That said, both myself and @GeeH observed the original test failures in our own environments as well, indicating that the functionality may very well be unreliable from one environment to another.

weierophinney:
I have had @ezimuel try this as well, and he was able to reproduce the same errors being seen previously, both with and without the `--enable-dba` flag in the PHP version he used.

ezimuel:
I just tested with PHP 5.6.5 and DBA installed (--enable-dba, --with-gdbm=/usr) and I've the same [19 errors](https://travis-ci.org/zendframework/zf2/jobs/51253876#L1085) of Travis-CI.

marc-mabe:
MH I'm sure I had tested it for years. I also had reported bugs to the php bug tracker but I can't remember this one :(
https://bugs.php.net/search.php?search_for=inifile&boolean=0&limit=30&order_by=&direction=DESC&cmd=display&status=All&bug_type=All&project=All&package_name%5B%5D=DBM%2FDBA+related&php_os=&phpver=&cve_id=&assign=&author_email=&bug_age=0&bug_updated=0

Can you open a new bug report please.

PS: for the cache adapter the inifile handler isn't necessary because of performance reasons.

weierophinney:
@marc-mabe My bigger question is: should we support anything other than the gdbm and qdbm adapters? A [comment on the dba_replace manual page](http://php.net/manual/en/function.dba-replace.php#93372) indicates that these are the only two adapters that can have reliable behavior. However, as you note, searching through the PHP buglist doesn't really show this to be the case. (I found some of the bugs you reported; they weren't related to this.)

Knowing that, then the issue is likely a libdb update, meaning we could potentially skip tests if we know that particular version of libdb will not work. Looking in phpinfo, however, there's no way to determine what version of libdb was compiled against. :-/

This also means, however, that developers cannot rely on the feature across PHP and OS combinations. So we need to make a decision: do we keep inifile support, or remove it? Based on your own feedback, it's likely a bad fit for caching anyways due to performance. As such, this patch appears to be appropriate, and we can message during 2.4 that we no longer support it.

Thoughts?

weierophinney:
BTW, would like _something_ to be merged ASAP so we can know for sure that the develop branch is stabilizing.

marc-mabe:
Support for the inifile handler can be removed because this handler doesn't fit caching requirements.

In normal cases it doesn't matter which handler will be used if all handler work exactly as the API of the DBA extension is described. But we don't live in an ideal world.

What I have seen are bugs in the DBA extension not in the libs wrapped. I would argue that we can be aware of bugs in the extension but if the are bugs in any of the wrapped libs we can't and shouldn't disable the hole adapter because of this. The only we can do is notice the user in the documentation.

weierophinney:
@marc-mabe Cool -- and that's exactly what this is doing. It doesn't disable the adapter, but does raise an exception if you try and use inifile as a handler. A user can always extend the Options class to _not_ have that behavior if they really, really want to use inifile, but we're essentially saying at this point that we recognize it may not work as expected, and, as such, are not maintaining it further.

PR #7233, opened 2015-02-19
Detect https on reversed proxy

snapshotpl:
I have application which works with reversed proxy. Generated urls by serverUrl view helper are with http on https pages. I have found that in $_SERVER exists key HTTP_X_FORWARDED_PROTO with information about scheme.


weierophinney:

PR #7234, opened 2015-02-21
return null if 'where' sql is empty

samson-fp:
When a predicate object with no conditions is passed into 'select', it renders down into '()'.  This results in a query like this which is invalid sql:

select ... where ()


weierophinney:
Please provide a unit test for this.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-db/issues/137

PR #7236, opened 2015-02-21
Add flexibility to FlashMessenger by determining hoops as parameter

ASvi:
Additional flexibility for FlashMessenger. It's better to determite hoops as parameter.

### DIFF
--- a/library/Zend/Mvc/Controller/Plugin/FlashMessenger.php
+++ b/library/Zend/Mvc/Controller/Plugin/FlashMessenger.php
@@ -150,17 +150,18 @@ class FlashMessenger extends AbstractPlugin implements IteratorAggregate, Counta
     /**
      * Add a message
      *
-     * @param  string         $message
+     * @param  string $message
+     * @param int $hoops
      * @return FlashMessenger Provides a fluent interface
      */
-    public function addMessage($message)
+    public function addMessage($message, $hoops = 1)
     {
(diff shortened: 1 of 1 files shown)


library/Zend/Mvc/Controller/Plugin/FlashMessenger.php
```diff
@@ -150,17 +150,18 @@ public function getNamespace()
     /**
      * Add a message
      *
-     * @param  string         $message
+     * @param  string $message
+     * @param int $hoops
      * @return FlashMessenger Provides a fluent interface
      */
-    public function addMessage($message)
+    public function addMessage($message, $hoops = 1)
```
Martin-P: `$hoops` is spelled with 1 `o`: `$hops`.

library/Zend/Mvc/Controller/Plugin/FlashMessenger.php
```diff
@@ -150,17 +150,18 @@ public function getNamespace()
...
     {
         $container = $this->getContainer();
         $namespace = $this->getNamespace();
 
         if (!$this->messageAdded) {
             $this->getMessagesFromContainer();
-            $container->setExpirationHops(1, null);
+            $container->setExpirationHops($hoops, null);
```
Martin-P: Same as above: `$hoops` is spelled with 1 `o`: `$hops`.

PR #7237, opened 2015-02-21
Add flexibility to FlashMessenger by determining hoops as parameter

ASvi:
Additional flexibility for FlashMessenger. It's better to determite hoops as parameter.


Martin-P:
Duplicate with: #7236.

PR #7239, opened 2015-02-21
Fixes #7238 : Logger::log() exception message

### LINKED ISSUE #7238: Logger::log($priority, message) got exception : "$priority must be an integer > 0 and < 8;"
When see-ing the Logger const : 
```php
    const EMERG  = 0;
    const ALERT  = 1;
    const CRIT   = 2;
    const ERR    = 3;
    const WARN   = 4;
    const NOTICE = 5;
    const INFO   = 6;
    const DEBUG  = 7;
```
that 0 = EMERG must be included. it means $priority >=0 ?

samsonasik:
Fixes #7238


weierophinney:

PR #7240, opened 2015-02-21
Feature: Zend\Mvc\HttpMethodListener

stefanotorresi:
as suggested in #6385, here is the event listener to validate standard HTTP request methods and to make the application return a 405 empty response for unsupported methods.

the listener is ready to be registered via the `listeners` key in `application.config.php` with the name `HttpMethodListener`; however, for the time being it's not registered by default as originally suggested by @weierophinney.

to make that happen, we would have to add a key to `Zend\Mvc\Application::$defaultListeners`, and I discovered that this would cause some tests to fail because they assert there is _precisely one_ default listener per event, which is a bit odd. Personally, I would update [these tests](https://github.com/stefanotorresi/zf2/blob/6cc9d3e913bad2051de3ed1c28709dd325eabee7/tests/ZendTest/Mvc/ApplicationTest.php#L140-L186) with a [different assertion](https://github.com/stefanotorresi/zf2/blob/6cc9d3e913bad2051de3ed1c28709dd325eabee7/tests/ZendTest/Mvc/ApplicationTest.php#L195-L203) to make them not interfere with future default listeners additions.

[Here is a proposal to update the relevant tests](https://github.com/stefanotorresi/zf2/commit/edf61d7de77724edb1e4731da1280bac36971548) ready to merge into this PR.

I will wait for for further instructions on the matter.

### DIFF
--- a/library/Zend/Mvc/Application.php
+++ b/library/Zend/Mvc/Application.php
@@ -66,6 +66,7 @@ class Application implements
     protected $defaultListeners = array(
         'RouteListener',
         'DispatchListener',
+        'HttpMethodListener',
         'ViewManager',
         'SendResponseListener',
     );
--- a/library/Zend/Mvc/HttpMethodListener.php
+++ b/library/Zend/Mvc/HttpMethodListener.php
@@ -41,7 +41,7 @@ class HttpMethodListener extends AbstractListenerAggregate
      * @param bool  $enabled
      * @param array $allowedMethods
      */
-    public function __construct($enabled = true, array $allowedMethods = array())
+    public function __construct($enabled = true, $allowedMethods = array())
     {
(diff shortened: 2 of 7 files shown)


weierophinney:
> we would have to add a key to Zend\Mvc\Application::$defaultListeners, and I discovered that this would cause some tests to fail because they assert there is precisely one default listener per event, which is a bit odd. Personally, I would update these tests with a different assertion to make them not interfere with future default listeners additions.

This makes sense to me; to be honest, it's an implementation detail at best, so I think we can relax the restrictions.

Please update the PR with the proposed changes, and I'll review again for 2.4!

stefanotorresi:
@weierophinney done! :satisfied:

Issue #7228, opened 2015-02-18
The function getLastRawRequest (in the Curl adapter in Http/Client) don't work in Http 1.1


fcywier:
Hello,
Since the version 2.3.2 of the framework, the function **getLastRawRequest() in the Curl adapter in the Http/Client don't return the last response**.

The line 339 has changed in Http/Client/Adapter/Curl.php :

```
curl_setopt($this->curl, CURLOPT_HTTP_VERSION, $curlHttp);
```
To
```
curl_setopt($this->curl, $curlMethod, $curlValue);
```
**So CURLOPT_HTTP_VERSION is set to CURL_HTTP_VERSION_1_1 now** (in my case).

To track the handle's request string, we have to set CURLINFO_HEADER_OUT to TRUE in the curl_setopt().

But, in the constructor of Zend\Http\Client\Adapter\Curl, we have a list of "invalidOverwritableCurlOptions".
**In it, we have the option "CURL_HTTP_VERSION_1_1" (integer 2) that invalid the CURLINFO_HEADER_OUT (integer 2 too)**.

I don't know if it is necessary to have the options CURL_HTTP_VERSION_1_1 and CURL_HTTP_VERSION_1_0 in the invalid option table but it broke the getLastRawRequest function in Http 1.1.
I think they can be deleted from this table.

Regards,
Frédéric

Martin-P:
Please see #6485 for background info about this change.

fcywier:
Thanks for this information Martin.
This change highlights the problem in ```Http\Client\Adapter\Curl.php (line 77)``` :

```
        $this->invalidOverwritableCurlOptions = array(
            CURLOPT_HTTPGET,
            CURLOPT_POST,
            CURLOPT_UPLOAD,
            CURLOPT_CUSTOMREQUEST,
            CURLOPT_HEADER,
            CURLOPT_RETURNTRANSFER,
            CURLOPT_HTTPHEADER,
            CURLOPT_POSTFIELDS,
            CURLOPT_INFILE,
            CURLOPT_INFILESIZE,
            CURLOPT_PORT,
            CURLOPT_MAXREDIRS,
            CURLOPT_CONNECTTIMEOUT,
            CURL_HTTP_VERSION_1_1,
            CURL_HTTP_VERSION_1_0,
        );
```
should by replaced by :
```
        $this->invalidOverwritableCurlOptions = array(
            CURLOPT_HTTPGET,
            CURLOPT_POST,
[...]

PR #7241, opened 2015-02-21
Removed non existing curl options

Martin-P:
Fix for #7228.

`CURL_HTTP_VERSION_1_0` and `CURL_HTTP_VERSION_1_1` are not cURL options. They are cURL _values_ for the option `CURLOPT_HTTP_VERSION`.


weierophinney:

PR #7245, opened 2015-02-22
[Navigation] Introducing NavigationAbstractServiceFactory

sandrokeil:
According to the feedback of #6717 I've created the Navigation Abstract Service Factory. If this abstract service factory is registered, a service with name `Zend\Navigation\Special` will use the `special` definition in configuration key `navigation`.

With one point I'm not satisfied. A service call must start with `Zend\Navigation\`  (e.g. `$sl->get('Zend\Navigation\Special')`). Maybe we could support `navigation.special` too?

I use the `ConstructedNavigationFactory` to create the Navigation Container. Is this ok or should I use the `AbstractNavigationFactory` (@weierophinney)? The problem here is that we have to implement the getName() and override the getPages() method. For the future (ZF 3.0) we could remove getName() and make getPages() abstract in `AbstractNavigationFactory`. This makes `DefaultNavigationFactory` and `ConstructedNavigationFactory` needless. But it's a BC break.

I will also update the documentation.

### DIFF
--- a/library/Zend/Navigation/Service/NavigationAbstractServiceFactory.php
+++ b/library/Zend/Navigation/Service/NavigationAbstractServiceFactory.php
@@ -1,0 +1,122 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace Zend\Navigation\Service;
+
+use Zend\Navigation\Navigation;
+use Zend\ServiceManager\AbstractFactoryInterface;
+use Zend\ServiceManager\ServiceLocatorInterface;
+
+/**
+ * Navigation abstract service factory
+ *
+ * Allows configuring several navigation instances. If you have a navigation config key named "special" then you can
+ * use $serviceLocator->get('Zend\Navigation\Special') to retrieve a navigation instance with this configuration.
+ */
+final class NavigationAbstractServiceFactory implements AbstractFactoryInterface
+{
+    /**
+     * Top-level configuration key indicating navigation configuration
+     *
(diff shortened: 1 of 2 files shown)


weierophinney:
> With one point I'm not satisfied. A service call must start with Zend\Navigation\ (e.g. $sl->get('Zend\Navigation\Special')). Maybe we could support navigation.special too?

I think `Zend\Navigation\` as a prefix is a reasonable default, and it's easier to support a single option than a configurable option. Developers can extend the factory if they want to change the prefix.

weierophinney:
> I use the ConstructedNavigationFactory to create the Navigation Container. Is this ok or should I use the AbstractNavigationFactory (@weierophinney)?

`ConstructedNavigationFactory` looks like the appropriate one to use, as it allows you to get the specific configuration section you need. Internally, it's extending `AbstractNavigationFactory` anyways; the difference is how it gets the page configuration, and that functionality is why it exists.

weierophinney:
> The problem here is that we have to implement the getName() and override the getPages() method. For the future (ZF 3.0) we could remove getName() and make getPages() abstract in AbstractNavigationFactory. This makes DefaultNavigationFactory and ConstructedNavigationFactory needless. But it's a BC break.

We can worry about that in a month or two. :)

froschdesign:
@sandrokeil 
Please add also a short example to docs. Thanks!

sandrokeil:
I will do it in the next few days.

froschdesign:
@sandrokeil 
:+1:

razonklnbd:
Developer can't inherit due to its final behavior. Can anyone please explain why this factory using "final" keyword at class declaration?
Also constant "SERVICE_PREFIX" is not using anywhere, I guess it may use somewhere in Zend project!

sandrokeil:
@razonklnbd Why do you want to inherit from this class, it's not necessary. Please consider to create your own factory for your special behaviour.

A good explanation when to use final, can be found [here](http://ocramius.github.io/blog/when-to-declare-classes-final/). That's the case.

razonklnbd:
@sandrokeil Thanks for your explanation.

PR #7248, opened 2015-02-23
Remove CURLOPT_POSTFIELDS from set of invalid curl options

igorwwwwwwwwwwwwwwwwwwww:
Right now there is no way to upload a file without loading the entire file into memory twice.
That is quite a bummer, because CURL supports it just fine. The CURLOPT_POSTFIELDS allows doing
just that.

Now it would of course be exellent if the built in file upload stuff in ZF would just use
said option. But as a workaround, it would be great to be able to override it.

Another thing: Getting this to work properly also requires messing around with the ENC_TYPE
in the client. Basically manually overriding it to ENC_FORMDATA.

### DIFF
--- a/library/Zend/Http/Client/Adapter/Curl.php
+++ b/library/Zend/Http/Client/Adapter/Curl.php
@@ -82,7 +82,6 @@ class Curl implements HttpAdapter, StreamInterface
             CURLOPT_HEADER,
             CURLOPT_RETURNTRANSFER,
             CURLOPT_HTTPHEADER,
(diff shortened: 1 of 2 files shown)


Ocramius:
Requires tests

localheinz:
Igor?

Ocramius:
:+1: 

@localheinz no, he's just a broken `GOTO` of @igorw

danizord:
@Ocramius hahaha :P

PR #7249, opened 2015-02-23
Fix/Multiple Issues in Zend\Mail

yurikuzn:
We use Zend\Mail in our application to check imap. Have been running into some serious issues and trying to fix them.

If is PR is not ok for a merge, just review the issues please.

### DIFF
--- a/library/Zend/Mail/Headers.php
+++ b/library/Zend/Mail/Headers.php
@@ -78,7 +78,7 @@ class Headers implements Countable, Iterator
                 $currentLine = trim($line);
             } elseif (preg_match('/^\s+.*$/', $line, $matches)) {
                 // continuation: append to current line
-                $currentLine .= trim($line);
+                $currentLine .= $line;
             } elseif (preg_match('/^\s*$/', $line)) {
                 // empty line indicates end of headers
                 break;
--- a/library/Zend/Mail/Storage/Part.php
+++ b/library/Zend/Mail/Storage/Part.php
@@ -102,10 +102,14 @@ class Part implements RecursiveIterator, Part\PartInterface
                 $this->headers = new Headers();
                 $this->headers->addHeaders($params['headers']);
             } else {
-                if (empty($params['noToplines'])) {
-                    Mime\Decode::splitMessage($params['headers'], $this->headers, $this->topLines);
+                if ($params['headers'] instanceof \Zend\Mail\Headers) {
+                    $this->headers = $params['headers'];
                 } else {
-                    $this->headers = Headers::fromString($params['headers']);
+                    if (empty($params['noToplines'])) {
+                        Mime\Decode::splitMessage($params['headers'], $this->headers, $this->topLines);
+                    } else {
+                        $this->headers = Headers::fromString($params['headers']);
+                    }
                 }
             }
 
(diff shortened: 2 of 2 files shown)


weierophinney:
Two things:

- Please rebase against current develop branch status (currently out-of-date, meaning we cannot merge without conflict).
- Please provide unit tests for the changes you're proposing.

Xerkus:
parameter value with space MUST be enclosed in double quotes. Provided example is malformed.

yurikuzn:
I don't think so. Couldn't find this in the specification. That example is from real life, some mail client generated such headers. That was not critical issue since it just didn't parse attachments. 2 other issues produced fatal error (couldn't be catched) that caused internal looping since the id of last email was be stored.

Xerkus:
ABNF definitions for header parameters are in RFC2045 and RFC822

    parameter := attribute "=" value
    value := token / quoted-string
    token := 1*<any (US-ASCII) CHAR except SPACE, CTLs, or tspecials>
    quoted-string = <"> *(qtext/quoted-pair) <">

library/Zend/Mail/Storage/Part.php
```diff
@@ -377,9 +385,12 @@ public function __get($name)
      */
     public function __isset($name)
     {
+        $headers = $this->getHeaders();
+        if (empty($headers) || !is_object($headers)) {
```
Xerkus: getHeaders() always returns headers container instance. This change is not needed.

library/Zend/Mail/Storage/Part.php
```diff
@@ -122,6 +126,10 @@ public function __construct(array $params)
      */
     public function isMultipart()
     {
+        if (!isset($this->contentType)) {
```
Xerkus: this part does not change behavior, but it avoids exception and try-catch on missing header. Change is ok.

library/Zend/Mail/Storage/Part.php
```diff
@@ -102,10 +102,14 @@ public function __construct(array $params)
                 $this->headers = new Headers();
                 $this->headers->addHeaders($params['headers']);
             } else {
```
Xerkus: elseif

library/Zend/Mail/Storage/Part.php
```diff
@@ -102,10 +102,14 @@ public function __construct(array $params)
...
-                if (empty($params['noToplines'])) {
-                    Mime\Decode::splitMessage($params['headers'], $this->headers, $this->topLines);
+                if ($params['headers'] instanceof \Zend\Mail\Headers) {
+                    $this->headers = $params['headers'];
```
Xerkus: Headers container should be cloned to avoid container sharing between several messages.

library/Zend/Mail/Headers.php
```diff
@@ -78,7 +78,7 @@ public static function fromString($string, $EOL = self::EOL)
                 $currentLine = trim($line);
```
Xerkus: Remove trim here as well.

library/Zend/Mail/Headers.php
```diff
@@ -78,7 +78,7 @@ public static function fromString($string, $EOL = self::EOL)
...
             } elseif (preg_match('/^\s+.*$/', $line, $matches)) {
                 // continuation: append to current line
-                $currentLine .= trim($line);
+                $currentLine .= $line;
```
Xerkus: This appears to be already fixed

PR #7251, opened 2015-02-23
Fixed tests for quoteValue() methods in all Platform classes

hschletz:
The tests for quoteValue() are incomplete. When quoteValue() raises an E_USER_NOTICE, PHPUnit converts it to an exception which is tested. The tests try to check the return value, but neither the return value generating code nor the assertion are executed. The tests pass, but only the error message is effectively tested.

I split the tests: One test checks the exception as usual, but the return value is discarded (It would be NULL anyway within a PHPUnit environment - in other context a value would be returned, but that is difficult to test). I also tightened the exception type check from PHPUnit_Framework_Error to PHPUnit_Framework_Error_Notice.

The other test executes quoteValue() with error suppression so that the return value can be checked. Since quoteValue() and quoteTrustedValue() are implemented separately, I added the full set of test values to the test.

### DIFF
--- a/tests/ZendTest/Db/Adapter/Platform/PostgresqlTest.php
+++ b/tests/ZendTest/Db/Adapter/Platform/PostgresqlTest.php
@@ -74,13 +74,24 @@ class PostgresqlTest extends \PHPUnit_Framework_TestCase
     /**
      * @covers Zend\Db\Adapter\Platform\Postgresql::quoteValue
      */
-    public function testQuoteValue()
+    public function testQuoteValueRaisesNoticeWithoutPlatformSupport()
     {
         $this->setExpectedException(
-            'PHPUnit_Framework_Error',
+            'PHPUnit_Framework_Error_Notice',
             'Attempting to quote a value in Zend\Db\Adapter\Platform\Postgresql without extension/driver support can introduce security vulnerabilities in a production environment'
         );
-        $this->assertEquals("'value'", $this->platform->quoteValue('value'));
+        $this->platform->quoteValue('value');
+    }
+
+    /**
+     * @covers Zend\Db\Adapter\Platform\Postgresql::quoteValue
+     */
+    public function testQuoteValue()
+    {
+        $this->assertEquals("E'value'", @$this->platform->quoteValue('value'));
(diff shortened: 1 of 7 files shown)


tests/ZendTest/Db/Adapter/Platform/PostgresqlTest.php
```diff
@@ -74,13 +74,24 @@ public function testGetQuoteValueSymbol()
...
+
+    /**
+     * @covers Zend\Db\Adapter\Platform\Postgresql::quoteValue
+     */
+    public function testQuoteValue()
+    {
+        $this->assertEquals("E'value'", @$this->platform->quoteValue('value'));
+        $this->assertEquals("E'Foo O\\'Bar'", @$this->platform->quoteValue("Foo O'Bar"));
+        $this->assertEquals('E\'\\\'; DELETE FROM some_table; -- \'', @$this->platform->quoteValue('\'; DELETE FROM some_table; -- '));
+        $this->assertEquals("E'\\\\\\'; DELETE FROM some_table; -- '", @$this->platform->quoteValue('\\\'; DELETE FROM some_table; -- '));
```
weierophinney: Can you please check this against the changes introduced to fix [ZF2015-02](http://framework.zend.com/security/advisory/ZF2015-02)? I'm pretty sure we changed escaping to better match the recommendations made by PostgreSQL.

hschletz:
Yes, the PR is based on yesterday's snapshot of the develop branch, which contains the fix in afe417248d71907b5f913e1e515f2b54bf1b0502. The improvement to quoteValue() and quoteTrustedValue() is the 'E' prefix. This had been added to testQuoteTrustedValue(), but forgotten in testQuoteValue(), where it made no difference for the reason mentioned in the initial post.

This PR contains the updated assertions copied to testQuoteValue(), in addition to the structural changes to the test.

weierophinney:
@hschletz Can you please rebase against develop, so  I can merge? I need that done by tomorrow, 19 March 2015; otherwise, we'll have to review after the 2.4 release.

hschletz:
Rebased against latest HEAD, merged into single commit.

PR #7254, opened 2015-02-24
Ship a StandaloneExtensionManager for Zend\Feed\Reader

### LINKED ISSUE #6419: missing dependency for zendframework/zend-feed stand-alone

weierophinney:
In 2.2 or 2.3, we modified `Zend\Feed` to reduce the number of dependencies, and introduced a separated interface for the extension manager to allow using a non-ServiceManager variant (which allowed reducing dependencies).

This patch completes that work by introducing a `StandaloneExtensionManager` that provides the base capabilities necessary to serve the shipped feed and entry extensions. `Zend\Feed\Reader\Reader::getExtensionManager()` has been modified to lazy-load this class when no other extension manager has been set.

This implementation does NOT support registering custom extensions, as the `ExtensionManagerInterface` does not specify such capabilities. If users want to allow or support custom extensions, they would need to create their own implementation or extend the current implementation.

This patch replaces #7179 and will resolve #6419.

### DIFF
--- a/tests/ZendTest/Feed/Reader/StandaloneExtensionManagerTest.php
+++ b/tests/ZendTest/Feed/Reader/StandaloneExtensionManagerTest.php
@@ -1,0 +1,70 @@
+<?php
+namespace ZendTest\Feed\Reader;
+
+use PHPUnit_Framework_TestCase as TestCase;
+use Zend\Feed\Reader\StandaloneExtensionManager;
+
+class StandaloneExtensionManagerTest extends TestCase
+{
+    public function setUp()
+    {
+        $this->extensions = new StandaloneExtensionManager();
+    }
+
+    public function testIsAnExtensionManagerImplementation()
+    {
+        $this->assertInstanceOf('Zend\Feed\Reader\ExtensionManagerInterface', $this->extensions);
+    }
+
+    public function defaultPlugins()
+    {
+        return array(
+            'Atom\Entry'            => array('Atom\Entry', 'Zend\Feed\Reader\Extension\Atom\Entry'),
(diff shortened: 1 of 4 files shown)


tests/ZendTest/Feed/Reader/StandaloneExtensionManagerTest.php
```diff
@@ -0,0 +1,70 @@
+<?php
```
samsonasik: missing license here...

tests/ZendTest/Feed/Reader/StandaloneExtensionManagerTest.php
```diff
@@ -0,0 +1,70 @@
...
+namespace ZendTest\Feed\Reader;
+
+use PHPUnit_Framework_TestCase as TestCase;
+use Zend\Feed\Reader\StandaloneExtensionManager;
+
+class StandaloneExtensionManagerTest extends TestCase
+{
+    public function setUp()
+    {
+        $this->extensions = new StandaloneExtensionManager();
```
samsonasik: need <code>protected $extensions</code> as class property

weierophinney: No, we don't. We often omit the property declarations and just use public properties in the test suite.

tests/ZendTest/Feed/Reader/StandaloneExtensionManagerTest.php
weierophinney: Fixed.

PR #7140, opened 2015-01-20
[wip] Added $data param to deleteList

TomHAnderson:
deleteList was not passed any data.  This corrects the abstract function declaration for deleteList and adds data parsing and passing to the onDispatch event.

PR #7255, opened 2015-02-24
Added $data param to deleteList

TomHAnderson:
This was a [wip] which was invalidated when the branch was rewritten.  Was https://github.com/zendframework/zf2/pull/7140

### DIFF
--- a/library/Zend/Mvc/Controller/AbstractRestfulController.php
+++ b/library/Zend/Mvc/Controller/AbstractRestfulController.php
@@ -116,7 +116,7 @@ abstract class AbstractRestfulController extends AbstractController
      *
      * @return mixed
      */
-    public function deleteList()
+    public function deleteList($data)
     {
         $this->response->setStatusCode(405);
 
@@ -347,6 +347,8 @@ abstract class AbstractRestfulController extends AbstractController
             // DELETE
             case 'delete':
                 $id = $this->getIdentifier($routeMatch, $request);
+                $data = $this->processBodyContent($request);
+
                 if ($id !== false) {
                     $action = 'delete';
                     $return = $this->delete($id);
@@ -354,7 +356,7 @@ abstract class AbstractRestfulController extends AbstractController
                 }
 
                 $action = 'deleteList';
-                $return = $this->deleteList();
+                $return = $this->deleteList($data);
                 break;
             // GET
             case 'get':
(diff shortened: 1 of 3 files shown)


weierophinney:
This needs a test for the new behavior, please.

TomHAnderson:
Modified existing deleteList to send array of ids.

tests/ZendTest/Mvc/Controller/RestfulControllerTest.php
```diff
@@ -137,9 +137,15 @@ public function testDispatchInvokesDeleteMethodWhenNoActionPresentAndDeleteInvok
...
     {
-        $this->request->setMethod('DELETE');
+        $entities = array(
+            array('id' => uniqid(), 'name' => __FUNCTION__),
+            array('id' => uniqid(), 'name' => __FUNCTION__),
+            array('id' => uniqid(), 'name' => __FUNCTION__),
+        );
+        $string = http_build_query($entities);
+        $this->request->setMethod('DELETE')
+                      ->setContent($string);
```
weierophinney: While this sets up data to pass to the controller, it does not test that the controller received that data. You need to do something with the data that you can then assert against later.

TomHAnderson:
The Travis test failures in PHP 5.5 & 5.6 do not correlate to the changes in this PR.

weierophinney:
I need to revert this, and immediately issue a 2.3.7 that removes it. The feature as merged is a BC break, as it adds a required argument to an existing signature; any code that extended it previously breaks — including the intended use case in zf-rest's `RestController` (tests break under 2.3.6, as reported by a number of users who updated their ZF versions after the release was made).

My plan is:

- Revert in master, and immediately tag and release 2.3.7
- Issue a new PR against develop that makes the argument optional.

library/Zend/Mvc/Controller/AbstractRestfulController.php
```diff
@@ -116,7 +116,7 @@ public function delete($id)
      *
      * @return mixed
      */
-    public function deleteList()
+    public function deleteList($data)
```
TomHAnderson: To avert a BC break my understanding is this line should be 
```
public function deleteList($data = null)
```

weierophinney:
This is now fixed in develop. Interestingly, making the argument optional is still a signature break, and, as such, I'm retaining the "BC Break" label so we can call it out in the upgrade notes.

PR #7258, opened 2015-02-25
Db Platform refactoring

hschletz:
This is a follow-up to #7059 against the latest develop branch. A common base class (AbstractPlatform) with some common implementations is already present there. This PR moves more methods to the base class, reducing the amount of duplicated code.

I double-checked that the critical quoteIdentifierChain(), quoteValue() and quoteTrustedValue() methods are inherited the right way. A second set of eyeballs should verify this, since the unit tests don't cover all possible characters that need escaping. The other quoting methods had already been ported.

### DIFF
--- a/library/Zend/Db/Adapter/Platform/AbstractPlatform.php
+++ b/library/Zend/Db/Adapter/Platform/AbstractPlatform.php
@@ -78,6 +78,14 @@ abstract class AbstractPlatform implements PlatformInterface
     /**
      * {@inheritDoc}
      */
+    public function quoteIdentifierChain($identifierChain)
+    {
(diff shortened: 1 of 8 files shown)


weierophinney:
@Ocramius Can you look over this, please?

weierophinney:
@hschletz This needs a rebase; I cannot merge this as-is.

@Ocramius Thoughts?

weierophinney:
Pinging @Ocramius — just need a :+1: or :-1: from you; I can do the merge.

hschletz:
Rebased against latest HEAD.

Ocramius:
@weierophinney :+1:

PR #7260, opened 2015-02-25
Add method to check that status code is correct

snapshotpl:


weierophinney:
Rebase this against current develop branch, please.

snapshotpl:
I think a little about this and I think static method will be better. What are think?

snapshotpl:
@weierophinney done

library/Zend/Http/Response.php
```diff
@@ -250,6 +249,17 @@ public function setStatusCode($code)
...
 
     /**
+     * Check is correct status code
+     *
+     * @param int $code
+     * @return bool
+     */
+    public static function isStatusCode($code)
+    {
+        return is_numeric($code) && defined('static::STATUS_CODE_' . $code);
```
malukenho: I don't know if that's the best way to check if a status code is valid, because I can create my own status code `555` for a response, and I can't know if is a status code in this way.

snapshotpl: This method check statuses defined in rfc

weierophinney: @malukenho The current behavior is such that it doesn't allow custom status codes, either.

I'm wondering if we should keep enforcing that or not.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-http/issues/75

PR #7268, opened 2015-02-26
Correct errors reported by php-cs-fixer

weierophinney:
Only observed using php-cs-fixer 1.5 on PHP 5.6, but causing 5.6 to fail currently.


weierophinney:
Now building here: https://travis-ci.org/weierophinney/zf2/builds/52310597

PR #7276, opened 2015-02-28
Adds preserve defined order as a factory option

carnage:


tests/ZendTest/Mvc/Service/FormAnnotationBuilderFactoryTest.php
```diff
@@ -18,17 +18,29 @@ class FormAnnotationBuilderFactoryTest extends TestCase
     public function testCreateService()
     {
         $mockElementManager = $this->getMock('Zend\Form\FormElementManager');
-        $mockValidatorManager = $this->getMock('Zend\Validator\ValidatorPluginManager');
-        $mockFilterManager = $this->getMock('Zend\Filter\FilterPluginManager');
```
weierophinney: Why are you removing these?

weierophinney: Granted, tests pass without them, so likely they aren't needed.

carnage: They were required by a previous implementation of the factory; I improved but hadn't removed them from the test

Issue #7244, opened 2015-02-21
Array to string conversion in Zend\Log\Writer\Db


Martin-P:
This was already reported in #2589, but that issue got closed 11 days ago (https://github.com/zendframework/zf2/issues/2589#issuecomment-73774768) with references to:
- PR #5254 which is about a db adapter via config and is unrelated to the issue
- Issue #5603 which is a duplicate of the issue

I tried to get it reopened, but without any success.

Testcase which shows the problem:
```php
/**
 * @group 2589
 */
public function testMapEventIntoColumnDoesNotTriggerArrayToStringConversion()
{
    $this->writer = new DbWriter($this->db, $this->tableName, array(
        'priority' => 'new-priority-field',
        'message'  => 'new-message-field' ,
        'extra'    => array(
            'file'  => 'new-file',
            'line'  => 'new-line',
            'trace' => 'new-trace',
        )
    ));

    // log to the mock db adapter
    $priority = 2;
    $message  = 'message-to-log';
    $extra    = array(
        'file'  => 'test.php',
        'line'  => 1,
        'trace' => array(
            array(
                'function' => 'Bar',
                'class'    => 'Foo',
                'type'     => '->',
                'args'     => array(
                    'baz',
                ),
            ),
        ),
    );
    $this->writer->write(array(
        'priority' => $priority,
        'message'  => $message,
        'extra'    => $extra,
    ));

    $this->assertContains('query', array_keys($this->db->calls));
    $this->assertEquals(1, count($this->db->calls['query']));

    foreach ($this->db->calls['execute'][0][0] as $fieldName => $fieldValue) {
        $this->assertInternalType('string', $fieldName);
        $this->assertInternalType('string', (string) $fieldValue);
    }
}
```

Result:
> 1) ZendTest\Log\Writer\DbTest::testMapEventIntoColumnDoesNotTriggerArrayToStringConversion
Array to string conversion

weierophinney:
@Martin-P Are you working on a PR for this, by any chance?

Martin-P:
@weierophinney 
Looking at it right now, but there are some issues:
- `serialize()`: not an option, because it can't handle closures
- `var_export()`: not an option, because it will create a huge string which leads to situations as in #2592 (every class with all its properties is exported)
- `print_r()`: not an option, because it also creates a huge string where all (nested) classes and properties are exported
- `json_encode()`: works, but objects are stored as stdClass and this causes loss of data

[...]

PR #7282, opened 2015-02-28
Fix: Array to string conversion in Zend\Log\Writer\Db

Martin-P:
Fix for issue #7244 where `Zend\Log\Writer\Db` can trigger an array to string conversion.

Because the method `eventIntoColumn` has the same issue as `mapEventIntoColumn`, I applied the patch to `eventIntoColumn` as well.

### DIFF
--- a/library/Zend/Log/Writer/Db.php
+++ b/library/Zend/Log/Writer/Db.php
@@ -163,7 +163,12 @@ class Db extends AbstractWriter
             if (is_array($value)) {
                 foreach ($value as $key => $subvalue) {
                     if (isset($columnMap[$name][$key])) {
-                        $data[$columnMap[$name][$key]] = $subvalue;
+                        if (is_scalar($subvalue)) {
+                            $data[$columnMap[$name][$key]] = $subvalue;
+                            continue;
+                        }
+
+                        $data[$columnMap[$name][$key]] = var_export($subvalue, true);
                     }
                 }
             } elseif (isset($columnMap[$name])) {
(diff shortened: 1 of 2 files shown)


Martin-P:
Please don't merge this yet. I am looking into a better approach for this by changing `Zend\Log\Formatter\Db` and have it use [`Zend\Log\Formatter\Base::normalize()`](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/library/Zend/Log/Formatter/Base.php#L72) which seems to created for this specific issue (*Normalize all non-scalar data types (except null) in a string value*).

Martin-P:
Nevermind, my other idea won't work, because the formatter is unaware of the database columns and therefore does not know if an array should be converted to string.

library/Zend/Log/Writer/Db.php
```diff
@@ -163,7 +163,12 @@ protected function mapEventIntoColumn(array $event, array $columnMap = null)
             if (is_array($value)) {
                 foreach ($value as $key => $subvalue) {
                     if (isset($columnMap[$name][$key])) {
-                        $data[$columnMap[$name][$key]] = $subvalue;
+                        if (is_scalar($subvalue)) {
+                            $data[$columnMap[$name][$key]] = $subvalue;
+                            continue;
+                        }
+
+                        $data[$columnMap[$name][$key]] = var_export($subvalue, true);
```
weierophinney: I wonder if this should be `json_encode()` instead. Thoughts?

Martin-P: When using `json_encode()` objects are stored as stdClass and not as the original class name and this causes loss of data. I looked at other possibilities too (see this comment  https://github.com/zendframework/zf2/issues/7244#issuecomment-76534596 and the one after that).

`var_export()` keeps the data readable. If the data needs to be processed further, the method [`Zend\Log\Formatter\Base::normalize()`](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/library/Zend/Log/Formatter/Base.php#L72) provides a nice way to do this, but I don't see how I can use that method (see my previous comment https://github.com/zendframework/zf2/pull/7282#issuecomment-76558776).

weierophinney: Fair enough!

weierophinney:

PR #7289, opened 2015-03-04
Default Expression $valueParameter to empty array

marcelto:
Adding an expression to a where object without any parameters will throw an exception unless an empty array is passed.  

$select->where->expression('1 = 0', []);//works
$select->where->expression('1 = 0');//will throw an exception

No parameters are being specified, it would be nice not to have to pass an empty array.  Also, this was not an issue in earlier versions prior to the /_[, $valueParameter, ... ]_/ addition.

### DIFF
--- a/library/Zend/Db/Sql/Predicate/Expression.php
+++ b/library/Zend/Db/Sql/Predicate/Expression.php
@@ -19,7 +19,7 @@ class Expression extends BaseExpression implements PredicateInterface
      * @param string $expression
      * @param int|float|bool|string|array $valueParameter
      */
-    public function __construct($expression = null, $valueParameter = null /*[, $valueParameter, ... ]*/)
+    public function __construct($expression = null, $valueParameter = array() /*[, $valueParameter, ... ]*/)
     {
         if ($expression) {
             $this->setExpression($expression);


library/Zend/Db/Sql/Predicate/Expression.php
```diff
@@ -19,7 +19,7 @@ class Expression extends BaseExpression implements PredicateInterface
      * @param string $expression
      * @param int|float|bool|string|array $valueParameter
      */
-    public function __construct($expression = null, $valueParameter = null /*[, $valueParameter, ... ]*/)
+    public function __construct($expression = null, $valueParameter = array() /*[, $valueParameter, ... ]*/)
```
Pittiplatsch: I can't see how this fixes your issue.
`array_slice(func_get_args(), 1)` does not care about default values, and returns an empty array no matter if `$valueParameter = null` or `$valueParameter = array()`.

marcelto: If $valueParameter is not passed and remains null it doesn't get to array_slice

Pittiplatsch: `$foo = is_array($valueParameter) ? $valueParameter : array_slice(func_get_args(), 1)`
is identical to
```
if (is_array($valueParameter)) {
    $foo = $valueParameter;
} else {
    $foo = array_slice(func_get_args(), 1);
}
```
So, with `$valueParameter == null` i.e. `!is_array($valueParameter)` the `array_slice()` part is rendered :-?

marcelto: Wow, sorry, new baby, not a lot of sleep.  There is definitely an issue here though.  I'll take another look.

PR #7290, opened 2015-03-04
Allow literal zero string as sql expression

marcelto:
A sql expression cannot currently be a literal zero (useful in certain unions) since setExpression will never be called.

$select = new Select('foo');
$select->columns([
   'bar' => new Expression('0') //fails, '0' is never set as the expression and the default empty string is used instead resulting in malformed sql
]);

### DIFF
--- a/library/Zend/Db/Sql/Expression.php
+++ b/library/Zend/Db/Sql/Expression.php
@@ -38,7 +38,7 @@ class Expression implements ExpressionInterface
      */
     public function __construct($expression = '', $parameters = null, array $types = array())
     {
-        if ($expression) {
+        if ($expression !== '') {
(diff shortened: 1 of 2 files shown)


malukenho:
Missing unit tests and fix build.

weierophinney:
@malukenho We have errors on the master branch right now due to a change in the Travis environment that causes 5.5 and 5.6 to error for `Zend\Cache`; the build errors are not related to this change.

However, @marcelto : we **do** need unit tests for this before I can accept it to merge.

malukenho:
thanks for the explanation @weierophinney

marcelto:
added new test case

weierophinney:

PR #7291, opened 2015-03-04
Add getExpression method to ExpressionInterface

marcelto:
The Sql Select object expects this interface to have the method getExpression.
...
if ($joinName instanceof ExpressionInterface) {
                $joinName = $joinName->getExpression();
...


Pittiplatsch:
Adding methods to an existing interface results in a BC break!

PR #7292, opened 2015-03-04
Fix call to method not in ExpressionInterface

marcelto:
The ExpressionInterface does not have the method getExpression

### DIFF
--- a/library/Zend/Db/Sql/Select.php
+++ b/library/Zend/Db/Sql/Select.php
@@ -739,18 +739,20 @@ class Select extends AbstractSql implements SqlInterface, PreparableSqlInterface
             } else {
                 $joinName = $join['name'];
             }
+
             if ($joinName instanceof Expression) {
                 $joinName = $joinName->getExpression();
             } elseif ($joinName instanceof TableIdentifier) {
                 $joinName = $joinName->getTableAndSchema();
                 $joinName = ($joinName[1] ? $platform->quoteIdentifier($joinName[1]) . $platform->getIdentifierSeparator() : '') . $platform->quoteIdentifier($joinName[0]);
+            } elseif ($joinName instanceof Select) {
+                $joinName = '(' . $this->processSubSelect($joinName, $platform, $driver, $parameterContainer) . ')';
+            } elseif (is_string($joinName) || (is_object($joinName) && method_exists($joinName, '__toString'))) {
+                $joinName = $platform->quoteIdentifier($joinName);
             } else {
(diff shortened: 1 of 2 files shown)


weierophinney:
@marcelto Can you please provide a unit test that would hit this particular condition?

marcelto:
Ok,the test case has been added.  I ended up changing `Select::processJoins` a bit more for the test case though.

library/Zend/Db/Sql/Select.php
```diff
@@ -739,18 +739,20 @@ protected function processJoins(PlatformInterface $platform, DriverInterface $dr
...
-            if ($joinName instanceof ExpressionInterface) {
+
+            if ($joinName instanceof Expression) {
                 $joinName = $joinName->getExpression();
             } elseif ($joinName instanceof TableIdentifier) {
                 $joinName = $joinName->getTableAndSchema();
                 $joinName = ($joinName[1] ? $platform->quoteIdentifier($joinName[1]) . $platform->getIdentifierSeparator() : '') . $platform->quoteIdentifier($joinName[0]);
+            } elseif ($joinName instanceof Select) {
+                $joinName = '(' . $this->processSubSelect($joinName, $platform, $driver, $parameterContainer) . ')';
+            } elseif (is_string($joinName) || (is_object($joinName) && method_exists($joinName, '__toString'))) {
```
Pittiplatsch: Use `is_callable()` instead of `__toString()`.

Martin-P: Can the magic method `__toString` be something else than `public`?

marcelto:
So it seems it can be but it shouldn't be anything but public.  Using `php 5.5` a simple test script shows that __toString will return a string in an echo as a protected or private method, however, a warning will be thrown.  Using `is_callable` seems good to me but I defer to you.
    

    Warning: The magic method __toString() must have public visibility and cannot be static in toStringTest.php on line 16
    
    
    Warning: The magic method __toString() must have public visibility and cannot be static in toStringTest.php on line 26
    
    I am an A string
    I am a B string
    I am a C string
    
    
    A::__toString (public) is callable
    B::__toString (protected) is not callable
    C::__toString (private) is not callable

library/Zend/Db/Sql/Select.php
weierophinney: Updated to `is_callable(array($joinName, '__toString'))` during merge.

weierophinney:

PR #7293, opened 2015-03-04
Expression check $valueParameter !== null

marcelto:
Adding an expression to a where object without any parameters will throw an exception unless an empty array is passed.

$select->where->expression('1 = 0', []);//works
$select->where->expression('1 = 0');//will throw an exception

No parameters are being specified, it would be nice not to have to pass an empty array. Also, this was not an issue in earlier versions prior to the /[, $valueParameter, ... ]/ addition.

### DIFF
--- a/tests/ZendTest/Db/Sql/Predicate/PredicateTest.php
+++ b/tests/ZendTest/Db/Sql/Predicate/PredicateTest.php
@@ -227,15 +227,20 @@ class PredicateTest extends TestCase
             array(array('foo = %s', array(0), array(Expression::TYPE_VALUE))),
             $predicate->getExpressionData()
         );
-
-        //test with no parameters
-        $predicate = new Predicate;
-        $this->assertSame($predicate, $predicate->expression('foo = bar'));
-        $this->assertEquals(
-            array(array('foo = bar', array(), array())),
-            $predicate->getExpressionData()
-        );
     }
+
+    /**
+    * @group 7293
+    */
+   public function testExpressionParametersDefaultToArray()
+   {
+       $predicate = new Predicate;
+       $this->assertSame($predicate, $predicate->expression('foo = bar'));
+       $this->assertEquals(
+           array(array('foo = bar', array(), array())),
+           $predicate->getExpressionData()
+       );
+   }
 
     /**
      * @testdox Unit test: Test literal() is chainable, returns proper values, and is backwards compatible with 2.0.*
--- a/tests/ZendTest/Db/Sql/Predicate/PredicateTest.php
+++ b/tests/ZendTest/Db/Sql/Predicate/PredicateTest.php
@@ -227,6 +227,14 @@ class PredicateTest extends TestCase
             array(array('foo = %s', array(0), array(Expression::TYPE_VALUE))),
             $predicate->getExpressionData()
         );
+
+        //test with no parameters
+        $predicate = new Predicate;
+        $this->assertSame($predicate, $predicate->expression('foo = bar'));
+        $this->assertEquals(
+            array(array('foo = bar', array(), array())),
+            $predicate->getExpressionData()
+        );
     }
 
     /**
--- a/library/Zend/Db/Sql/Predicate/Expression.php
+++ b/library/Zend/Db/Sql/Predicate/Expression.php
@@ -25,10 +25,6 @@ class Expression extends BaseExpression implements PredicateInterface
             $this->setExpression($expression);
         }
 
-        if ($valueParameter !== null) {
-            $this->setParameters(is_array($valueParameter) ? $valueParameter : array_slice(func_get_args(), 1));
-        } else {
-            $this->setParameters(array());
-        }
+        $this->setParameters(is_array($valueParameter) ? $valueParameter : array_slice(func_get_args(), 1));
     }
 }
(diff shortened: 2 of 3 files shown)


Ocramius:
Requires a test case

Pittiplatsch:
Sorry, but I really don't get you. As I already stated in #7289, there is NO problem with `$valueParam == null`.
`$this->setParameters(is_array($valueParameter) ? $valueParameter : array_slice(func_get_args(), 1));`
already results in 
`$this->setParameters(array());`
when `$valueParam == null`, because it is left out. `func_get_args()` ignores optional params being left out - that's what `func_get_args()` is all about: respect actually given function arguments!

Just an idea: Do you EXPLICITLY pass `$valueParam` as `null`? This obviously and correclty results in 
`$this->setParameters(array(null));`

Martin-P:
> Adding an expression to a where object without any parameters will throw an exception unless an empty array is passed.

Please take a look at this unittest: [`ZendTest\Db\Sql\Predicate\ExpressionTest::testCanPassNoParameterToConstructor`](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/tests/ZendTest/Db/Sql/Predicate/ExpressionTest.php#L35-L42)

That test passes, which suggests it is already possible to leave out the parameters. If there is an issue with your use case, it is not caused by `Zend\Db\Sql\Predicate\Expression`.

Martin-P:
Looking at the code shows the problem is not in `Zend\Db\Sql\Predicate\Expression`, but in [`Zend\Db\Sql\Predicate\Predicate::expression()`](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/library/Zend/Db/Sql/Predicate/Predicate.php#L238-L254). *That* method expects a second argument and should trigger a PHP error (not an exception). Perhaps you can post the exception you get?

marcelto:
Currently I'm passing an empty array for these calls. Otherwise I receive a fatal error and an exception.  It looks like defaulting $parameters to an empty array in Zend\Db\Sql\Predicate\Predicate::expression() fixes the issue.  Would this be the more agreeable solution?

Fatal error: Uncaught exception 'Zend\Db\Sql\Exception\RuntimeException' with message 'The number of replacements in the expression does not match the number of parameters' in '...vendor\zendframework\zendframework\library\Zend\Db\Sql\Expression.php on line 143

Zend\Db\Sql\Exception\RuntimeException: The number of replacements in the expression does not match the number of parameters in '...\vendor\zendframework\zendframework\library\Zend\Db\Sql\Expression.php on line 143

Pittiplatsch:
> It looks like defaulting $parameters to an empty array in Zend\Db\Sql\Predicate\Predicate::expression() fixes the issue.

Defaulting an empty array seems to be a valid, simple solution.
Defaulting to null could also work, but would be far more complex, as the same logic like in `Zend\Db\Sql\Predicate\Expression` (via `func_get_args()`) had to be implemented to get things work correctly.

IMO, an empty array is simple fine here.

Either possibility however needs new tests as well ;-)

marcelto:
Hmmm....If you wanted the functionality of `/*[, $valueParameter, ... ]*/` to be accessible via the `where` wouldn't the `func_get_args()` logic be necessary?

Pittiplatsch:
Yepp, that's what I meant with "far more complex". However, I wouldn't expose this weird functionality to `Predicate::expression()`, as this kind of parameter handling in the long run usually causes more problems than it solves IMO. E.g. extending the parameter list later on due to some kind of extended behaviour is almost impossible when working with param lists of dynamic length. Dunno why this IMO weird param handling made it into `Zend\Db\Sql\Predicate\Expression` at all...

marcelto:
a test has been added

tests/ZendTest/Db/Sql/Predicate/PredicateTest.php
```diff
@@ -227,6 +227,14 @@ public function testExpression()
             array(array('foo = %s', array(0), array(Expression::TYPE_VALUE))),
             $predicate->getExpressionData()
         );
+
+        //test with no parameters
```
Martin-P: I think it is better to add a new test instead of adding to existing ones. A new test would be:
```php
public function testExpressionParametersDefaultToArray()
{
    $predicate = new Predicate;
    $this->assertSame($predicate, $predicate->expression('foo = bar'));
    $this->assertEquals(
        array(array('foo = bar', array(), array())),
        $predicate->getExpressionData()
    );
}
```

weierophinney:
I tried merging this just now, but the new test actually fails:

```
1) ZendTest\Db\Sql\Predicate\PredicateTest::testExpressionParametersDefaultToArray
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => Array (...)
+    0 => 'foo = bar'
 )

tests/ZendTest/Db/Sql/Predicate/PredicateTest.php:242
```

(note: I separated the test into it's own method, per @Martin-P , placing it directly below the method in which @marcelto originally placed it in)

In other words, the patch is not working as expected. I'll delay it for a later release.

marcelto:
I added the new/separate test case.  It passes locally and travis passed 5.3 and 5.4.

Martin-P:
>  It passes locally and travis passed 5.3 and 5.4. 

Please look again, because it does not pass on Travis at all.

marcelto:
I could have sworn the first two passed.  My apologies, I must have been looking at the wrong build.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-db/issues/133

PR #7296, opened 2015-03-05
Fix infinite loop when chaining peering service managers

fabiang:
Hello,

we have a use case where we add a peering service manager (SM) to the SM of an ZF2 application. Also the peering SM becomes the SM of the application added as peering SM, which allows us to call services from the applications SM.

For services which can't be found this runs into a infinite loop. The applications SM calls the peering SM, which calls back the application SM and so on.

This PR tries to fix that by passing the current SM instance to the peering SM so they skip the caller SM when looping over `peeringServiceManagers`.

### DIFF
--- a/library/Zend/ServiceManager/ServiceManager.php
+++ b/library/Zend/ServiceManager/ServiceManager.php
@@ -115,6 +115,11 @@ class ServiceManager implements ServiceLocatorInterface
     protected $canonicalNamesReplacements = array('-' => '', '_' => '', ' ' => '', '\\' => '', '/' => '');
 
     /**
+     * @var ServiceLocatorInterface
+     */
+    protected $serviceManagerCaller;
+
+    /**
      * Constructor
      *
      * @param ConfigInterface $config
@@ -446,7 +451,7 @@ class ServiceManager implements ServiceLocatorInterface
         if (!isset($this->shared[$cName])) {
             return $this->shareByDefault();
         }
-        
+
         return $this->shared[$cName];
     }
 
@@ -718,9 +723,16 @@ class ServiceManager implements ServiceLocatorInterface
         }
 
         if ($usePeeringServiceManagers) {
+
+            $caller = $this->serviceManagerCaller;
             foreach ($this->peeringServiceManagers as $peeringServiceManager) {
-                if ($peeringServiceManager->has($name)) {
-                    return true;
+                // ignore peering service manager if they are the same instance
+                if ($caller !== $peeringServiceManager) {
+                    $peeringServiceManager->serviceManagerCaller = $this;
+                    if ($peeringServiceManager->has($name)) {
+                        return true;
+                    }
+                    $peeringServiceManager->serviceManagerCaller = null;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


fabiang:
* made test for chained service managers more clear
* reset instance after calling has/get on peering service manager
* remove method, only use protected variable

tests/ZendTest/ServiceManager/ServiceManagerTest.php
```diff
@@ -1189,4 +1189,53 @@ public function testIsSharedThrowsExceptionWhenPassedNameWhichDoesNotExistAnywhe
         $this->setExpectedException('Zend\ServiceManager\Exception\ServiceNotFoundException');
         $this->serviceManager->isShared('foobarbazbat');
     }
+
+    /**
+     * @covers Zend\ServiceManager\ServiceManager::loopPeeringServiceManagers
+     * @covers Zend\ServiceManager\ServiceManager::setServiceManagerCaller
```
weierophinney: These two statements are causing PHPUnit errors to occur; it looks like PHPUnit does not like `@covers` for non-public methods: https://travis-ci.org/zendframework/zf2/jobs/53314224#L486

I'll remove the statements during merge.

library/Zend/ServiceManager/ServiceManager.php
```diff
@@ -1002,10 +1011,33 @@ protected function retrieveFromPeeringManager($name)
...
+            // ignore peering service manager if they are the same instance
+            if ($caller !== $peeringServiceManager) {
+                // pass this instance to peering service manager
+                $peeringServiceManager->serviceManagerCaller = $this;
+                if ($peeringServiceManager->has($name)) {
+                    $this->shared[$name] = $peeringServiceManager->isShared($name);
+                    return $peeringServiceManager->get($name);
+                }
+                $peeringServiceManager->serviceManagerCaller = null;
             }
```
weierophinney: Let's tighten this up:

```php
if ($caller === $peeringServiceManager) {
    continue;
}
// ... promote the body of the conditional
```

I can do this during merge.

library/Zend/ServiceManager/ServiceManager.php
```diff
@@ -718,9 +723,16 @@ public function has($name, $checkAbstractFactories = true, $usePeeringServiceMan
...
+
+            $caller = $this->serviceManagerCaller;
             foreach ($this->peeringServiceManagers as $peeringServiceManager) {
-                if ($peeringServiceManager->has($name)) {
-                    return true;
+                // ignore peering service manager if they are the same instance
+                if ($caller !== $peeringServiceManager) {
+                    $peeringServiceManager->serviceManagerCaller = $this;
+                    if ($peeringServiceManager->has($name)) {
+                        return true;
```
weierophinney: Shouldn't `$peeringServiceManager->serviceManagerCaller` be reset to `null` before returning?

library/Zend/ServiceManager/ServiceManager.php
```diff
@@ -1002,10 +1011,33 @@ protected function retrieveFromPeeringManager($name)
...
-            if ($peeringServiceManager->has($name)) {
-                $this->shared[$name] = $peeringServiceManager->isShared($name);
-                return $peeringServiceManager->get($name);
+            // ignore peering service manager if they are the same instance
+            if ($caller !== $peeringServiceManager) {
+                // pass this instance to peering service manager
+                $peeringServiceManager->serviceManagerCaller = $this;
+                if ($peeringServiceManager->has($name)) {
+                    $this->shared[$name] = $peeringServiceManager->isShared($name);
+                    return $peeringServiceManager->get($name);
```
weierophinney: Shouldn't `$peeringServiceManager->serviceManagerCaller` be reset to `null` before returning?

weierophinney:

tests/ZendTest/ServiceManager/ServiceManagerTest.php
fabiang: Its ok to @cover non-public methods, but the given one didnt exists any more.

fabiang:
thank you, :+1:

PR #7299, opened 2015-03-05
Fix reflection for one line docblocks

lstrojny:
This fixes an issue where the parsing of single line docblocks fail and results in a syntax error with Ocramius/ProxyManager.

/cc @Ocramius

### DIFF
--- a/tests/ZendTest/Code/Reflection/TestAsset/TestSampleClass13.php
+++ b/tests/ZendTest/Code/Reflection/TestAsset/TestSampleClass13.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace ZendTest\Code\Reflection\TestAsset;
+
+/** Short Class Description */
+class TestSampleClass13
+{
+    /** Short Method Description */
+    public function doSomething()
+    {}
+
+    /** @return Something This describes something */
+    public function returnSomething()
+    {}
+}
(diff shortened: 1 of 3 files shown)


malukenho:
@lstrojny build is failing

tests/ZendTest/Code/Reflection/TestAsset/TestSampleClass13.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace ZendTest\Code\Reflection\TestAsset;
+
+/** Short Class Description */
+class TestSampleClass13
+{
+    /** Short Method Description */
+    public function doSomething()
+    {}
```
malukenho: You can put each curly braces on one line.

weierophinney: In this case, it makes more sense to omit this class from our php_cs rules.

weierophinney: Actually, that path is already excluded from our php-cs-fixer ruleset.

weierophinney:
@malukenho Note: please do not comment about build failures; we have status integration already with Travis, and authors will get notifications of build failures. Currently, *all* builds against master fail due to an environment issue anyways, so we need to check the results manually anyways to see if new, unknown failures are present.

tests/ZendTest/Code/Reflection/TestAsset/TestSampleClass13.php
Ocramius: This asset is actually supposed to have this particular structure to trigger a failure

malukenho:
@weierophinney Okay, not problem!

weierophinney:

PR #7301, opened 2015-03-08
added cache support to form annotation builder

carnage:
As title. I'm sure there was an issue raised in relation to caching the output of the form builder; but I can't find it.

### DIFF
--- a/library/Zend/Form/Annotation/AnnotationBuilder.php
+++ b/library/Zend/Form/Annotation/AnnotationBuilder.php
@@ -89,13 +89,16 @@ class AnnotationBuilder implements EventManagerAwareInterface, FormFactoryAwareI
      */
     protected $preserveDefinedOrder = false;
 
+    /**
+     * @param StorageInterface $cache
+     */
     public function setCache(StorageInterface $cache)
     {
         $this->cache = $cache;
     }
 
     /**
-     * @return \Zend\Cache\Storage\StorageInterface
+     * @return StorageInterface
      */
     public function getCache()
     {
--- a/library/Zend/Form/Annotation/AnnotationBuilder.php
+++ b/library/Zend/Form/Annotation/AnnotationBuilder.php
@@ -10,6 +10,7 @@
 namespace Zend\Form\Annotation;
 
 use ArrayObject;
+use Zend\Cache\Storage\StorageInterface;
 use Zend\Code\Annotation\AnnotationCollection;
 use Zend\Code\Annotation\AnnotationManager;
 use Zend\Code\Annotation\Parser;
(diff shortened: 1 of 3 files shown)


library/Zend/Form/Annotation/AnnotationBuilder.php
```diff
@@ -83,6 +89,19 @@ class AnnotationBuilder implements EventManagerAwareInterface, FormFactoryAwareI
      */
     protected $preserveDefinedOrder = false;
 
+    public function setCache(StorageInterface $cache)
+    {
+        $this->cache = $cache;
+    }
+
+    /**
+     * @return \Zend\Cache\Storage\StorageInterface
```
malukenho:
```php
/**
 * @return StorageInterface
 */
```

library/Zend/Form/Annotation/AnnotationBuilder.php
```diff
@@ -203,6 +222,17 @@ public function getFormSpecification($entity)
             }
         }
 
+        if (is_object($entity)) {
+            $cacheKey = 'FormAnnotationBuilder-' . get_class($entity);
+        } else {
+            $cacheKey = 'FormAnnotationBuilder-' . $entity;
+        }
```
malukenho: Maybe you can move it for another method, and you can test this one separated.

```php
$cacheKey = $this->generateKeyFor($entity)
```

carnage: I agree with this, I'll sort that out later

weierophinney:
Caching of the forms created from annotations should be handled at the application layer, and not as a feature of the annotation builder:

```php
$form = $cache->getItem('YourFormName', $success);
if (! $success) {
    // build form using annotation builder
    $cache->setItem('YourFormName', $form);
}
```

The feature is not needed within the annotation builder itself.

carnage:
I think you misunderstand what this is caching. It doesn't cache the forms themselves, but the intermediate form config array created by the expensive operation of parsing annoations. This representation is (should be) internal to the annotation builder. 

To access this to cache a form a user would have to call a different method on the builder cache that and then take care of creating the form using the factory themselves. Adding unnessisary steps just to get the same level of performance that building a form manually using array config would give them.

Caching intermediate forms of data structures is a pattern found embedded accross many frameworks which use annotations as meta data as the reading (using reflection) and parsing (regex's/string manipulation) is an expensive operation. Even PHP itself supports caching of its parsed form via APC - you wouldn't expect a user to take care of that in their applications would you?

Furthermore caching a form is something which is not going to work very well in all cases. Form elements, validators and filters are all fairly complex objects which could include uncacheable properties eg closures, db connections, request/user specific data.

I have been thinking about the structure of the parser in general as it is gaining too many responsibilties, however a refactor is not something that could be done for 2.4 and since 2.4 is supposed to be LTS, having caching support in is better than not having it at all because it polutes the architecture.

PR #7302, opened 2015-03-08
Hostname validator - TLD list update

Tauop:
Hi,

This is an update of TLD list in Hostname Validator.

BTW, I've made this little code to generate it.
Can we add it somewhere ?

Patrick

``` php
<?php
include 'init_autoloader.php';
use Zend\Dom\Query as DomQuery;
use Zend\Http\ClientStatic;

$response = ClientStatic::get('http://www.iana.org/domains/root/db');
if (!$response->isSuccess()) {
    echo "Error: can't get www.iana.org/domains/root/db".PHP_EOL;
    exit(1);
}

$dom = new DomQuery($response->getBody());
foreach ($dom->execute('span.domain.tld > a') as $node) {
    echo  "'".substr($node->nodeValue, 1)."',".PHP_EOL;
}
```


Tauop:
bad branch, sorry... :-1:

PR #7303, opened 2015-03-08
update TLD list

Tauop:
Hi,

This is an update of TLD list in Hostname Validator.

BTW, I've made this little code to generate it.
Can we add it somewhere ?

Patrick

``` php
<?php
include 'init_autoloader.php';
use Zend\Dom\Query as DomQuery;
use Zend\Http\ClientStatic;

$response = ClientStatic::get('http://www.iana.org/domains/root/db');
if (!$response->isSuccess()) {
    echo "Error: can't get www.iana.org/domains/root/db".PHP_EOL;
    exit(1);
}

$dom = new DomQuery($response->getBody());
foreach ($dom->execute('span.domain.tld > a') as $node) {
    echo  "'".substr($node->nodeValue, 1)."',".PHP_EOL;
}
```


sasezaki:
:+1: 

> Can we add it somewhere ?

I think `bin` dir is good appropriate place. please make another Pull Request.

moderndeveloperllc:
@Tauop Having it in `bin/` would be nice. I had thought about doing something like this, but your solution is better than what I could have done. <sigh> One of these days I'm actually gonna rewrite the validator to only work on the A names of domains (the punycode version) so that the file isn't a mismash of several different scripts and text directions. All the length restrictions, etc are on the A name not the composed U name anyways.

malukenho:
> I think bin dir is good appropriate place. please make another Pull Request

:+1:

weierophinney:
> BTW, I've made this little code to generate it.
> Can we add it somewhere ?

As others have suggested, I'd love to see this in the `bin/` dir; we can then run it for each new maintenance release. Please submit it as a separate pull request.

Tauop:
Thanks for all your feedback.
new PR => #7307 7307

PR #7306, opened 2015-03-10
Corrected docblock annotation type.

0xPaul:


weierophinney:
We specifically chose `mixed` as the type to allow objects that can serialize to string, as well as streams.

0xPaul:
OK, good job, but you failed to mention that in the docblock, and moreover, you specifically chose to specify *string* as the type for the field it's encapsulating on line 22. So if that's your story I strongly recommend you set the record straight.

PR #7307, opened 2015-03-11
Automatic TLDs update for Zend\Validator\Hostname

Tauop:
Hi,

I improve the update script provided in #7303 :
- in `bin/` directory
- update `$validTlds` definition in file `library/Zend/Validator/Hostname.php`
- several errors detection

usage :  `php ./bin/update_hostname_validator.php`

Is the script name ok ?

Patrick


moderndeveloperllc:
My only reservation is that his only does new TLDs, there is no mechanism if a gTLD is removed from the list. Might it be better to generate the entire list and simply replace $validTLDs? Of course if you want to spend the time, you could also look into the [Zend\Code\Generator](http://framework.zend.com/manual/current/en/modules/zend.code.generator.examples.html) classes and see of that might help.

moderndeveloperllc:
A ha! I found what I was looking for. The Root Zone Database is also available as a daily-generated text file, though only with A names - https://data.iana.org/TLD/tlds-alpha-by-domain.txt

Which would be best to pull from? The Validator class does have its home-grown `decodePunycode()` function that could be used if `idn_to_utf8()` was not available. It would help remove the dependance on checking for specific CSS to parse TLDs from an HTML file.

Tauop:
Hello,

> My only reservation is that this script only inserts new TLDs, there is no mechanism if a gTLD is 
> removed from the list. Might it be better to generate the entire list and simply replace $validTLDs? Of 
> course if you want to spend the time, you could also look into the Zend\Code\Generator classes and 
> see of that might help.

I don't make a diff.
I put the new list in place.

Zend\Code has so less document. Can't figure out how to use it.
Usage Zend\Code\Generator : don't know how to parse an existing class, and if the generated class will respect zf2 coding style... that's why I made a "replace" algorithme.

Patrick

Tauop:
> A ha! I found what I was looking for. The Root Zone Database is also available as a daily-generated 
> text file, though only with A names - https://data.iana.org/TLD/tlds-alpha-by-domain.txt

Very nice !

> Which would be best to pull from? The Validator class does have its home-grown decodePunycode()
> function that could be used if idn_to_utf8() was not available. It would help remove the dependance
> on checking for specific CSS to parse TLDs from an HTML file.

I will push a new PR using this method.

Patrick

Tauop:
@moderndeveloperllc 
I update the PR, with usage of ```https://data.iana.org/TLD/tlds-alpha-by-domain.txt```.
The good news is that we don't take into account :
- 'test' like ```.テスト```
- 'country-code not assigned' TLD like ```.bl```.

BTW, I don't understand why ```.oracle``` is present in 'text-version' of the database, but not in the 'html-version' :-/

Besides, I use ```Zend\Validator\Hostname::decodePunycode()``` is ```idn_to_utf8()``` is not present. However, I have to use reflexion to make this method accessible :-/

Patrick

moderndeveloperllc:
Sorry about the code generation misunderstanding. It was too late last night when I was trying to mentally parse your code! I remember when I was thinking about rewriting the validator last year I too noticed that the text list had more than the HTML list. The `.oracle` domain [definitely exists](https://www.icann.org/resources/agreement/oracle-2014-06-19-en), though I'm not sure anyone can actually buy one. It looks to be a vanity gTLD that's owned by Oracle.

I imagine that @weierophinney will want you to write a unit test for this.

Tauop:
@moderndeveloperllc 
I don't see what to unittest, as this is a script you run before a release, and you see modifications made by commit Hostname validator. BTW, where do I put unit test ?

Otherwise, I have break a testcase : 
```
1) ZendTest\Validator\HostnameTest::testAdditionalUTF8TLDs
The input appears to be a DNS hostname but cannot match TLD against known list
The input does not appear to be a valid URI hostname
The input does not appear to be a valid local network name - test123.δοκιμή
Failed asserting that false matches expected true.
/home/travis/build/zendframework/zf2/tests/ZendTest/Validator/HostnameTest.php:504
```

For me, ```test123.δοκιμή``` is not a valid Hostname, as ```δοκιμή``` is not listed on https://data.iana.org/TLD/tlds-alpha-by-domain.txt.
However, ```δοκιμή``` is listed on http://www.iana.org/domains/root/db as type = "test" and Sponsoring organisation = "Internet Assigned Numbers Authority".

What do you think ?
Is a 'test' TLD valid ??

Patrick

moderndeveloperllc:
I would say that is not a valid TLD. The [root domain database page](http://www.iana.org/domains/root/db/xn--jxalpdlp.html) for that TLD indicates, "This domain is not present in the root zone at this time."

Tauop:
FYI, The last Tracis failed build is not due to my PR
```
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
   1) ServiceManager.php (empty_return)
      ---------- begin diff ----------
      --- Original
      +++ New
      @@ @@
       
      -        return null;
      +        return;
           }
```

PR #7308, opened 2015-03-11
Allow ControllerManager to automatically add invokable classes

danizord:
I think the [ControllerManager#validatePlugin()](https://github.com/zendframework/zf2/blob/master/library/Zend/Mvc/Controller/ControllerManager.php#L101) is enough to make sure that it does not instantiate arbitrary classes as controllers, but only instances of `DispatchableInterface`. So that we can allow `ControllerManager` to automatically add invokable classes for convenience.

Also it comes in handy for beginners who do not understand the need to register the controller in so many places.

### DIFF
--- a/library/Zend/Mvc/Controller/ControllerManager.php
+++ b/library/Zend/Mvc/Controller/ControllerManager.php
@@ -26,13 +26,6 @@ use Zend\Stdlib\DispatchableInterface;
 class ControllerManager extends AbstractPluginManager
 {
     /**
-     * We do not want arbitrary classes instantiated as controllers.
-     *
-     * @var bool
-     */
-    protected $autoAddInvokableClass = false;
-
-    /**
      * Constructor
      *
      * After invoking parent constructor, add an initializer to inject the


malukenho:
:+1:

carnage:
I don't think this is a good idea, it will change assumptions that people may have regarding what can be dispatched.

I have an application (online game) which has a module for testing purposes which registers several controllers that allow cheating. In production the module and controllers are not loaded auto add invokable classes /could/ allow an attacker to dispatch these disabled controllers.

macnibblet:
No, this is a terrible idea, we already have enough magic :cry:

danizord:
@carnage, this is a routing problem, therefore it should be avoided by your routing configuration, not by your service locator.

If you are resolving the controller name from a route param, then you are exposing your application to attacks and unwanted behavior in anyway.

@macnibblet I can't see drawbacks in this approach, so the magic does not hurt there :)

carnage:
@danizord The default skelleton app config comes with this configuration. 

As for my app - there are other protections in place; I was simply giving an example. Also Defence in Depth.

gabrielsch:
:+1: 

IMO there's no need to register invokables if I'm going to use controller FQCN as controller name.

Ocramius:
> IMO there's no need to register invokables if I'm going to use controller FQCN as controller name.

It's related to a security issue fixed in `release-2.0.0-beta1`: no-go.

Ocramius:
Nuking this, sorry :-(

PR #7309, opened 2015-03-11
Hotfix/additional form options error case

carnage:
Reopened #7131 against develop

### DIFF
--- a/library/Zend/Form/Annotation/ElementAnnotationsListener.php
+++ b/library/Zend/Form/Annotation/ElementAnnotationsListener.php
@@ -163,6 +163,11 @@ class ElementAnnotationsListener extends AbstractAnnotationsListener
                 $specification['type'] = 'Zend\Form\Fieldset';
             }
 
+            if (isset($elementSpec['spec']['options'])) {
+                $specification['options'] = isset($specification['options']) ? $specification['options'] : array();
+                $specification['options'] = array_merge($elementSpec['spec']['options'], $specification['options']);
+            }
+
             // Add element spec:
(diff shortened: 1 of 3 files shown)


library/Zend/Form/Annotation/ElementAnnotationsListener.php
```diff
@@ -163,6 +163,11 @@ public function handleComposedObjectAnnotation($e)
                 $specification['type'] = 'Zend\Form\Fieldset';
             }
 
+            if (isset($elementSpec['spec']['options'])) {
+                $specification['options'] = isset($specification['options']) ? $specification['options'] : array();
+                $specification['options'] = array_merge($elementSpec['spec']['options'], $specification['options']);
```
malukenho: Can't be
```php
if (isset($elementSpec['spec']['options']) && isset($specification['options'])) {
    $specification['options'] = array_merge($elementSpec['spec']['options'], $specification['options']);
}
```

??

weierophinney: @malukenho Good point; I'll take care of that on merge.

weierophinney: Actually, @malukenho, it's correct as specified; we're trying to set the `$specification['options']` and/or update it; if the conditional is changed as you specify, `$specification['options']` will be empty later.

malukenho: Thanks for explanation @weierophinney :)

PR #7310, opened 2015-03-11
Callback Authentication Adapter

kanellov:
Hi,
I have implemented an authentication adapter using callback function. The main authentication logic is accomplished by the callback function. The callback must return the identity on success or false on failure.

Please take a look and tell me what you think about it!


kanellov:
Errors in unit testing

kanellov:
Fixed errors in unit tests

weierophinney:

PR #7313, opened 2015-03-11
Changed name (key) of validation message

Wilt:
Changed value of `INVALID` constant that is used as name for errors and key for validation messages from `stringLengthInvalid` to `stringTypeInvalid`.
This name is wrong since the error will occur on failed variable type validation [in the `is_string` method on line 189](https://github.com/zendframework/zf2/blob/master/library/Zend/Validator/StringLength.php#L189) and not on failed length validation.

### DIFF
--- a/library/Zend/Validator/StringLength.php
+++ b/library/Zend/Validator/StringLength.php
@@ -14,7 +14,7 @@ use Zend\Stdlib\StringWrapper\StringWrapperInterface as StringWrapper;
 
 class StringLength extends AbstractValidator
 {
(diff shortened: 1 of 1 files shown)


weierophinney:
Changing the value of a constant is a BC break, as it means that existing translation strings will no longer be matched (as the constant value is used as a key, and the constant is often not used within translation files, but rather the value it resolves to). Additionally, the constant values are, by convention, prefixed with the current validator name, and the change you're suggesting introduces an inconsistency.

PR #7314, opened 2015-03-11
add getter method for object prototype to Zend\Db\ResultSet\HydratingResultSet

k8ns:
add getter method for object prototype to Zend\Db\ResultSet\HydratingResultSet

It needs for Table which have injected TableGateway to create factory method that will return new Entity:
public function getNew()
{
    return clone $this->tableGateway->getResultSetPrototype()->getObjectPrototype();
}


weierophinney:
Sounds good; can you add a unit test, please? (You have until Thursday, 19 March 2015, if you want to get this into 2.4.)

sopinorg:
Unit test has been added

weierophinney:

PR #7315, opened 2015-03-12
array_filter compatibility

tklever:
Compatibility function for use in 5.3 - 5.5 systems lacking array_filter flags


tklever:
Travis reported a failure, but it looks like all the Stdlib tests ran file. Not sure what caused the failure.

PR #7320, opened 2015-03-12
Added a comprehensive html entity map in the Escaper

pouete:
Hello ZF2,
As you can see, my pull request is really simple, i just wanted to use characters with accents from the extended utf8 table in html entities, thus, i extended the htmlNamedEntityMap to most of the html entities.


Martin-P:
You can use characters like ë just as they are. Can you explain why you think this is needed?

padraic:
This is a security component. The characters you reference are not a security concern and should not be allowed to litter the class and obscure its purpose. It would also break XML serialisations of HTML 5 (as rarely as they are used).

If you need to represent certain international characters as entities, then you should do so via your own view helpers (prior to Escaper being invoked) or address the underlying problem that requires you to use entities in order to represent these characters in output. This looks a lot like the anglocentric kludge of using htmlentities() instead of htmlspecialchars() to cover over a lack of UTF-8 consistency (which is NOT a security issue and so requires no fixing).

pouete:
I understand the security issue here and why you want to keep this class tidy.
To anwer to Martin-p i have a form and uses a "pattern" in a addition to a "oninvalid" attribute. 
i would like to translate the oninvalid message that pops, so i injected the translator and translate it, but the security feature fires me an error. 
Also, do you have any doc/readings that explain how a utf 8 character could be used for security abuse ?

Martin-P:
> i would like to translate the oninvalid message that pops, so i injected the translator and translate it, but the security feature fires me an error.

I don't know what the error is you are seeing, but looking at your PR translating things like ë should not be any problem at all. My guess is you are doing something wrong in your encoding.

PR #7326, opened 2015-03-14
added doc-block @ Zend\InputFilter\InputInterface

kokspflanze:
added doc-block @ Zend\InputFilter\InputInterface for better code completion


library/Zend/InputFilter/InputInterface.php
```diff
@@ -14,26 +14,121 @@
 
 interface InputInterface
 {
+    /**
+     * @param bool $allowEmpty
+     *
```
Martin-P: Empty line is not needed here and can be removed

library/Zend/InputFilter/InputInterface.php
```diff
@@ -14,26 +14,121 @@
...
+     * @return InputInterface
```
Martin-P: When the object itself is returned `@return self` is used (same for all other docblock changes in this PR).

PR #7327, opened 2015-03-14
Rbac callback assertion

kanellov:
Hello,

I've implemented an Rbac Assertion Class that uses a callback to do the assertion. 

Here is an example:

``` php
<?php
use Zend\Permissions\Rbac\Rbac;
use Zend\Permissions\Rbac\Assertion\Callback;

// User is assigned the foo role with id 5
// News article belongs to userId 5
// Jazz article belongs to userId 6

$rbac = new Rbac();
$user = $mySessionObject->getUser();
$news = $articleService->getArticle(5);
$jazz = $articleService->getArticle(6);

$rbac->addRole($user->getRole());
$rbac->getRole($user->getRole())->addPermission('edit.article');

$assertionCb = function ($user, $article) {
    return function ($rbac) use ($user, $article) {
        return $user->getId() == $article->getUserId();
    };
};

// true always - bad!
if ($rbac->isGranted($user->getRole(), 'edit.article')) {
    // hacks another user's article
}

$assertion = new Callback($assertionCb($user, $news));

// true for user id 5, because he belongs to write group and user id matches
if ($rbac->isGranted($user->getRole(), 'edit.article', $assertion)) {
    // edits his own article
}

$assertion = new Callback($assertionCb($user, $jazz));

// false for user id 5
if ($rbac->isGranted($user->getRole(), 'edit.article', $assertion)) {
    // can not edit another user's article
}
```

I borrowed the example from Zend Framework manual and from [Dynamic Assertions](http://framework.zend.com/manual/current/en/modules/zend.permissions.rbac.examples.html#dynamic-assertions) section and I altered it to make use of the proposed Assertion Class.

PR #7328, opened 2015-03-14
Acl callback assertion

kanellov:
Like the Rbac callback assertion proposed in [#7327](https://github.com/zendframework/zf2/pull/7327), I've implemented an **ACL Callback assertion class**. Here is an example of use:

``` php
<?php
use Zend\Permissions\Acl\Acl;
use Zend\Permissions\Acl\Assertion\Callback;

$validIps = array(
    10.10.10.10,
);
$acl = new Acl();
$assertion = new Callback(function ($acl, $role, $resource, $privilege) use ($validIps) {
    return in_array($_SERVER['REMOTE_ADDR'], $validIps);
});
$acl->allow(null, null, null, $assertion);  
```


weierophinney:

PR #7336, opened 2015-03-18
Allowing the route match controller to override the controller class

tklever:
Updates to InjectTemplateListener to allow for the RouteMatch "controller" parameter to override the controller's class.

Use case is to handle controllers that are built by abstract factories, which require extra rendering logic because this resolver will always use the get_class() method over what's in the RouteMatch (oftentimes a base or parent class). 

Also helpful with the reverse, which is extended or decorator controllers appearing (and rendering) as their parent controller.

The behavior is off by default for backwards compatibility (1 test broke, so close).

### DIFF
--- a/library/Zend/Mvc/View/Http/InjectTemplateListener.php
+++ b/library/Zend/Mvc/View/Http/InjectTemplateListener.php
@@ -33,6 +33,13 @@ class InjectTemplateListener extends AbstractListenerAggregate
     protected $controllerMap = array();
 
     /**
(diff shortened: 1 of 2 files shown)


tklever:
Travis reporting what I believe are false positives. 5.3 and 5.6 failed, but the Mvc test block looks clean

weierophinney:
I like it — clean implementation, and fully backwards compatible (because it's opt-in)! Merging for 2.4.

PR #7287, opened 2015-03-03
AbstractOptions isset issue, #7286

rms230:
When calling isset() on an object variable that is a class extending AbstractOptions an exception is thrown.

An example useage of this is found in the EventManagerAwareTrait where the trait checks a class variable isset, if the class that is using the EventManagerAwareTrait extends AbstractOptions an exception is thrown.

This pull request fixes AbstractOptions by checking that the method exists and that the value is not equal to null, thus stopping the exception being thrown.

This pull request also fixes issue #7286

Issue #7286, opened 2015-03-03
disabled checkbox


GhTigran:
When adding 'disabled' => true to Form Checkbox element attributes, in view file disabled is added to input checkbox element only, but it's not added to the hidden element as well, which will cause unintentional submittion hidden value.

Solution: When both use_hidden_element and disabled are set to true on checkbox form element. Add disabled attribute to both input checkbox and associated hidden element while drawing the form.

Martin-P:
> Solution: When both use_hidden_element and disabled are set to true on checkbox form element. Add disabled attribute to both input checkbox and associated hidden element while drawing the form.

Although I understand what you are saying, adding a attribute `disabled` to the hidden input does fix anything. Anyone who knows how to change the value of a hidden input can also remove the `disabled` attribute. You should never trust user input if you want a safe application.

froschdesign:
> Add disabled attribute to both input checkbox and associated hidden element while drawing the form.

Do not add the hidden element.

Martin-P:
@froschdesign :+1: I did not even think of that :smile:

Martin-P:
@froschdesign On second thought, that also does not fix the issue. The application itself should know when to ignore elements and not by (lack of) user input :wink: 

It seems this already covered in [`Zend\Form\Fieldset::bindValues()`](https://github.com/zendframework/zf2/blob/ff04baf16fdd958ff0eba26143de3ef014dc4488/library/Zend/Form/Fieldset.php#L588-L593).

@GhTigran Can you provide a unittest where data is *not* ignored when a checkbox is disabled?

PR #7341, opened 2015-03-18
Set disabled attribute also on checkbox hidden element

Martin-P:
Fix for #7286. This mainly is a split up of #7287, but with a correct unittest.


weierophinney:
Merged to develop for release with 2.4. Thanks, @Martin-P !

PR #7342, opened 2015-03-18
Do not throw exception when isset is used on AbstractOptions existing object variable

Martin-P:
Same fix as suggested by @rms230 in PR #7287. Split it up and slightly changed the assertion for `BadMethodCallException`.

Original description of #7287:

> When calling isset() on an object variable that is a class extending AbstractOptions an exception is thrown.
> 
> An example useage of this is found in the EventManagerAwareTrait where the trait checks a class variable isset, if the class that is using the EventManagerAwareTrait extends AbstractOptions an exception is thrown.
> 
> This pull request fixes AbstractOptions by checking that the method exists and that the value is not equal to null, thus stopping the exception being thrown.

### DIFF
--- a/library/Zend/Stdlib/AbstractOptions.php
+++ b/library/Zend/Stdlib/AbstractOptions.php
@@ -147,7 +147,9 @@ abstract class AbstractOptions implements ParameterObjectInterface
      */
     public function __isset($key)
     {
-        return null !== $this->__get($key);
+        $getter = 'get' . str_replace('_', '', $key);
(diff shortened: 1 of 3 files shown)


Martin-P:
Ping @weierophinney
Just making sure this part of the original PR #7287 is included in the 2.4.0 release too :smiley:

weierophinney:
Pushed to develop for release with 2.4; thanks, @Martin-P !

Issue #7154, opened 2015-01-25
Zend\Db\Adapter\Driver\Pdo\Result Not using \PDO::FETCH_ASSOC in current() method


itrascastro:
public function getById($id)
    {
        $stm = $this->db->createStatement('SELECT * FROM User WHERE id = ?');
        $rowset = $stm->execute(array($id));
        $row = $rowset->current();
    }

current() is using FETCH_ARRAY instead of FETCH_ASSOC as it is supposed to be

Ocramius:
Did you check the test suite? Maybe that's the expected behavior...

itrascastro:
Can I force it to be \PDO::FETCH_ASSOC in current() call or elsewhere?

Ocramius:
Marking as `awaiting author updates`: please check the test suite first.

Martin-P:
`\PDO::FETCH_ASSOC` is hardcoded in [`Zend\Db\Adapter\Driver\Pdo\Result` line 119](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/library/Zend/Db/Adapter/Driver/Pdo/Result.php#L119):
```php
$this->currentData = $this->resource->fetch(\PDO::FETCH_ASSOC);
```
It should be impossible not to use `\PDO::FETCH_ASSOC`.

@Ocramius:
The test suite only contains a single test for `current()`, that's all: [ZendTest\Db\Adapter\Driver\Pdo\ResultTest](https://github.com/zendframework/zf2/blob/67f098af070b29d5042e89e936604df3193d2212/tests/ZendTest/Db/Adapter/Driver/Pdo/ResultTest.php)

itrascastro:
That is why I am surprised because it is hardcoded. But it happens to me with the code I posted, it seems that it is using \PDO::FETCH_ARRAY. I cannot find a reason for that. I have attached an screenshot:

![current-return-array](https://cloud.githubusercontent.com/assets/3638437/5905160/57eae9b4-a58e-11e4-8a0f-5e2675b79e84.png)

Martin-P:
Your screenshot shows exactly what `\PDO::FETCH_ASSOC` should do. It shows an array with 5 named elements:
* `$row['id']` has value `5`
* `$row['email']` has value `user5@email.com`
* `$row['password']` has value `1234`
* etc.

What do you think `var_dump($row);` should display?

itrascastro:
[...]

PR #7343, opened 2015-03-18
Pdo fetch mode support

yp28:
Issue #7154 

Any feedback on the PR would be much appreciated.


weierophinney:
Looks good, @YP28 !

Issue #7192, opened 2015-02-06
`Form/Fieldset` dependency on `FormInterface`


claytondaley:
I'm trying to build an extensible HTML table generator and want to provide HTML table builders a comparable experience -- infrastructure and syntax -- to form construction.  Obviously, I don't need a bunch of the form infrastructure (input filters or validation) so I expected to extend `Fieldset` instead of `Form`.

As I worked my way through this possibility, I discovered that the type hint for `Fieldset->prepareElement()` is actually `FormInterface` per an inheritance from `ElementPrepareAwareInterface`.  Obviously, any class that extends `Fieldset` will not meet this requirement.  I'm trying to find out if this dependency is really obligatory or if you'd consider a BC-incompatible reduction to `FieldsetInterface`.

In favor of reducing this requirement to `FieldsetInterface`:

 - `FieldsetInterface` is the interface that actually extends `ElementPrepareAwareInterface`
 - `FieldsetInterface` is inherited by all forms so this would be transparent to most users
 - Unless I missed something, none of the `prepareElement` implementations in `Zend/Form` depend on methods in `Form` but not in the `Fieldset`.
 - Makes `Fieldset` usable in isolation.  The only other "Form" dependency is the `Form/Factory` which is really a common factory -- and appears to be so out of convenience not obligation.

In favor of staying at `FormInterface`:

 - BC break
 - Custom code may require a method from the `FormInterface` in their `PrepareElement` calls.
 - Form and Fieldset share a common Factory so they'd still be "coupled".  As mentioned above, this appears to be a matter of convenience not coupling.

claytondaley:
Incidentally, `get/setFormFactory` has similar (possibly worse) issues.

 - The factory needs to be available in `Fieldset` because it supports [factory-backed extension](https://zf2.readthedocs.org/en/latest/in-depth-guide/zend-form-zend-form-fieldset.html#creating-your-first-fieldset) to the fieldset.
 - The methods is defined in `FromFactoryAwareInterface` which is included in `FieldsetInterface` and used in `Fieldset`
 - However, the hint for `@returns` is `Form`.

Presumably, the right hint is `FieldsetInterface`.  At minimum, a change to `FormInterface` would be consistent with a previously accepted change -- #3526.

In the long run (maybe ZF3), it'd be better to clean up names (~~Form~~FactoryAwareInterface) and split (but extend) the factories so they're easier to inherit.

claytondaley:
cloded by #7344

[...]

PR #7344, opened 2015-03-19
Remove `FormInterface` dependency from `Fieldset`

claytondaley:
This eliminates the dependency of  `Fieldset` on `FormInterface` so it can be extended (see #7192 for background).

### DIFF
--- a/library/Zend/Form/Element/Collection.php
+++ b/library/Zend/Form/Element/Collection.php
@@ -15,7 +15,6 @@ use Zend\Form\ElementInterface;
 use Zend\Form\Exception;
 use Zend\Form\Fieldset;
 use Zend\Form\FieldsetInterface;
-use Zend\Form\FormInterface;
 use Zend\Stdlib\ArrayUtils;
 
 class Collection extends Fieldset
@@ -464,10 +463,10 @@ class Collection extends Fieldset
     /**
      * Prepare the collection by adding a dummy template element if the user want one
      *
-     * @param  FormInterface $form
+     * @param  FieldsetInterface $form
      * @return mixed|void
      */
-    public function prepareElement(FormInterface $form)
+    public function prepareElement(FieldsetInterface $form)
     {
         if (true === $this->shouldCreateChildrenOnPrepareElement) {
             if ($this->targetElement !== null && $this->count > 0) {
--- a/library/Zend/Form/Element/Csrf.php
+++ b/library/Zend/Form/Element/Csrf.php
@@ -11,7 +11,7 @@ namespace Zend\Form\Element;
 
 use Zend\Form\Element;
 use Zend\Form\ElementPrepareAwareInterface;
-use Zend\Form\FormInterface;
+use Zend\Form\FieldsetInterface;
 use Zend\InputFilter\InputProviderInterface;
 use Zend\Validator\Csrf as CsrfValidator;
 
@@ -149,7 +149,7 @@ class Csrf extends Element implements InputProviderInterface, ElementPrepareAwar
     /**
      * Prepare the form element
      */
-    public function prepareElement(FormInterface $form)
+    public function prepareElement(FieldsetInterface $form)
     {
         $this->getCsrfValidator()->getHash(true);
     }
--- a/library/Zend/Form/Element/DateSelect.php
+++ b/library/Zend/Form/Element/DateSelect.php
@@ -11,7 +11,7 @@ namespace Zend\Form\Element;
 
 use DateTime as PhpDateTime;
 use Zend\Form\Exception\InvalidArgumentException;
-use Zend\Form\FormInterface;
+use Zend\Form\FieldsetInterface;
 use Zend\Validator\ValidatorInterface;
 use Zend\Validator\Date as DateValidator;
 use Exception;
@@ -132,10 +132,10 @@ class DateSelect extends MonthSelect
     /**
      * Prepare the form element (mostly used for rendering purposes)
      *
-     * @param  FormInterface $form
+     * @param  FieldsetInterface $form
      * @return mixed
      */
-    public function prepareElement(FormInterface $form)
+    public function prepareElement(FieldsetInterface $form)
     {
         parent::prepareElement($form);
 
--- a/library/Zend/Form/Element/DateTimeSelect.php
+++ b/library/Zend/Form/Element/DateTimeSelect.php
@@ -11,7 +11,7 @@ namespace Zend\Form\Element;
 
 use DateTime as PhpDateTime;
 use Exception;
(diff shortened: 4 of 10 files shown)


weierophinney:

adamlundrigan:
In one of my modules ([LdcZendFormCTI](https://github.com/adamlundrigan/LdcZendFormCTI)) I extend the Collection class and override the [`prepareElement` method](https://github.com/adamlundrigan/LdcZendFormCTI/blob/master/src/Form/Element/NonuniformCollection.php#L115).  With this change I can't support both ZF 2.3.x and 2.4.x as the method signature of `prepareElement` has changed.  

Travis-CI failure w/ error message: https://travis-ci.org/adamlundrigan/LdcZendFormCTI/jobs/55336753#L243

claytondaley:
Yeah.  `FormInterface` is a superset of `FieldsetInterface` so this is [LSP](http://en.wikipedia.org/wiki/Liskov_substitution_principle) valid but PHP invalid (as of [5.4](http://php.net/manual/en/language.oop5.abstract.php)).

What do you recommend? I can introduce a different method (e.g. `prepareElements`, appending 's', with the revised signature and we can deprecate the other.

weierophinney:
@adamlundrigan I'm just now seeing your comment. I think the BC break is likely a good one, but I'd like to know the extent of the breakage, and how difficult it would be to fix. Could you drop me an email, please?

adamlundrigan:
@weierophinney I'll check it out in more detail tonight and drop you an email

weierophinney:
After discussion with @adamlundrigan , I'm reverting this patch for 2.4. While the change is easily accommodated internal to the framework, it breaks any userland extensions, as they need to update their typehints to work with post-patch code; additionally, there may be potential unintended breakages with regard to internal classes (e.g., file elements currently pull their encoding from the passed object; if that object is now a fieldset, that functionality breaks, as encoding is not part of the fieldset interface).

We can reconsider a PR such as this one for v3.

claytondaley:
The typehint issue is a serious problem.  As I mentioned above, this can be resolved by adding a second method with the new type hint -- that calls the current method internally, by default.  I'm happy to adjust the PR accordingly.

Can you provide some more details on your concerns around `File`?  It's frustrating to be excluded from the discussion that give rise to this concern -- all the more so since I can't figure out why this is perceived as a problem.  Here's a plausible sequence that continues to work:

 - I create a class `NonZfForm` that inherits from `Fieldset`
 - Prepare is called using an instance of the root element, in this case a `NonZfForm`
 - As a fieldset, `NonZfForm` has the interface (`setAttribute`) so the `Form\File` call `$fieldset->setAttribute('enctype', 'multipart/form-data');` works fine.
 - If I render `NonZfForm` as a `<form>` and populate the attributes from the `getAttributes` array, the `enctype` will appear in the right place.

If I'm not actually rendering a form (likely since I'm not inheriting from `Form`), the attribute is vestigial, but unlikely to be fatal in any way.

adamlundrigan:
Apologies, I should have cross-posted my message to Matthew here when I sent it:

> I've given it some thought and, personally, I don't think this BC break is a good one at this point in the ZF2 lifecycle.

> Any custom form/fieldset/element that has implemented prepareElement will now emit a fatal error.  The fix is very straightforward for apps built on the framework -- just change the typehint -- but for libraries bundling custom form elements it creates a split, since the same class now can't support both <=2.3 and 2.4.  Those libraries would have to implement an autoloader hack so that different class definitions are loaded for <=2.3 and 2.4.  
>
> The patch could also allow some unintended behaviour, for example the File element uses prepareElement to set the correct enctype on the form.  if a fieldset is passed to prepareElement the enctype is set on the fieldset and not on the form.
>
> https://github.com/zendframework/zf2/blob/9908c0bf11af8441f18624f1e3b1449cdf0d3d6d/library/Zend/Form/Element/File.php#L37
>
> In my view that's a lot of pain for the improvement this brings: fixing an overly-restrictive typehint (unless there's more to the patch that I'm missing)

Regarding the File example, this PR doesn't cause the issue described because the current prepareElement behaviour (passing in the Form instance) hasn't changed.  What if that were to change and the form preparation passed in the element's immediate parent fieldset rather than the Form instance   ?  The typehint now allows this, and results in a change in behaviour as the enctype is no longer being set on the Form but on whatever fieldset is passed to prepareElement.

I don't really know if that's an issue to worry about...the code may have safeguards against that which I have missed, but the typehint change makes it a possiblity whereas previously it wasn't.

claytondaley:
Thanks for providing the background.  While I isolated this particular fix in its own PR,

 - I've already implemented a complete [StaticTable implementation](https://github.com/claytondaley/ZendTable) that depends this change.  Basically, the columns in the table are defined by a fieldset (and each can use a `HydratorStrategy` to format its cells).
 - I've implemented the `Table` code in an upgraded `ZfcUserAdmin` [branch](https://github.com/claytondaley/ZfcUserAdmin/tree/table)

I'm already using both internally, but I intend to submit them for consideration after some more cleanup (including tests).  Of course, they depend on the typehint change so I'd have to leave improvements out of PRs until it gets resolved inside ZF.

===

In retrospect, the easiest fix for hybrid support would be a second chain that uses a different name.  For example, 2.3 calls `prepareElement` (with the old typehint) and 2.4 calls `prepareElements` (with the trailing `s` and new typehint).  Anyone wishing to support both versions can place the same code under both names (tree traversal issues make my original proposal... complex).

claytondaley:
The more I analyze the `File` issue, the harrier it gets.  The presence of a `File` element shouldn't say anything about how it's to be implemented.  What if my view helper for `File` asked for a URL and my InputFilter actually downloaded it?  What if I had a JS uploader that responded with the ID for my file?  All seem valid, but the `prepareElement` code in `File` assumes a very specific rendering strategy (which seems to violate Separation of Concerns).
===
On top of this, the [implementation](https://github.com/zendframework/zf2/blob/master/library/Zend/Form/Form.php#L229) of `prepareElement` inside `Form` currently passes on its `$form` parameter to child objects.

 - If I nested a `Form` inside a `Form`, the top-level `Form` would get the `enctype` attribute rather than the child form that actually holds the element.
 - It's also valid to call `$form->prepareElement(new Form())` which would have a similarly problematic result.

We could call this a bug in `Form`, but it's presently wrong for `File` to assume that the `$form` parameter is the right place for this attribute seems flimsy and depends on (sensible but technically arbitrary) conventions of use.
===
If we take the Separation of Concerns point seriously, is there a place to move this logic?  If not, it may be necessary to deem Form a closed system and fix `Form::prepareElement` so that it always uses `$this`.  That would make the implicit assumption inside `File` explicit (and therefore safe).

PR #7345, opened 2015-03-19
use the target host and not the proxy host in the client hello request

lsmith77:
see also https://github.com/zendframework/zf1/pull/530


samsonasik:
ping @weierophinney

weierophinney:

PR #7346, opened 2015-03-19
Add missing return statement.

ffraenz:


fabiocarneiro:
ohgodwhy?

mwillbanks:
Please supply a unit test so that this does not happen in the future.

weierophinney:
Um… it's only missing if you expect a fluent interface. Unfortunately, it's documented that way, so I can see why the expectation is there. :)

As @mwillbanks notes, please add a unit test so that we know the fluent contract exists and is expected.

ffraenz:
This is the first time I am dealing with unit tests, so bear with me :)

fabiocarneiro:
@ffraenz You could use assertSame instead of assertEquals, that will assert if the return is the same instance and not another instance of the same type.

 Also i didn't see the point of asserting setFieldType in the same test. You could at least split it and make the test name more explicit, like ReturnsSameRendererInstanceWhenResolverIsSet and ReturnsSameRendererInstanceWhenFieldTypeIsSet

ffraenz:
@fabiocarneiro This makes perfect sense to me, thank's!

Xerkus:
@weierophinney may be it will be better NOT to introduce fluent interface? It is kinda bad and that is not BC break since it never was there.

fabiocarneiro:
@Xerkus I agree with you, but some ppl consider docblocks change as bc break...

tests/ZendTest/View/Renderer/FeedRendererTest.php
```diff
@@ -118,4 +118,16 @@ public function testSettingUnacceptableFeedTypeRaisesException()
         $this->setExpectedException('Zend\View\Exception\InvalidArgumentException', 'expects a string of either "rss" or "atom"');
         $this->renderer->setFeedType('foobar');
     }
+    
+    public function testReturnsSameRendererInstanceWhenResolverIsSet()
+    {
+        $returnValue = $this->renderer->setResolver(null);
```
mwillbanks: You will need to actually set a resolver here.

weierophinney:
@Xerkus I'd agree with you, except that this setter would then be inconsistent with the rest of the class, which _does_ implement a fluent interface on existing setters. As such, I'd consider this a necessary fix.

PR #7348, opened 2015-03-19
fix docblock

yp28:


weierophinney:
Closing; since tests are excluded from the API docs, and the intention of the test can be determined in the test body, I don't think we need the change.

PR #7359, opened 2015-03-23
Missing return statement in PHPDoc in AbstractNavigationFactory

ASvi:
Missing return statement in PHPDoc in AbstractNavigationFactory preparePages()


weierophinney:
In looking at this, the proper fix would be:

- Document as `null|array`
- Update the `injectComponents()` docblock's return value as `array`

as those are the actual return values. I can make that change on merge.

PR #7332, opened 2015-03-23
Zend\Http\Client sends content-type: application/x-www-form-urlencoded with GET request

weierophinney:
We have an api connection with Zend\Http\Client.

If we do a PUT request and with the same connection we do a GET 

```
    $request = new Request();
    $request->setMethod(Request::METHOD_GET);
    $request->setUri($url);

    $headers = $request->getHeaders();
    $headers->addHeaderLine('Accept: application/sellside.ad.list-v2+json, application/json');
    $headers->addHeaderLine('Authorization: Bearer ' . $this->getAccessToken());

    $response = $this->httpClient->send($request);
```

the Client class keeps the encType and sets the content type in the header to   'Content-Type' =>  "application/x-www-form-urlencoded" A GET shouldn't have this type. The provider of the API reports a HTTP 415 because of this issue.

We managed to solve this issue with Client->reset() but I think this isn't the solution to this.

This content-type is set at line 1150 in Zend\Http\Client.php

Thanks

### DIFF
--- a/library/Zend/Http/Client.php
+++ b/library/Zend/Http/Client.php
@@ -412,13 +412,16 @@ class Client implements Stdlib\DispatchableInterface
      */
     public function setEncType($encType, $boundary = null)
     {
-        if (!empty($encType)) {
-            if (!empty($boundary)) {
-                $this->encType = $encType . "; boundary={$boundary}";
-            } else {
-                $this->encType = $encType;
-            }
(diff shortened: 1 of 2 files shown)


weierophinney:
Build passes on my branch: https://travis-ci.org/weierophinney/zf2/builds/55533126

PR #7295, opened 2015-03-23
Zend\Http\Header\GenericHeader introduces BC break

weierophinney:
V2.3.4 (and possibly earlier) has a BC break to version 2.2.3 (and possibly later)
Links to: https://github.com/zendframework/zf2/pull/5301

This is caused by the insistence that underscores are converted to hyphens at L92.  It is debatable as to whether RFC2616 actually insists that underscores are not allowed.

Irrespective of that, the code 
- should not change header names under the covers
- should throw an exception as it does a few lines later for non RFC compliance, which would at least warn developers that something is wrong, rather than trying to figure out why their code suddenly breaks when a service endpoint is hit, and is consistent with the exception throwing later.

In addition, it would be useful to include a NonRfc2616Header that can be used in place of GenericHeader with appropriate documentation in the reference guide.

### DIFF
--- a/library/Zend/Http/Header/GenericHeader.php
+++ b/library/Zend/Http/Header/GenericHeader.php
@@ -88,18 +88,16 @@ class GenericHeader implements HeaderInterface
             throw new Exception\InvalidArgumentException('Header name must be a string');
         }
 
-        // Pre-filter to normalize valid characters, change underscore to dash
-        $fieldName = str_replace('_', '-', $fieldName);
-
         /*
-         * Following RFC 2616 section 4.2
+         * Following RFC 7230 section 3.2
          *
-         * message-header = field-name ":" [ field-value ]
-         * field-name     = token
-         *
-         * @see http://tools.ietf.org/html/rfc2616#section-2.2 for token definition.
+         * header-field = field-name ":" [ field-value ]
+         * field-name   = token
+         * token        = 1*tchar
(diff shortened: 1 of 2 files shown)


weierophinney:
Tests passed against my branch: https://travis-ci.org/weierophinney/zf2/builds/55541093

library/Zend/Http/Header/GenericHeader.php
```diff
@@ -88,18 +88,16 @@ public function setFieldName($fieldName)
...
+         * header-field = field-name ":" [ field-value ]
+         * field-name   = token
+         * token        = 1*tchar
+         * tchar        = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
+         *                "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
          */
-        if (!preg_match('/^[!#-\'*+\-\.0-9A-Z\^-z|~]+$/', $fieldName)) {
+        if (!preg_match('/^[!#$%&\'*+\-\.\^_`|~0-9a-zA-Z]+$/', $fieldName)) {
             throw new Exception\InvalidArgumentException(
                 'Header name must be a valid RFC 2616 (section 4.2) field-name.'
```
Freeaqingme: Given how RFC 2616 is superceded by RFC 7230 (and various others), maybe we should also just update the exception message as well?

weierophinney: Good point; I'll change it now.

PR #7311, opened 2015-03-23
Using TableGateway::insert() fails when table has an alias

weierophinney:
I want to use a single table `Users` via **TableGateway**. I need to put an alias on a table to be able to use it in `select` query. I'm using a **PDO** driver with **MySQL** platform.

When you create a new **TableGateway**:

```
new \Zend\Db\TableGateway\TableGateway(array('U' => 'Users'), $adapter);
```

you can use this instance to `select` whatever you like. However, when you try to `insert` anything using that instance, `Zend\Db\Sql\Insert` tries to quote table identifier (using adapter's platform) and it fails on

```
Zend\Db\Adapter\Platform\Mysql:
public function quoteIdentifier($identifier)
    {
        return '`' . str_replace('`', '``', $identifier) . '`';
    }
```

as `$identifier` is an array of `array('U' => 'Users')` in that case.

I understand there are no table aliases in SQL insert of any flavour, still, I believe this is a bug as `Zend\Db\Sql\Insert` is missing something similar to what `Zend\Db\Sql\Select` does:

```
if (is_array($table)) {
   $alias = key($this->table);
   $table = current($this->table);
} 
```

Also, it happens with `TableIdentifier` as well if you add table with alias to it. As `TableGateway` honours table notation with alias, I'd expect to still be able to do any CRUD actions on it.

### DIFF
--- a/library/Zend/Db/TableGateway/AbstractTableGateway.php
+++ b/library/Zend/Db/TableGateway/AbstractTableGateway.php
@@ -216,8 +216,13 @@ abstract class AbstractTableGateway implements TableGatewayInterface
     protected function executeSelect(Select $select)
     {
         $selectState = $select->getRawState();
-        if ($selectState['table'] != $this->table && (is_array($selectState['table']) && end($selectState['table']) != $this->table)) {
-            throw new Exception\RuntimeException('The table name of the provided select object must match that of the table');
+        if ($selectState['table'] != $this->table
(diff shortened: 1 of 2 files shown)


weierophinney:
Build on my own branch passes: https://travis-ci.org/weierophinney/zf2/builds/55558431

mirfilip:
Related issue #7430.

PR #7361, opened 2015-03-24
Missing parameter annotation in PHPDoc in Zend\Http\PhpEnvironment\Request

ASvi:
Missing parameter annotation in PHPDoc in Zend\Http\PhpEnvironment\Request


weierophinney:
I've updated the description to be more accurate: this is not a return annotation, but a parameter annotation.

PR #7364, opened 2015-03-24
Switch to PSR-4

afurculita:
This PR changes all the composer.json from all the library components to handle the autoload using PSR4

### DIFF
--- a/library/Zend/Authentication/composer.json
+++ b/library/Zend/Authentication/composer.json
@@ -8,11 +8,10 @@
     ],
     "homepage": "https://github.com/zendframework/zf2",
     "autoload": {
-        "psr-0": {
+        "psr-4": {
             "Zend\\Authentication\\": ""
         }
     },
-    "target-dir": "Zend/Authentication",
     "require": {
         "php": ">=5.3.23",
         "zendframework/zend-stdlib": "self.version"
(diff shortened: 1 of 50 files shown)


Martin-P:
PSR-4 was proposed in PR #6357.

afurculita:
This PR refers to all the components, not the full stack. Also as Jordi Boggiano specifies in her article http://seld.be/notes/psr-4-autoloading-support-in-composer, the "target-dir" is deprecated, it was just a hack and it recommends PSR-4

Ocramius:
This requires a change of how the submodules are split: may be a bit more complicated than "just doing it"

gianarb:
In my opinion this "transversal" commit between components create a lot of history problem :)
PSR-0 is deprecated and this is an important improvement but maybe we could apply it after split..

Ocramius:
@gianarb this is just the `composer.json`: no directory structure changes, as far as I can see

weierophinney:
I tested this just now by cloning the `Component_ZendVersion` repo, making the changes as noted in this PR, and installing; worked perfectly.

weierophinney:
I'm also making this change to the project `composer.json`:

```json
"psr-4": {
    "Zend\\": "library/Zend/"
}
```

I'm leaving the tests as PSR-0, as we have some test assets with underscores in class names, which PSR-4 does not support; we'll update those when we split the components out into their own repos.

Thanks for the patch, @afurculita !

PR #7365, opened 2015-03-24
Tiny patch

pchapl:
Just have no idea for what here used slow function to switch between two constants. Is there deep meaning?

### DIFF
--- a/library/Zend/Db/ResultSet/ResultSet.php
+++ b/library/Zend/Db/ResultSet/ResultSet.php
@@ -46,7 +46,8 @@ class ResultSet extends AbstractResultSet
      */
     public function __construct($returnType = self::TYPE_ARRAYOBJECT, $arrayObjectPrototype = null)
     {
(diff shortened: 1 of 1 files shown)


library/Zend/Db/ResultSet/ResultSet.php
```diff
@@ -46,7 +46,7 @@ class ResultSet extends AbstractResultSet
      */
     public function __construct($returnType = self::TYPE_ARRAYOBJECT, $arrayObjectPrototype = null)
     {
-        $this->returnType = (in_array($returnType, array(self::TYPE_ARRAY, self::TYPE_ARRAYOBJECT))) ? $returnType : self::TYPE_ARRAYOBJECT;
+        $this->returnType = (self::TYPE_ARRAY === $returnType) ? $returnType : self::TTYPE_ARRAYOBJECT;
```
Pittiplatsch: `self::TYPE_ARRAYOBJECT` - note the wrong double "T"!

Maks3w:
Just for semantic reasons. Before was clear valid arguments and then the default fallback.

weierophinney:
Closing, as the branch is no longer available.

PR #7270, opened 2015-02-26
PHP 7 on Travis-CI (redux)

weierophinney:
This is a rebased version of #7263, on a clean develop branch that has passing tests on all supported PHP 5 versions.

PR #7263, opened 2015-02-26
add php 7 on .travis.yml

samsonasik:

PR #7366, opened 2015-03-24
PHP 7 on Travis-CI (re-redux)

weierophinney:
This is a version of #7270 (which was in turn a rebased version of #7263), rebased against master, to support PHP 7.

I have tested both against a nightly version from 2 weeks ago, as well as the Coercive STH patch. The latterwhich provided me with a number of issues in the ZF2 codebase with regard to passing invalid arguments to internal functions, and the bulk of the commits in this branch address those, keeping compatibility with PHP 5 versions as well as PHP 7 nightly.

There is one issue I've found in `Zend\Stdlib\ArrayUtils::filter()`: apparently, PHP 7 nightly has a regression in its `array_filter` implementation; the third argument, which indicates whether to pass values only, keys only, or both to the composed callback, is ignored, causing one test case to fail.

Otherwise, this version should be ready to merge anytime.

### DIFF
--- a/library/Zend/Cache/Storage/Adapter/Dba.php
+++ b/library/Zend/Cache/Storage/Adapter/Dba.php
@@ -264,7 +264,8 @@ class Dba extends AbstractAdapter implements
 
         $this->_open();
 
-        do { // Workaround for PHP-Bug #62491 & #62492
+        // Workaround for PHP-Bug #62491 & #62492
+        do {
             $recheck     = false;
             $internalKey = dba_firstkey($this->handle);
             while ($internalKey !== false && $internalKey !== null) {
@@ -377,8 +378,10 @@ class Dba extends AbstractAdapter implements
         $prefix      = ($namespace === '') ? '' : $namespace . $options->getNamespaceSeparator();
         $internalKey = $prefix . $normalizedKey;
 
+        $cacheableValue = (string) $value; // dba_replace requires a string
+
         $this->_open();
-        if (!dba_replace($internalKey, $value, $this->handle)) {
+        if (!dba_replace($internalKey, $cacheableValue, $this->handle)) {
             throw new Exception\RuntimeException("dba_replace('{$internalKey}', ...) failed");
         }
 
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,5 @@
 .*.sw*
 .*.un~
 build/
-composer.lock
 nbproject
 tmp/
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ php:
   - hhvm-nightly
 
 install:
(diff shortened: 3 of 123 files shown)


weierophinney:
Only failure on PHP 7 at this point is with `Zend\Stdlib\ArrayUtils::filter()`, which is due to a regression in PHP 7's `array_filter()` implementation, which I have reported here: https://bugs.php.net/bug.php?id=69299

weierophinney:
This change will require some documentation:

- `Zend\Filter\Int` is deprecated (`int` is a reserved keyword in PHP 7); `Zend\Filter\ToInt` is preferred.
- `Zend\I18n\Validator\Int` is deprecated; `Zend\I18n\Validator\IsInt` is preferred.
- `Zend\I18n\Validator\Float` is deprecated (`float` is a reserved keyword in PHP 7); `Zend\I18n\Valdidator\IsFloat` is preferred.
- `Zend\Db\Sql\Ddl\Column\Float` is deprecated; `Zend\Db\Sql\Ddl\Column\Floating` is preferred.
- `Zend\XmlRpc\Value\String` has been renamed to `Zend\XmlRpc\Value\Text` (`string` is a reserved keyword in PHP 7). This should only affect developers who are directly instantiating XML-RPC value objects, which is so uncommon as to not warrant a BC shim.

For filters and validators, if you are pulling them from the plugin managers, you will continue to be able to do so under the old names, and the new class will be resolved; if you are instantiating them directly, you will receive a deprecation notice indicating the new class to use.

library/Zend/Cache/Storage/Adapter/Dba.php
```diff
@@ -377,8 +378,10 @@ protected function internalSetItem(& $normalizedKey, & $value)
         $prefix      = ($namespace === '') ? '' : $namespace . $options->getNamespaceSeparator();
         $internalKey = $prefix . $normalizedKey;
 
+        $cacheableValue = (string) $value; // dba_replace requires a string
+
         $this->_open();
-        if (!dba_replace($internalKey, $value, $this->handle)) {
+        if (!dba_replace($internalKey, $cacheableValue, $this->handle)) {
```
DASPRiD: This variable is only used once, why not cast it in-place?

weierophinney: So I could provide a comment as to why the casting was happening.

weierophinney:
all is well! As @DASPRiD has reviewed, I'll merge, and push an RC3 immediately.

weierophinney:
So, @DASPRiD pointed out two more PHP RFCs that would affect this:

- https://wiki.php.net/rfc/reserve_more_types_in_php_7
- https://wiki.php.net/rfc/reserve_even_more_types_in_php_7

I found 5 more classes that will need shims for forward and backwards compat:

- `Zend\Form\Annotation\Object`
- `Zend\Filter\Null`
- `Zend\Paginator\Adapter\Null`
- `Zend\Mail\Transport\Null`

I'll work on these now. Until those are ready, I'm marking this as WIP.

gianarb:
:+1: Thanks!!

PR #7369, opened 2015-03-26
Composer update to 2.4-RC3

weierophinney:
During `composer update` I get:

```
[RuntimeException]                                                                                                                                                             
  Could not load package zendframework/zendframework in http://packagist.org: [UnexpectedValueException] Could not parse version constraint ^1.0: Invalid version string "^1.0"
```

Dash is new composer feature available in master branch. In our company we use only last "stable" version (current alpha9). I think we should also define supported composer version in zf2. Currently we are not able to test RC or switch to 2.4 in future :(

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,8 +13,8 @@
         "zendframework/zendxml": "~1.0-dev"
     },
     "require-dev": {
-        "doctrine/annotations": "^1.0",
-        "ircmaxell/random-lib": "^1.1",
+        "doctrine/annotations": "~1.0",
+        "ircmaxell/random-lib": "~1.1",
         "mikey179/vfsStream": "~1.2",
         "fabpot/php-cs-fixer": "~1.0",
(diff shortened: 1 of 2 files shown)


snapshotpl:
Thank @weierophinney for fast feedback and fix! Looks good for me :+1:

weierophinney:
@snapshotpl I've tagged 2.4.0rc4 at this time with this change; please let me know if you have any further issues!

snapshotpl:
@weierophinney Unfortunately we still get issue. After research we have found that in file .composer/cache/repo/https---packagist.org/provider-zendframework$zendframework.json in rc3 ^ exists and fails composer update. Possible solution for that is remove affected tag (rc3) from packagist.

snapshotpl:
@weierophinney any chance to remove RC3?

Ocramius:
Removing tags is not an option

Issue #7368, opened 2015-03-26
Zend\Http\Response should treat 410 like a not found


coogle:
According to RFC 2616 (http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) A server has the option of returning a HTTP 410 status code when a resource is not found and moreover the server understands the resource will never return (gone).

From a ZF perspective, it seems to me that the Response::isNotFound() method should return true in this case, and at present it does not. A 410 is a 404, with more insight into the status of the resource.

Martin-P:
Sounds like a bad idea IMHO. I understand a `410` can be interpreted as a `404`, but is fundamentally different. A `410` indicates the resource existed once and is gone now, a `404` however indicates a resource that was never there at all. Furthermore the term "Not found" is always used in a `404` context, because it is the phrase that belongs to a `404` status code. When you add a `410` to the same description "Not found" (in ZF) you create confusion, because the term which was meant for one specific response code (`404`) is suddenly used for multiple response codes (`404` and `410`).

As a compromise a new method `isGone()` looks like a clean solution to me. That way you can make it easier if someone wants to check if a resource is gone.

macnibblet:
I agree with @Martin-P because frankly they are two different things and if the api is so advanced it supports 410 then you should be looking for that specifically.  

This is also not that complicated
```php
if ($response->isNotFound() || $response->isGone()) 
```

coogle:
It's not that "can be interpreted" is a 404, the RFC Specifically states that if the server *knows* the resource was there and is gone it should return 410. 404 Does NOT mean the resource was never there at all, just that simply it cannot find anything at the moment -- they are not two different things but simply the same thing with more detail. 

#### 404
> The server has not found anything matching the Request-URI. No indication is given of whether the 
> condition is temporary or permanent. The 410 (Gone) status code SHOULD be used if the server 
> knows, through some internally configurable mechanism, that an old resource is permanently 
> unavailable and has no forwarding address. This status code is commonly used when the server does > not wish to reveal exactly why the request has been refused, or when no other response is applicable.

I will be personally satisfied with an `isGone()`, but I disagree that `isNotFound()` returning true on a 404 OR a 410 is somehow confusing or broken behavior. The RFC is pretty explicit on the issue, and the argument that users will be confused when as it was pointed out that this is an "advanced case" is pretty flimsy to me.

GeeH:
[...]

PR #7370, opened 2015-03-26
Add isGone to response

asgrim:
As per the discussion in zendframework/zf2#7368 this change adds a new
method `\Zend\Http\Response->isGone()` to determine if the status code is
a "410 Gone". This works exactly the same as the similar methods like
`isNotFound` simply by checking the status code.

This change takes the opinion that a 410 Gone should not be included in
the `isNotFound()` method. Further discussion can be seen in the issue.

### DIFF
--- a/library/Zend/Http/Response.php
+++ b/library/Zend/Http/Response.php
@@ -389,6 +389,16 @@ class Response extends AbstractMessage implements ResponseInterface
     }
 
     /**
+     * Does the status code indicate the resource is gone?
+     *
+     * @return bool
+     */
+    public function isGone()
+    {
+        return (410 === $this->getStatusCode());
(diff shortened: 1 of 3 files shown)


library/Zend/Http/Response.php
```diff
@@ -389,6 +389,16 @@ public function isNotFound()
     }
 
     /**
+     * Does the status code indicate the resource is not found?
```
Martin-P: You can use the status code description here (Gone):
> Does the status code indicate the resource is gone?

asgrim: Thanks, amended :)

asgrim:
Closed in favour of rebased PR #7536

PR #7372, opened 2015-03-26
Zend\Mail, Zend\Mime: parsing rework

echampet:
Hi,

I've ran ~100000 mails through Zend\Mail\Message::fromString(), and send them to an smtp server (postfix), and here are some fixes.

I really think that we should kill the whole encoding thing and use mb_detect_encoding (don't know if mb dependency is ok)

Please start to review/comment.


echampet:
hi guys,
if update my PR, will someone merge it?

gianarb:
Hi! Sorry for delay but now every component is in its personal repository.. 
Can you create new PR here?
https://github.com/zendframework/zend-mail
Thanks a lot!

PR #7373, opened 2015-03-27
run composer self-update at before_install in .travis.yml

samsonasik:
so we always use latest composer


Maks3w:
There is no reason for to be on the edge.

PR #7376, opened 2015-03-27
Fixes DocBlock for BasePath helper in PhpRenderer

froschdesign:


library/Zend/View/Renderer/PhpRenderer.php
```diff
@@ -30,7 +30,7 @@
  *
  * Convenience methods for build in helpers (@see __call):
  *
- * @method \Zend\View\Helper\BasePath basePath($file = null)
+ * @method string basePath($file = null)
```
Maks3w: should be `null|string`

froschdesign:
ping @Maks3w

PR #7379, opened 2015-03-28
Why to call `info()` again

Moln:
Connection is not broken.
Why to call `info()` again?


library/Zend/Cache/Storage/Adapter/RedisResourceManager.php
```diff
@@ -119,9 +119,10 @@ public function getResource($id)
             //in case new server was set then connect
             if (!$resource['initialized']) {
                 $this->connect($resource);
+
+                $info = $resource['resource']->info();
+                $resource['version'] = $info['redis_version'];
```
weierophinney: This is already present in current master:

- https://github.com/zendframework/zf2/blob/master/library/Zend/Cache/Storage/Adapter/RedisResourceManager.php#L123-L125

Additionally, because it calls `return` within the conditional block, the call to `info()` later in the method is still required.

PR #7382, opened 2015-03-29
Minor change on bootstraps of tests

malukenho:


Maks3w:
This file will be removed on ZF 2.5.0

PR #7384, opened 2015-03-30
Add ValidatorServiceInterface for session validators

### LINKED ISSUE #7381: [Feature] Allow Session validators to be created using the service manager
At the moment it is not possible to inject any dependencies into a session validator. There is simply no way to do that, since all session validators are created inside the `SessionManager` or the `ValidatorChain`.

In a perfect world, the word `new` would never have made it into any of the Session classes, but here we are.

I think I can hack a way around this by extending `ValidatorInterface` and add a method called `setData()`, which could be used to inject the session value to validate, instead of using the constructor to do that. This would allow a factory to create the validator, and the `ValidatorChain` to inject the session value to validate.

By extending the `ValidatorInterface` there will be no BC breaks, but the code will get pretty ugly. I still think that would be better than the current situation.

Do you have any thoughts on this?

larsnystrom:
This PR fixes #7381 without any BC breaks. It allows a session validator to be instantiated with dependencies and adds some default functionality to the `SessionManagerFactory` to `get()` session validators from the `ServiceManager`.

It introduces the new interface `Zend\Session\Validator\ValidatorServiceInterface` which extends `Zend\Session\Validator\ValidatorInterface`. It modifies the `Zend\Session\Service\SessionManagerFactory` to also inject validator services into the session manager. Lastly it modifies `Zend\Session\SessionManager` and `Zend\Session\ValidatorChain` to attach the validator services to the `ValidatorChain` and inject the reference value from the current session into the validator service using the `setData()` method defined in the `ValidatorServiceInterface`.

A validator which implements the `ValidatorServiceInterface` can be registered under the key `services` in the configuration array for session validators, like this:

```
array (
    'session_manager' => array(
        'validators' => array(
            'OldSessionValidatorWillWork',
            'services' => array(
                'NewValidatorServicesGoHere',
            ),
        ),
    ),
);
```

The example above illustrates that the old way of registering validators still works, and that new validator services goes under the `services` key of the `validators` array.

All validators under the `services` key must:
- implement the `ValidatorServiceInterface`, and
- be registered with the `ServiceManager`.

I'm open for discussion, but I think this is the best way to implement this feature without a BC break.

Caveats:
- In the old way of doing things we only registered session validators when creating the session. On subsequent requests the session validators would be instantiated automatically. The new validator services work differently in that they must be injected in the `SessionManager` on each request. If a service validator isn't injected it will be improperly created by the `ValidatorChain`, which may cause unknowing developers some trouble. 

Things I haven't thought of yet: (hey, that's a paradox!)
- How to insert the validators when creating the session. I'm not sure how to do this without removing them all, re-getting them from the service manager, injecting the reference value and then attaching them again. There are probably other ways, but it's still overly complicated.

### DIFF
--- a/library/Zend/Session/Validator/ValidatorServiceInterface.php
+++ b/library/Zend/Session/Validator/ValidatorServiceInterface.php
@@ -1,0 +1,17 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
+ * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+namespace Zend\Session\Validator;
+
+/**
+ * Session validator interface created by factory
+ */
+interface ValidatorServiceInterface extends ValidatorInterface
+{
+	public function setData($data);
+}
--- a/library/Zend/Session/AbstractManager.php
+++ b/library/Zend/Session/AbstractManager.php
@@ -54,19 +54,26 @@ abstract class AbstractManager implements Manager
     protected $validators;
 
     /**
+     * @var array
+     */
+    protected $validatorServices;
+
+    /**
      * Constructor
      *
      * @param  Config|null      $config
      * @param  Storage|null     $storage
      * @param  SaveHandler|null $saveHandler
      * @param  array            $validators
+     * @param  array            $validatorServices
      * @throws Exception\RuntimeException
      */
     public function __construct(
         Config $config = null,
         Storage $storage = null,
         SaveHandler $saveHandler = null,
-        array $validators = array()
+        array $validators = array(),
+        array $validatorServices = array()
     ) {
         // init config
         if ($config === null) {
@@ -118,6 +125,7 @@ abstract class AbstractManager implements Manager
         }
 
         $this->validators = $validators;
+        $this->validatorServices = $validatorServices;
     }
 
     /**
--- a/library/Zend/Session/Service/SessionManagerFactory.php
+++ b/library/Zend/Session/Service/SessionManagerFactory.php
@@ -17,6 +17,7 @@ use Zend\Session\Container;
 use Zend\Session\SaveHandler\SaveHandlerInterface;
 use Zend\Session\SessionManager;
 use Zend\Session\Storage\StorageInterface;
+use Zend\Session\Validator\ValidatorServiceInterface;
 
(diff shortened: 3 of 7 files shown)


Martin-P:
I like the way you fixed this without a BC break, but it also feels a bit like a workaround for the somewhat messy implementation of the `SessionValidators`. Personally I think `Zend\Session` can use a bit of a cleanup, so `SessionValidators` are created at *one*  point in the process. Yes, that would mean a BC break, but looking at the current implementation I think that it can only be an improvement of the code.

See also my other comment (https://github.com/zendframework/zf2/issues/7381#issuecomment-87581173) regarding the `SessionValidators`:
> Validators can be created at multiple points in the process: inside `SessionManager` and inside `ValidatorChain`. IMO the `ValidatorChain` should never create its own validators and should only be a dumb object containing the callback functions for validating the session. The validators itself should be injected at *one* point in the process and the session values can be injected there. Also that would make it easier to implement the feature with the `ServiceManager`.

larsnystrom:
I'm all for refactoring of `Zend\Session`, but since it would cause a BC break it will probably have to wait until ZF3 (or until somebody splits ZF into it's separate components). There are some serious issues with the current implementation, so I'm looking forward to that, but without word from higher up the chain of command here, it's just not possible.

I also agree with you that the `ValidatorChain` should never create it's own objects. However, that's another thing that can't be changed without a BC break. To be honest I think we should scrap the `ValidatorChain` altogether and just use an array in the `SessionManager`. I don't see the point in using a priority queue when we can't even specify priorities. 

I also think it was a bad idea to force validation of the session in `SessionManager::start()`, especially when you can't specify which values to validate before running it.

One thing that could be done is adding an argument to `ValidatorInterface::isValid()`, to provide the session storage to the validator. That way you could validate pretty much anything you'd like, even certain combinations of values. It would also be a more intuitive way of doing the validation, following the same pattern used in `Zend\Validator` where `isValid()` always takes the value to be validated as the argument.

I guess this component was put together in a hurry. Splitting ZF into separate components would allow us to start fixing the real problems here. In the meantime we can stuff like I did here.

library/Zend/Session/Validator/ValidatorServiceInterface.php
```diff
@@ -0,0 +1,17 @@
...
+ */
+namespace Zend\Session\Validator;
+
+/**
+ * Session validator interface created by factory
+ */
+interface ValidatorServiceInterface extends ValidatorInterface
+{
+	public function setData($data);
+}
```
samsonasik: new line after <code>}</code> for eof

GeeH:
New issue can be found at: https://github.com/zendframework/zend-session/issues/48

PR #7386, opened 2015-03-30
Fix Http\Response decode gzip

sokac:
Reference:
- http://php.net/manual/en/function.gzinflate.php#77336
- http://www.faqs.org/rfcs/rfc1952.html

p.s. added test case which breaks old code. However, I'm not sure about its license.


GeeH:
New issue can be found at: https://github.com/zendframework/zend-http/issues/71

PR #7390, opened 2015-03-31
update @return to use mixed in ServiceLocatorInterface::get($name)

samsonasik:
this can be happen when we have a class like the following : 

``` php
class Foo
{
    public function __construct(BarInterface $bar)
    {}
}
```

if we use scrutinizer to check the code, when make it as service and we instantiate the class using factory, we will get documentation issue : 

```
$serviceLocator->get('Bar') is of type object|array, but the function expects a object<BarInterface>. 
```

that's why I think we can use @mixed.

### DIFF
--- a/library/Zend/ServiceManager/ServiceLocatorInterface.php
+++ b/library/Zend/ServiceManager/ServiceLocatorInterface.php
@@ -19,7 +19,7 @@ interface ServiceLocatorInterface
      *
      * @param  string  $name
      * @throws Exception\ServiceNotFoundException
-     * @return object|array
+     * @return mixed
      */
     public function get($name);
 


Martin-P:
Isn't it better to create a new issue at Scrutinizer for this? I don't like the idea to make return annotations more vague and affect all users with their IDE because of some third party software.

samsonasik:
although without scrutinizer check, I still think @mixed is safer

Martin-P:
`mixed` is always safer when using multiple types, but is the annotation there for safety or to inform the user/IDE about the return value they can expect? I'd rather have `object|array`, because then I know I can expect an `object` or `array` and do not have to build in checks for `int`, `string`, `bool` or `null`.

Just imagine if everything with 2 or more types would be changed into `mixed`. I think I can throw away my IDE :smile:

samsonasik:
rely-ing with IDE is equals than rely-ing 3rd party software imo. btw, I've found that <code>Zend\ServiceManager\FactoryInterface::createService()</code> is using <code>@return mixed</code>, it means that the instance can be _mixed_ to support this proposal

Martin-P:
> rely-ing with IDE is equals than rely-ing 3rd party software imo.

The big difference is you need an editor/IDE to write code, but you don't need Scrutinizer to write code.

> it means that the instance can be mixed to support this proposal

IMO it means the return value of `Zend\ServiceManager\FactoryInterface::createService()` is faulty and needs to be updated. You use the `ServiceManager` to get certain classes, not to get a `boolean` or an `int`.

stefanotorresi:
@Martin-P while I understand your point, the fact that the `ServiceManager` should not be used to retrieve scalars is merely a recommendation, as AFAIK there is no code in the concrete implementation to ensure that that's actually the case.

Besides, if `FactoryInterface::createService()` hint is not the same, consistency is due one way or another, and I'd argue in favor of this PR.

Issue #7388, opened 2015-03-31
Adding getServiceLocator() to ServiceLocatorInterface


manuakasam:
Hey guys,

so I was just reading the newest Blog by @samsonasik and it got me thinking. Here a link to the article https://samsonasik.wordpress.com/2015/03/31/zend-framework-2-using-__invokepluginmanager-manager-in-services-factory/

Basically it teaches the approach of using Invokable classes as ServiceFactories. And I don't like it.

# Where is the Problem?

As a community we are teaching newcomers to differentiate between two kinds of services: `invokables` and `factories`. If now the community starts to only use `invokables` because of proper type-hinting, then newcomers will simply be blown away with a bit of "wtf?".

While it is perfectly fine to come up with new ways of doing things, I feel that the problem is of a different kind of nature and that we should rather focus on solving said problem than to find workarounds for it.

# Analyzing the Problem

Service-Factories for the default `ServiceManager` we're writing a Factory like this:

```php
class FooFactory implements FactoryInterface
{
    public function createService(ServiceLocatorInterface $sl) 
    {
        return new FooService(
            $sl->get('FooDependency')
        );
    }
}
```

All nice and clean. Both parties, Users and Code-Inspection, are very happy.

However when we're outside the scope of the normal `ServiceManager`, for example when writing Controller-Factories, things are changing. Take a look at this example:

```php
class BarControllerFactory implements FactoryInterface
{
    public function createService(ServiceLocatorInterface $sl) 
    {
        /** @var $realSl ControllerManager
        $realSl = $sl->getServiceLocator();
        
        return new BarController(
            $realSl->get('BarDependency')
        );
    }
}
```

And to me this is where we run into issues. While we're successfully tricking our IDE into giving us proper code-completion, code-inspection tools like Scrutinizer are analyzing the variable based on what we pass into the factory. That that's only the `ServiceLocatorInterface`. An interface that does not know about `getServiceLocator()`.

From a programming point of view I do understand why we're using the shared interface. However the issue that I have - and I'm sure I can't be the only one - is, that we're relying on something that isn't defined by the interface. In essence that's bad design.

# What can be done?

Now I haven't given this many thorough thoughts yet, but as far as a quick brainstorm is concerned I see two possible solutions.

## Create new Interfaces

[...]

PR #7392, opened 2015-03-31
improvments for Zend\InputFilter

kokspflanze:
HI,

added missing methods @ Zend\InputFilter\InputInterface
added some doc-blocks in Zend\InputFilter*Interface
little code improvements, for example signature say return $this and the method return nothing

Thanks=)

### DIFF
--- a/library/Zend/InputFilter/InputFilterAbstractServiceFactory.php
+++ b/library/Zend/InputFilter/InputFilterAbstractServiceFactory.php
@@ -22,17 +22,14 @@ class InputFilterAbstractServiceFactory implements AbstractFactoryInterface
     protected $factory;
 
     /**
-     * @param ServiceLocatorInterface $inputFilters
+     * @param ServiceLocatorInterface $services
      * @param string                  $cName
      * @param string                  $rName
      * @return bool
      */
-    public function canCreateServiceWithName(ServiceLocatorInterface $inputFilters, $cName, $rName)
+    public function canCreateServiceWithName(ServiceLocatorInterface $services, $cName, $rName)
     {
-        $services = $inputFilters->getServiceLocator();
-        if (! $services instanceof ServiceLocatorInterface
-            || ! $services->has('Config')
-        ) {
+        if (!$services->has('Config')) {
             return false;
         }
 
@@ -52,9 +49,8 @@ class InputFilterAbstractServiceFactory implements AbstractFactoryInterface
      * @param string                  $rName
      * @return \Zend\InputFilter\InputFilterInterface
      */
-    public function createServiceWithName(ServiceLocatorInterface $inputFilters, $cName, $rName)
+    public function createServiceWithName(ServiceLocatorInterface $services, $cName, $rName)
     {
-        $services  = $inputFilters->getServiceLocator();
         $allConfig = $services->get('Config');
         $config    = $allConfig['input_filter_specs'][$rName];
 
--- a/library/Zend/InputFilter/InputFilterAbstractServiceFactory.php
+++ b/library/Zend/InputFilter/InputFilterAbstractServiceFactory.php
@@ -22,14 +22,17 @@ class InputFilterAbstractServiceFactory implements AbstractFactoryInterface
     protected $factory;
 
     /**
-     * @param ServiceLocatorInterface $services
+     * @param ServiceLocatorInterface $inputFilters
      * @param string                  $cName
      * @param string                  $rName
      * @return bool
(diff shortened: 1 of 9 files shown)


kokspflanze:
i will fix the phpunit error tomorrow, sry=[

library/Zend/InputFilter/InputFilterAbstractServiceFactory.php
```diff
@@ -52,9 +49,8 @@ public function canCreateServiceWithName(ServiceLocatorInterface $inputFilters,
      * @param string                  $rName
      * @return \Zend\InputFilter\InputFilterInterface
      */
-    public function createServiceWithName(ServiceLocatorInterface $inputFilters, $cName, $rName)
```
Martin-P: This is actually correct, because it is an instance of `InputFilterPluginManager` and not the `ServiceManager` itself.

kokspflanze: so the method signature has to be changed to AbstractPluginManager ?
For me was the problem that the ServiceLocatorInterface doesnt have the method getServiceLocator.

Martin-P: No, it can stay just the way it was. Regarding the method `getServiceLocator` there already is an other issue for this and I think it depends on the outcome of that discussion whether it needs a change to ` AbstractPluginManager`. See #7388.

kokspflanze:
thanks @Martin-P for the hint.

I revert the InputFilterAbstractServiceFactory class.

library/Zend/InputFilter/InputInterface.php
```diff
@@ -26,6 +26,13 @@ public function setAllowEmpty($allowEmpty);
      */
     public function setBreakOnFailure($breakOnFailure);
 
+
+    /**
+     * @param bool $continueIfEmpty
+     * @return self
+     */
+    public function setContinueIfEmpty($continueIfEmpty);
+
```
weierophinney: This is defined in `EmptyContextInterface`, and is there because we cannot add to existing interfaces (causes a BC break). This is why `Input` implements `EmptyContextInterface`.

library/Zend/InputFilter/InputInterface.php
```diff
@@ -63,6 +70,12 @@ public function setValidatorChain(ValidatorChain $validatorChain);
     public function setValue($value);
 
     /**
+     * @param  mixed $value
+     * @return self
+     */
+    public function setFallbackValue($value);
```
weierophinney: We cannot add methods to existing interfaces (causes a BC break). We *can* create a new interface that defines this method, and update `Input` to use that interface (instead of only defining it within the concrete class).

library/Zend/InputFilter/InputInterface.php
```diff
@@ -79,6 +92,11 @@ public function allowEmpty();
     public function breakOnFailure();
 
     /**
+     * @return bool
+     */
+    public function continueIfEmpty();
```
weierophinney: Also defined in `EmptyContextInterface` already.

weierophinney:
@kokspflanze — As noted, with the current changes, I cannot merge this as it introduces BC breaks.

If you remove the new methods from the `InputInterface` entirely, I can merge it for 2.4.1. If you want to move the `setFallbackValue()` method into a new interface, I can merge for 2.5.0.

kokspflanze:
@weierophinney thanks for the feedback, i will change it asap. =)

library/Zend/InputFilter/InputInterface.php
kokspflanze: We use this method @ [Input](https://github.com/zendframework/zf2/blob/develop/library/Zend/InputFilter/Input.php#L300) , i think we have to add something to support the old `InputInterface` too, maybe with `method_exists`, or you know a better way?

library/Zend/InputFilter/InputInterface.php
kokspflanze: We use this method in [Factory](https://github.com/zendframework/zf2/blob/develop/library/Zend/InputFilter/Factory.php#L231), so we should add a method_exists, or how should be the handling for that.

jerv13:
After updating to 2.4.1, all the Zend\Filter\Boolean filters are causing validation errors.  I have not investigated in detail, but is likely related to this change.

kokspflanze:
@jerv13 can you give an example?

PR #7394, opened 2015-04-01
fixed return type on disconnect method in pgsql connection

yp28:
According to the inherited docs this should return an instance of ConnectionInterface. Fixed this return type, asking myself why the feedback of the pg_close (boolean) method wasn't returned.

Anyone knows why the feedback isn't returned?

### DIFF
--- a/tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
+++ b/tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
@@ -57,10 +57,11 @@ class ConnectionTest extends \PHPUnit_Framework_TestCase
      */
     public function testDisconnect()
     {
-        $this->markTestIncomplete(
-            'This test needs work.'
-        );
-        $this->assertInstanceOf('\Zend\Db\Adapter\Driver\ConnectionInterface',$this->connection->disconnect());
+        if (! extension_loaded('pgsql')) {
+            $this->markTestSkipped('pgsql extension not loaded');
+        }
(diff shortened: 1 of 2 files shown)


YP28:
I will try to provide a unittest later today so this won't happen again. I'll if I can implement some more tests on the PostgreSQL driver since there are hardly any.

YP28:
Help needed on the unittest. How to mock the resource in the connection to allow the pg_close method to execute?

weierophinney:
@YP28 Best bet for getting assistance for the tests is to join the #zftalk.dev IRC channel on Freenode; somebody there will likely be able to assist.

tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
```diff
@@ -52,6 +52,19 @@ public function testResource()
...
     /**
+     * Test disconnect method to return instance of ConnectionInterface
+     * @TODO Exception was raised due to invalid resource. Add resource to make this test pass.
+     */
+    public function testDisconnect()
+    {
+        if (! extension_loaded('pgsql')) {
+            $this->markTestSkipped('pgsql extension not loaded');
+        }
+        $this->assertInstanceOf('\Zend\Db\Adapter\Driver\ConnectionInterface', $this->connection->disconnect());
```
Maks3w: Since we are testing `return $this` is better to use `assertSame($this->connection, $this->connection->disconnect());`

weierophinney: I made the change on merge.

Issue #7377, opened 2015-03-27
We can't create a multiple connection with Pgsql


wandersonwhcr:
Olá pessoas!

I'm trying to use a multiple connection with `pgsql` adapter, but it *always* retrieve the same PHP resource, even setting a multiple configuration in `ServiceManager`.

Why?

http://php.net/manual/en/function.pg-connect.php

If we use same configuration in both connections, we'll always receive same resource. But if we pass `PGSQL_CONNECT_FORCE_NEW` parameter, all works fine.

[8d603705 master Zend\Db\Adapter\Driver\Pgsql\Connection](https://github.com/zendframework/zf2/blob/8d603705224076429540aa38af7f74e9e3079b85/library/Zend/Db/Adapter/Driver/Pgsql/Connection.php)

Thanks.

samsonasik:
linked with #7396

GeeH:
This issue has been moved from the `zendframework` repository as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html
New issue can be found at: https://github.com/zendframework/zend-db/issues/129

PR #7396, opened 2015-04-01
added connection type to allow force new connection

yp28:
#7377

Connection now allows to specify type.

### DIFF
--- a/tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
+++ b/tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
@@ -80,4 +80,25 @@ class ConnectionTest extends \PHPUnit_Framework_TestCase
             $getConnectionString->invoke($this->connection)
         );
     }
+
+    /**
+     * @expectedException \Zend\Db\Adapter\Exception\InvalidArgumentException
+     */
+    public function testSetConnectionTypeException()
+    {
+        $this->connection->setType(3);
+    }
+
+    /**
+     * Test the connection type setter
+     */
+    public function testSetConnectionType()
+    {
+        $type = PGSQL_CONNECT_FORCE_NEW;
+        $this->connection->setType($type);
+        $reflectionProperty = new \ReflectionProperty($this->connection, 'type');
+        $reflectionProperty->setAccessible(true);
+        $this->assertEquals($type, $reflectionProperty->getValue($this->connection));
+    }
+
 }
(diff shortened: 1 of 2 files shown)


YP28:
Test failed due to use of undefined constant PGSQL_CONNECT_ASYNC. This constant is only available if the PGSQL extension is enabled in PHP.

tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
```diff
@@ -80,4 +80,25 @@ public function testGetConnectionStringEncodeSpecialSymbol()
...
+    /**
+     * Test the connection type setter
+     */
+    public function testSetConnectionType()
+    {
+        $type = PGSQL_CONNECT_FORCE_NEW;
+        $this->connection->setType($type);
+        $reflectionProperty = new \ReflectionProperty($this->connection, 'type');
+        $reflectionProperty->setAccessible(true);
+        $this->assertEquals($type, $reflectionProperty->getValue($this->connection));
```
Martin-P: Instead of `ReflectionProperty` you can use `readAttribute()` for this:

```php
public function testSetConnectionType()
{
    $type = PGSQL_CONNECT_FORCE_NEW;
    $this->connection->setType($type);
    $this->assertEquals($type, self::readAttribute($this->connection, 'type'));
}
```

YP28: Cool, thanks for the feedback!

samsonasik:
I think you can add : 
```php
 if (! extension_loaded('pgsql')) {
     $this->markTestSkipped('pgsql extension not loaded');
}
```

tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
```diff
@@ -80,4 +80,26 @@ public function testGetConnectionStringEncodeSpecialSymbol()
             $getConnectionString->invoke($this->connection)
         );
     }
+
+    /**
+     * @expectedException \Zend\Db\Adapter\Exception\InvalidArgumentException
+     */
+    public function testSetConnectionTypeException()
+    {
+        $this->connection->setType(3);
```
weierophinney: This test also needs to be skipped if pgsql is not present, as it's exercising code that will check for those constants. I can make that change on merge.

Maks3w:
`PGSQL_CONNECT_ASYNC` was added on PHP 5.6

weierophinney:
I'll create a PR that fixes this.

PR #7403, opened 2015-04-02
register missing ZendXml namespace in demos/Zend/XmlRpc/xmlrpc-upc-lookup.php

samsonasik:


samsonasik:
I'm thinking if we can use composer autoloader instead in all demos file ?
```php
include_once __DIR__ . '/../../../vendor/autoload.php';
```
as registering namespace under vendor/zendframework/zendxml with Zend\Loader is we are sure already use composer.

weierophinney:
Honestly, I'm not sure if any of the demos actually even work at this point, and I'm evaluating removing them entirely for 2.5.

PR #7405, opened 2015-04-02
Minimize duplicated code on ArrayUtils

malukenho:


library/Zend/Stdlib/ArrayUtils.php
```diff
@@ -238,12 +238,7 @@ public static function iteratorToArray($iterator, $recursive = true)
                 continue;
             }
 
-            if ($value instanceof Traversable) {
```
Ocramius: The `instanceof` check is in place because it is less expensive than the `is_array()` call.

It looks like duplication, but this is really performance-sensitive code :-(

Ocramius:
Closing as `inapplicable`

malukenho:
@Ocramius if I use `$value instanceof Traversable || is_array($value)` instead?

library/Zend/Stdlib/ArrayUtils.php
```diff
@@ -238,12 +238,7 @@ public static function iteratorToArray($iterator, $recursive = true)
...
-                $array[$key] = static::iteratorToArray($value, $recursive);
-                continue;
-            }
-
-            if (is_array($value)) {
+            if (is_array($value) || $value instanceof Traversable) {
                 $array[$key] = static::iteratorToArray($value, $recursive);
```
Ocramius: Wondering if we can optimize this loop by iterating `if (is_array())` (instead of recursing if it is an `array`)

Maks3w:
Some parts of the code are written by this way just for make the code more legible.

May this could be more close to both purposes:

```php
if (
  $value instanceof Traversable
  || is_array($value)
) {
  // ...
}
```

Ocramius:
Yeah, what I wanted to suggest is to loop instead of doing a recursive call to `static::iteratorToArray()`

malukenho:
@Ocramius I tried using `foreach`, but I ended up repeating a lot more code.

Ocramius:
> but I ended up repeating a lot more code.

Yes, this happens a lot in performance-oriented code

GeeH:
New issue can be found at: https://github.com/zendframework/zend-stdlib/issues/62

PR #7410, opened 2015-04-03
Fixes #7356 : absolutiseUri() // link uri fix

### LINKED ISSUE #7356: Absolute URI doesn't work when link begin with //
Hi,

The function absolutiseUri in the class Feedset ( Feed/Reader) doesn't work if the link begin with "//".

Example : 

//example.com/feed return http://example.com/example.com/feed

The error is locate here : 

$link = $uri->getScheme() . '://' . $uri->getHost() . '/' . $this->canonicalizePath($link);

Thanks for your help.

samsonasik:
Fixes #7356

### DIFF
--- a/library/Zend/Feed/Reader/FeedSet.php
+++ b/library/Zend/Feed/Reader/FeedSet.php
@@ -67,17 +67,19 @@ class FeedSet extends ArrayObject
     protected function absolutiseUri($link, $uri = null)
     {
         $linkUri = Uri::factory($link);
-        if (!$linkUri->isAbsolute() or !$linkUri->isValid() && $uri !== null) {
-            $uri = Uri::factory($uri);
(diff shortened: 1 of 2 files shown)


samsonasik:
travis build only not work in php 5.3, not sure why...

weierophinney:
@samsonasik At first glance, my suspicion was that the difference is a change in `parse_url` between 5.3 and 5.4. Prior to 5.3, URIs that were scheme-relative were not parsed correctly; `parse_url('//example.com')` would return a _path_ component of `//example.com`. PHP 5.4 corrects his to return a _host_ component of `example.com`.

On looking at `absolutiseUri()`, however, I see it uses `Zend\Uri`, which implements its own parser. As such, the issue is likely in `Zend\Uri\Uri::parse()` somewhere.

samsonasik:
moved to https://github.com/zendframework/zend-feed/pull/2

PR #7411, opened 2015-04-03
remove unused use statement

samsonasik:


samsonasik:
latest php-cs-fixer updated make more unused uses detected, latest commit got error "line_after_namespace" , not related with this PR

Maks3w:
This was the same of #7460 finally done by #7477 and d2bda8f

PR #7412, opened 2015-04-03
Use === and !== in null checks instead of == and !=

samsonasik:
Test : 

``` php
$foo = '';
echo "\r\nTest with == \r\n";
if ($foo == null) {
    echo 'foo is null';
} else {
    echo 'foo is not null';
}

echo "\r\n-------------------\r\n";

echo "\r\nTest with ===\r\n";
if ($foo === null) {
    echo 'foo is null';
} else {
    echo 'foo is not null';
}
```

Result : 

```
Test with == 
foo is null
 ------------------- 
Test with === 
foo is not null
```


Xerkus:
best name and description :+1:

PR #7413, opened 2015-04-04
increase loop performance by assign count() to variable

samsonasik:
test : 

``` php
$data = range(1, 100000);

$time_start1 = microtime(true);
for ($i = 0; $i < count($data); $i++) {

}
$time_end1 = microtime(true);
$time1 = $time_end1 - $time_start1;

echo 'uses count() in iteration take time : ' . $time1;
echo "\r\n";

$time_start2 = microtime(true);
for ($i = 0, $count = count($data); $i < $count; $i++) {

}
$time_end2 = microtime(true);
$time2 = $time_end2 - $time_start2;

echo 'uses assignment of count() before iterated first take time ' . $time2;
echo "\r\n";
```

Result : 

```
uses count() directly in iteration take time : 0.11619091033936
uses assignment of count() before iterated first take time 0.017376184463501
```

### DIFF
--- a/library/Zend/Db/Sql/Platform/Mysql/Ddl/AlterTableDecorator.php
+++ b/library/Zend/Db/Sql/Platform/Mysql/Ddl/AlterTableDecorator.php
@@ -135,8 +135,8 @@ class AlterTableDecorator extends AlterTable implements PlatformDecoratorInterfa
                 if ($insert) {
                     $j = isset($j) ? $j : 0;
                     $sql = substr_replace($sql, $insert, $insertStart[$j], 0);
-                    $countInsertStart = count($insertStart);
-                    for (; $j < $countInsertStart; ++$j) {
+                    $insertStartCount = count($insertStart);
+                    for (; $j < $insertStartCount; ++$j) {
                         $insertStart[$j] += strlen($insert);
                     }
                 }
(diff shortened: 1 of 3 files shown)


library/Zend/Db/Sql/Platform/Mysql/Ddl/AlterTableDecorator.php
```diff
@@ -135,7 +135,8 @@ protected function processAddColumns(PlatformInterface $adapterPlatform = null)
                 if ($insert) {
                     $j = isset($j) ? $j : 0;
                     $sql = substr_replace($sql, $insert, $insertStart[$j], 0);
-                    for (; $j < count($insertStart); ++$j) {
+                    $countInsertStart = count($insertStart);
```
Martin-P: Count is a noun here not a verb, so it should be `$insertStartCount` instead of `$countInsertStart`.

samsonasik: done ;)

library/Zend/Ldap/Dn.php
```diff
@@ -553,9 +553,9 @@ public static function explodeDn(
         }
         $ret = array();
         for ($i = 0, $count = count($k); $i < $count; $i++) {
-            if (is_array($k[$i]) && is_array($v[$i]) && (count($k[$i]) === count($v[$i]))) {
+            if (is_array($k[$i]) && is_array($v[$i]) && (($countKey = count($k[$i])) === count($v[$i]))) {
```
Martin-P: Sorry, forgot this one: `$keyCount` instead of `$countKey`.

samsonasik: done ;)

PR #7416, opened 2015-04-04
Added $data parameter to delete() in AbstractRestfulController

ahoppen:
Now that the $data parameter has been added to `deleteList` in #7255, it would be inconsistent to not provide it to `delete`. 
Data for a delete request can be useful for example to validate a user’s password before deleting him.


GeeH:
New issue can be found at: https://github.com/zendframework/zend-mvc/issues/172

PR #7418, opened 2015-04-04
Update .php_cs

afurculita:
This just turns on the caching on php_cs

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -25,37 +25,37 @@ $finder = Symfony\CS\Finder\DefaultFinder::create()
             return false;
         }
     });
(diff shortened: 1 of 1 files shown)


localheinz:
@afurculita 

Maybe also have a look at this: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1076

Maks3w:
@afurculita Please rewrite the PR just adding the feature and removing every unrelated change.

/cc @keradus

Maks3w:
See this commit https://github.com/keradus/zf2/commit/6861cfd982face089361d6e8bad7667cf1581ce8. This is the correct way.

Issue #7407, opened 2015-04-02
Sql-query breaks on 2.4.0 compare to 2.3.7


freax:
...
        $select->columns(array(
            'count'        => new \Zend\Db\Sql\Expression('COUNT(*)'),
            'yearMonth'    => new \Zend\Db\Sql\Expression('FROM_UNIXTIME(date, "%Y-%m")'),
        ));
        $select->where(array('user' => $user));
        $select->group('yearMonth');

        $adapter = $this->getDbAdapter();
        $selectString = $sql->getSqlStringForSqlObject($select);

        echo $selectString;

In 2.3.7 this code echoed:

        SELECT COUNT(*) AS `count`, FROM_UNIXTIME(date, "%Y-%m") AS `yearMonth` FROM `table` WHERE `user` = '1' GROUP BY `yearMonth`

and everything was fine, but in 2.4.0:

        SELECT COUNT(*) AS `count`, FROM_UNIXTIME(date, "%%Y-%%m") AS `yearMonth` FROM `table` WHERE `user` = '1' GROUP BY `yearMonth`

And I have to rollback to 2.3.7 because of this. May be it related to #7400.

weierophinney:
Unrelated to #7400. I'll see if I can hunt this down.

samsonasik:
I think it related with  https://github.com/zendframework/zf2/blob/master/library/Zend/Db/Sql/Expression.php#L136 on code : 
```php
$expression = str_replace('%', '%%', $this->expression);
```

freax:
@samsonasik this line exists in 2.3.7 too
As I can see, at least the whole `\Zend\Db\Sql\Expression::getExpressionData()` was been rewritten

Martin-P:
I created a unittest for this and checked `Zend\Db\Sql\Expression` back to the version where it passes. The line @samsonasik is referring to is indeed the line causing the problem: commit https://github.com/zendframework/zf2/commit/455f3e1ede297594e241f419119c146f7e571809#diff-7a16e1c9e1d21e2bbc77ffe77ae82484 (PR #1443). It seems to be a part of the ZF 2.0.0 release which means the bug should also be present in ZF 2.3.7 (it is when I run the unittest below on ZF 2.3.7).

Before the introduction of that line the test passes, after the introduction of that line the test fails. 
```php
/**
 * @group 7407
 */
public function testGetExpressionDatePreservesPercentageSign()
{
[...]

PR #7422, opened 2015-04-04
Get correct expression string when Zend\Db\Sql\Expression is used

Martin-P:
Fix for #7407.

The processing of `Zend\Db\Sql\Expression` assumed the value of `$part` was the correct value to be used, however when an instance of `Zend\Db\Sql\Expression` is provided the correct value is in `Zend\Db\Sql\Expression::getExpression()`.

### DIFF
--- a/library/Zend/Db/Sql/AbstractSql.php
+++ b/library/Zend/Db/Sql/AbstractSql.php
@@ -108,6 +108,13 @@ abstract class AbstractSql implements SqlInterface
         $expressionParamIndex = &$this->instanceParameterIndex[$namedParameterPrefix];
 
         foreach ($parts as $part) {
+            // #7407: use $expression->getExpression() to get the unescaped version of the expression
+            if (is_string($part) && $expression instanceof Expression) {
+                $sql .= $expression->getExpression();
+
+                continue;
+            }
+
             // if it is a string, simply tack it onto the return sql "specification" string
             if (is_string($part)) {
                 $sql .= $part;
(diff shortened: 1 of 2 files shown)


library/Zend/Db/Sql/AbstractSql.php
```diff
@@ -108,6 +108,13 @@ protected function processExpression(ExpressionInterface $expression, PlatformIn
         $expressionParamIndex = &$this->instanceParameterIndex[$namedParameterPrefix];
 
         foreach ($parts as $part) {
+            // #7407: use $expression->getExpression() to get the unescaped version of the expression
+            if (is_string($part) && $expression instanceof Expression) {
```
freax: May be it'll be better:

    if (is_string($part)) {
        if ($expression instanceof Expression) {
            // #7407: use $expression->getExpression() to get the unescaped version of the expression
            $sql .= $expression->getExpression();
        } else {
            // simply tack it onto the return sql "specification" string
            $sql .= $part;
        }

        continue;
    } else (is_array($part)) {
        //
    } else {
        throw // Elements returned from getExpressionData() array must be a string or array.
    }

Martin-P: Also possible, but I tried to avoid nesting `if`.

freax: To me It looks better then `continue` and we get rid of `throw` in a middle of a function

Martin-P: When looking at different pull requests I often see the preferred way is to use `continue` instead of `if`/`elseif`/`else` structures. Personally I find the `continue` easier to read, but it is all just a mather of preference.

For now I will leave it as it is, but if the ZF2 maintainers want the other structure it is no problem and I can change it at that moment.

PR #7424, opened 2015-04-04
implement "@todo Ideally the helper should implement an interface." in FormCollection view helper

samsonasik:


samsonasik:
I've recreated the PR at https://github.com/zendframework/zend-form/pull/1

PR #7425, opened 2015-04-04
update variable name : $myParams to $pageParams

samsonasik:


library/Zend/Navigation/Page/Mvc.php
```diff
@@ -130,18 +130,18 @@ public function isActive($recursive = false)
                     $reqParams['controller'] = $reqParams[ModuleRouteListener::ORIGINAL_CONTROLLER];
                 }
 
-                $myParams   = $this->params;
+                $routeMatchParams   = $this->params;
```
Martin-P: These are **not** the parameters from the `RouteMatch`, but parameters from the `MvcPage` itself. The parameters from the `RouteMatch` are on [line 127](https://github.com/zendframework/zf2/pull/7425/files#diff-3a344f2c266dfd5b1af21a719cd58822R127) in the variable `$reqParams` (also not the best name, better is `$requestParams`).

If you want to rename `$myParams`, `$pageParams` is a better choice, because after all those are the parameters from the `MvcPage`.

samsonasik: done ;)

PR #7426, opened 2015-04-05
Fixed form collections with nested fieldsets that implements InputfilterProviderInterface

meNeither:
Here is a fix for the usage of collections and the `InputFilterProviderInterface`.

The problem was: If you're using `getInputFilterSpecification()` for your fieldsets and embed them in a collection, the input filters were never added to the form, resulting in errors when using `setValidationGroup()` or the inability to bind data to nested fieldsets.

The fix also ensures that infinite nested collections with `InputFilterProviderInterface`-fieldsets are working correctly.

### DIFF
--- a/library/Zend/Form/Form.php
+++ b/library/Zend/Form/Form.php
@@ -823,11 +823,7 @@ class Form extends Fieldset implements FormInterface
 
                             // We need to copy the inputs to the collection input filter
                             if ($inputFilter instanceof CollectionInputFilter) {
-                                foreach ($inputFilter->getInputs() as $filterName => $input) {
-                                    if (!$inputFilter->getInputFilter()->has($filterName)) {
-                                        $inputFilter->getInputFilter()->add($input, $filterName);
-                                    }
-                                }
+                                $inputFilter = $this->addInputsToCollectionInputFilter($inputFilter);
                             }
 
                             // Add child elements from target element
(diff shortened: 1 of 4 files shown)


library/Zend/Form/Form.php
```diff
@@ -807,8 +807,31 @@ public function attachInputFilterDefaults(InputFilterInterface $inputFilter, Fie
...
+                            $inputFilter->add($collectionContainerFilter, $name);
+
+                            // We need to copy the inputs to the collection input filter
+                            if ($inputFilter instanceof CollectionInputFilter) {
+                                foreach ($inputFilter->getInputs() as $filterName => $input) {
+                                    if (!$inputFilter->getInputFilter()->has($filterName)) {
+                                        $inputFilter->getInputFilter()->add($input, $filterName);
+                                    }
+                                }
+                            }
```
samsonasik: I think it can be copied into separate private method to reduce repetitive addition block code

meNeither: You're right. I added `addInputsToCollectionInputFilter`.

tests/ZendTest/Mvc/Controller/Plugin/FilePostRedirectGetTest.php
```diff
@@ -274,59 +274,6 @@ public function testReuseMatchedParametersWithSegmentController()
         $this->assertEquals(303, $prgResultRoute->getStatusCode());
     }
 
-    public function testFieldsetAmountInFormEqualsFieldsetsInInputFilter()
```
meNeither: I deleted this test, because now a `CollectionInputFilter` is used for collections and there is no need for checking the same amount of fieldsets and `InputFilter`s. Please correct me, if I am wrong.

Issue #7428, opened 2015-04-07
ClassReflection::getTraits() behaves inconsistently


stepashka69:
Hello,

The problem is that sometimes the function returns *array* and sometimes (when there are not traits) *null*.
It would be also nice to have a phpdoc for the function.

The problem was caught on the following code:
```
// $class is an instance of ClassReflection
foreach ($class->getTraits() as $trait) {
//...
}
```

The code gives PHP error "Invalid argument supplied for foreach()" when there are no traits in the class. 

The problem is in file ClassReflection.php line 185 . It returns null but it will be more logical to return $vals (empty array at that stage). 

```
    public function getTraits()
    {
        $vals = array();
        $traits = parent::getTraits();
        if (! $traits) {
            return; // It would be better to return $vals here
        }

        foreach ($traits as $trait) {
            $vals[] = new ClassReflection($trait->getName());
        }

        return $vals;
    }
```

samsonasik:
linked with #7431

GeeH:
This issue has been moved from the `zendframework` repository as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html
New issue can be found at: https://github.com/zendframework/zend-code/issues/77

PR #7431, opened 2015-04-07
zf7428 fixed issue with inconsistent getTrait() behaviour #7428

stepashka69:
1) Made getTraits() to behave the same was a parent method - return null only in case of error and empty array otherwise
2) Added phpdoc comment
3) Added tests for getTraits()

### DIFF
--- a/library/Zend/Code/Reflection/ClassReflection.php
+++ b/library/Zend/Code/Reflection/ClassReflection.php
@@ -187,7 +187,7 @@ class ClassReflection extends ReflectionClass implements ReflectionInterface
         $vals = array();
         $traits = parent::getTraits();
         if ($traits === null) {
-            return null;
+            return;
         }
 
         foreach ($traits as $trait) {
--- a/library/Zend/Code/Reflection/ClassReflection.php
+++ b/library/Zend/Code/Reflection/ClassReflection.php
@@ -177,12 +177,17 @@ class ClassReflection extends ReflectionClass implements ReflectionInterface
         return $methods;
     }
 
+    /**
+     * Returns an array of reflection classes of traits used by this class.
+     *
+     * @return array|null
+     */
     public function getTraits()
     {
         $vals = array();
         $traits = parent::getTraits();
-        if (! $traits) {
-            return;
+        if ($traits === null) {
+            return null;
         }
 
         foreach ($traits as $trait) {
--- a/tests/ZendTest/Code/Reflection/TestAsset/TestTraitClass3.php
+++ b/tests/ZendTest/Code/Reflection/TestAsset/TestTraitClass3.php
@@ -1,0 +1,29 @@
+<?php
+namespace ZendTest\Code\Reflection\TestAsset;
+
+//issue #7428
+trait TestTraitClass3
+{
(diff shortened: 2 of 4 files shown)


library/Zend/Code/Reflection/ClassReflection.php
```diff
@@ -177,12 +177,17 @@ public function getMethods($filter = -1)
...
+     *
+     * @return array|null
+     */
     public function getTraits()
     {
         $vals = array();
         $traits = parent::getTraits();
-        if (! $traits) {
-            return;
+        if ($traits === null) {
```
Martin-P: `$traits === null` is not possible here, because `ReflectionClass:getTraits()` returns `array` or `null` in case of an error (see [manual](http://php.net/manual/en/reflectionclass.gettraits.php)). `if (! $traits) {` is correct here.

library/Zend/Code/Reflection/ClassReflection.php
```diff
@@ -177,12 +177,17 @@ public function getMethods($filter = -1)
...
+            return null;
```
Martin-P: `return null;` was removed from classes because it is the same as only `return;`. See #7014.

library/Zend/Code/Reflection/ClassReflection.php
stepashka69: if `$traits` is empty array then `if (! $traits)` is `true` and the return of `getTraits()` is `null`. It is incorrect behaviour which broke a lot of unittests in my projects.

Martin-P: Ah, sorry, my mistake.

library/Zend/Code/Reflection/ClassReflection.php
stepashka69: fixed

steverhoades:
:+1: Thanks for catching that.

tests/ZendTest/Code/Reflection/TestAsset/TestTraitClass3.php
```diff
@@ -0,0 +1,29 @@
...
+	/**
+	* @param bool $autoFetchingAllowed
+	* @return Model_AbstractModel
+	*/
+	public function setDummy($dummy)
+	{
+		$this->dummy = boolval($dummy);
+		return $this;
+	}
+}
```
malukenho: Use spaces instead of tabs.

tests/ZendTest/Code/Reflection/TestAsset/TestTraitClass4.php
```diff
@@ -0,0 +1,26 @@
...
+
+	/**
+	* Constructor
+	*
+	* @param bool $other
+	*/
+	public function __construct($other)
+	{
+		$this->other = $other;
+	}
```
malukenho: Same above

stepashka69: Fixed that. I've also removed unnecessary code from test classes. The only check done on them is presence of traits. Other functions and variables are not needed.

PR #7439, opened 2015-04-09
Use === and !== for bool check

samsonasik:


DASPRiD:
Why did you close it again?

samsonasik:
because travis got many errors, can't check one by one for now

PR #7441, opened 2015-04-09
Soap fixes: throw exception in construct, camelcase fixes for name of classes

freax:
This PR provides following fixes:
- throw exception in construct instead of silent return from script in `Zend\Soap\Client\Common`
- camelcase for name of using classes in `Zend\Soap\Client\DotNet` and `Zend\Soap\Client\Local`

### DIFF
--- a/library/Zend/Soap/Client/Common.php
+++ b/library/Zend/Soap/Client/Common.php
@@ -10,10 +10,7 @@
 namespace Zend\Soap\Client;
 
 use SoapClient;
-
-if (! extension_loaded('soap')) {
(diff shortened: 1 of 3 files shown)


Maks3w:
@freax Add a unit test testing the exception is thrown and you will see why the exception never is thrown ;-)

If the extension is not loaded the class can't extend from the builtin SoapClient class. So you will receive a PHP Error

Maks3w:
Also please don't add cosmetic changes into the PR. instead create new PR with just those changes.

Issue #7443, opened 2015-04-10
Form collection - PriorityList iterator remove elements bug in populateValues


Qronicle:
When calling populateValues on a collection, the iterator loops over all current elements and checks whether they are still in the new data array. If not the element is removed from the collection. It seems however that removing the element causes problems with the iterator, as it might skip elements.

###### Example
A form is loaded with the collection having 3 fieldsets with names [6, 3, 4], but submitted is a new fieldset with name 7, with all other fieldsets removed. The resulting fieldset names after populateValues should be [7], but are [3, 7]

If I put a dump on the first line of the replaced/removed loop, It only gets in there two times for fieldsets 6 and 4. When I put the actual remove line in a comment, it gets in there for all fieldsets.

Tested in ZF 2.4.0. Probably related to the removal of the byName property.

Edit: Went a bit deeper and seems to be a problem with the Zend\Stdlib\PriorityList class, but can be fixed separately in the collection class by keeping a separate array with elements that need to be removed:

```php
        // Check to see if elements have been replaced or removed
        $removeElementNames = array();
        foreach ($this->iterator as $name => $elementOrFieldset) {
            if (isset($data[$name])) {
                continue;
            }
            if (!$this->allowRemove) {
                throw new Exception\DomainException(sprintf(
                    'Elements have been removed from the collection (%s) but the allow_remove option is not true.',
                    get_class($this)
                ));
            }
            $removeElementNames[] = $name;
        }
        foreach ($removeElementNames as $name) {
            $this->remove($name);
        }
```

adamlundrigan:
Fixed in #7446

PR #7446, opened 2015-04-14
bugfix removing multiple fieldsets does't work

svycka:
at the same time iterating collection items and removing causes problem:
When we have 3 items but we delete second and third. It fails and does not remove third element. So we have to move remove out of foreach

### DIFF
--- a/library/Zend/Form/Element/Collection.php
+++ b/library/Zend/Form/Element/Collection.php
@@ -209,6 +209,7 @@ class Collection extends Fieldset
         }
 
         // Check to see if elements have been replaced or removed
+        $removed = [];
         foreach ($this->iterator as $name => $elementOrFieldset) {
             if (isset($data[$name])) {
                 continue;
@@ -221,6 +222,10 @@ class Collection extends Fieldset
                 ));
             }
 
+            $removed[] = $name;
+        }
+
+        foreach ($removed as $name) {
             $this->remove($name);
         }
 
--- a/library/Zend/Form/Element/Collection.php
+++ b/library/Zend/Form/Element/Collection.php
@@ -209,7 +209,7 @@ class Collection extends Fieldset
         }
 
         // Check to see if elements have been replaced or removed
-        $removed = [];
+        $removed = array();
         foreach ($this->iterator as $name => $elementOrFieldset) {
             if (isset($data[$name])) {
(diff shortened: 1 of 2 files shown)


svycka:
fixes #7443

Maks3w:
Please add a unit test

Maks3w:
So the issue is modify and read the iterator (PriorityList) in the same loop.

Should this to be fixed in PriorityList instead?

svycka:
@Maks3w I don't think this should be allowed in PriorityList. Also don't see how this could be implemented.
```
foreach PriorityList
    remove
    remove
endforeach
PriorityList->commit
```
does not feel right

svycka:
@Maks3w if you don't like how it is fixed I could change:
```php
foreach ($this->iterator as $name => $elementOrFieldset) {
// to
foreach ($this->iterator->getIterator() as $name => $elementOrFieldset) {
```
this would also fix the issue.

Maks3w:
I delegate to other @zendframework/community-review-team member for their opinions.

To have a temportal `$removed` array is weird for me

svycka:
I have found commit witch introduced the bug 
PR: #6240 
commit: https://github.com/zendframework/zf2/commit/9a5dd6055120e50e8103fc8d037679edee2b48c1#diff-31fe6580a729ea0519189ac9ef45928aL212

library/Zend/Form/Element/Collection.php
```diff
@@ -221,6 +222,10 @@ public function populateValues($data)
                 ));
             }
 
+            $removed[] = $name;
+        }
+
+        foreach ($removed as $name) {
```
weierophinney: Why not just remove as we discover valid names?

weierophinney: Interestingly, because doing so causes unexpected behavior.

Maks3w: IMO this should be fixed in PriorityList instead

weierophinney: I've tried a few different approaches at this time. The problem is that doing removal while inside the loop leads to inconsistent results; the underlying iterator is not updated in place correctly. As such, the first pass is to determine which we can remove, and then we remove each.

Issue #4492, opened 2013-05-16
Form\Collection: allow remove all objects


thestanislav:
There is no way to remove all objects from collection at this time. According to the logic I need to pass an empty array to Zend\Form\Element\Collection::populateValues. But empty arrays are ignored https://github.com/zendframework/zf2/blob/master/library/Zend/Form/Element/Collection.php#L185

The next possible upcoming issue is that there is not way to create the html element in view script that would generate an empty array. May be some type conversion should be implemented to handle this.

svycka:
fixed with https://github.com/zendframework/zend-form/pull/14

PR #7447, opened 2015-04-14
Fixed: Form/Element/Collection can't remove all elements

svycka:
related issues: #7446 #4492

### DIFF
--- a/library/Zend/Form/Element/Collection.php
+++ b/library/Zend/Form/Element/Collection.php
@@ -195,11 +195,6 @@ class Collection extends Fieldset
             ));
         }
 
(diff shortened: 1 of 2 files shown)


svycka:
just fixed test because it was checking for empty collection to be empty.. maybe someone can help with fix?

svycka:
tests failing because of another bug fixed in #7446 when merged should be ok

svycka:
@weierophinney is it possible to merge this in 2.4.1? It is impossible to use Form/Element/Collection without workarounds.

svycka:
fixed with https://github.com/zendframework/zend-form/pull/14

Issue #7445, opened 2015-04-14
ZF 2.4.0 form validation BC break


AlexeyKosov:
I don't see any documentation notes about form validation BC break.
Here's the case:

        $form = new \Zend\Form\Form();
        $form->add(array(
            'type' => 'Text',
            'name' => 'email',
        ));
        $form->getInputFilter()->add(array(
            'name' => 'email',
            'required' => false,
            'allow_empty' => false,
            'continue_if_empty' => false,
        ));

        $data = array(
            'foo' => 'bar',
        );

        $form->setData($data);
        if ($form->isValid()) {
            echo "Valid\n";
        } else {
            echo "Invalid\n";
            var_dump($form->getMessages());
        }

Output:

ZF 2.3.7
```
   Valid
```
ZF 2.4.0
```
   Invalid

   array (size=1)
     'email' => 
       array (size=1)
         'isEmpty' => string 'Value is required and can't be empty' (length=36)
```

Martin-P:
This break is caused by PR #4567. The exact issue you are experiencing was already mentioned in the comments in that PR: https://github.com/zendframework/zf2/pull/4567#discussion_r4659825 

> That would allow this to work -- but it also means that order of operations becomes critical -- you would be required to set the "allow empty" flag *after* setting the required flag if you wanted to set it to `false`.

Unfortunatelly it was only said in words it was fixed, but no tests have been added to actually show it was fixed :disappointed: 

##### isValid() returns `false`:
```php
$input = new \Zend\InputFilter\Input();
$input->setName('email')
      ->setRequired(false)
      ->setAllowEmpty(false)
      ->setContinueIfEmpty(false);
var_dump($input->isValid()); // false
```

##### different ordering and now isValid() returns `true`:
```php
$input = new \Zend\InputFilter\Input();
$input->setName('email')
      ->setAllowEmpty(false)
      ->setContinueIfEmpty(false)
      ->setRequired(false);
var_dump($input->isValid()); // true
```

Pittiplatsch:
@Martin-P From https://github.com/zendframework/zf2/pull/4567#issuecomment-19606897 (the answer to your quoted comment):
>I believe the correct way to solve is [...] to set "allow empty" flag to the reverse of "required" by default. In fact, the factory was already doing this if the user didn't explicitly specify allow_empty.
>
>I've updated the factory too to ensure that regardless of the order of the specification, the allow_empty flag specified by the user (if any) is always respected..

[...]

PR #7448, opened 2015-04-14
Allow use of setRequired() at different positions

Martin-P:
`Zend\InputFilter\Input` would return `false`/`true` depending of when `setRequired()` was called. This PR fixes that so it does not matter anymore when `setRequired()` is called.

This is **not** a fix for #7445, but an issue I noticed looking into issue #7445.

### DIFF
--- a/library/Zend/InputFilter/Input.php
+++ b/library/Zend/InputFilter/Input.php
@@ -147,7 +147,7 @@ class Input implements InputInterface, EmptyContextInterface
     public function setRequired($required)
     {
         $this->required = (bool) $required;
-        $this->setAllowEmpty(!$required);
+
         return $this;
     }
 
@@ -323,6 +323,10 @@ class Input implements InputInterface, EmptyContextInterface
         $empty     = ($value === null || $value === '' || $value === array());
 
         if ($empty && $this->allowEmpty() && !$this->continueIfEmpty()) {
+            return true;
+        }
+
+        if ($empty && !$this->isRequired()) {
             return true;
         }
 
--- a/tests/ZendTest/InputFilter/InputTest.php
+++ b/tests/ZendTest/InputFilter/InputTest.php
@@ -407,4 +407,32 @@ class InputTest extends TestCase
         $this->assertTrue($input2->isRequired());
         $this->assertTrue($input2->allowEmpty());
     }
+
+    /**
+     * @group 7445
+     */
+    public function testInputIsValidWhenUsingSetRequiredAtStart()
+    {
+        $input = new Input();
+        $input->setName('foo')
+              ->setRequired(false)
+              ->setAllowEmpty(false)
+              ->setContinueIfEmpty(false);
+
+        $this->assertTrue($input->isValid());
+    }
+
+    /**
+     * @group 7445
+     */
+    public function testInputIsValidWhenUsingSetRequiredAtEnd()
+    {
+        $input = new Input();
+        $input->setName('foo')
+              ->setAllowEmpty(false)
+              ->setContinueIfEmpty(false)
+              ->setRequired(false);
+
+        $this->assertTrue($input->isValid());
+    }
 }


Martin-P:
Did not run the tests in `BaseInputFilterTest`. The failing unittests have nothing to do with `BaseInputFilter` and shouldn't even be there IMO, they should be in `InputTest` because those tests only test `InpuTest` issues... I will look at the tests and move them them to the correct location.

library/Zend/InputFilter/Input.php
```diff
@@ -326,6 +326,10 @@ public function isValid($context = null)
             return true;
         }
 
+        if ($empty && !$this->isRequired()) {
```
samsonasik: <code>!$this->isRequired()</code> can be change with <code>$this->isRequired() === false</code> imo

Martin-P: And `$empty` should be `$empty === true`? :wink: I see no valid reason for comparing a method, which returns a boolean value by definition, to `true`/`false`, but perhaps you can provide a use case for that?

samsonasik: for example : 
```php
$foo = null;
if (!$foo) { echo "not"; } else { echo "yes"; }
```
produce 'not'.

but 

```php
$foo = null;
if ($foo === false) { echo "not"; } else { echo "yes"; }
```
produce 'yes'.

Well, It indeed problematic, I tried to provide a PR about it in last days, but it got many errors, as many codes should cast it to real (bool).

Martin-P:
@weierophinney 
Can you provide answers to some questions, because I think some things are wrong in the `InputFilter` section.

1. Should `allow_empty` be the opposite of `required`? If a value is not required does not mean it should be empty when it *is* present, so this can not be assumed by default.
2. Determining if a combination of `allow_empty` and `required` is possible should be done in `isValid()` and **never** inside the methods `setRequired()` and `setAllowEmpty()`. When you do that in those methods you make the validation dependent on the ordering of settings.
3. Should `null` be considered an empty value ([line 323 of `Zend\InputFlter\Input`](https://github.com/zendframework/zf2/blob/66af8c73089ce88766898393a63de915fcdfc658/library/Zend/InputFilter/Input.php#L323)? Forms can not send `null` as a value. `null` should mean the element is not present and thus is for `required` and not to determine if a value is empty.
4. Can you clarify what should happen in this case:
```php
$input = new Input();
$input->setName('foo')
      ->setRequired(false)
      ->setAllowEmpty(false)
      ->setValue('');

$this->assertFalse($input->isValid());
```

Do you agree with my points? I am willing to attempt to fix this, but it is too much work to create a fix and having to revert many things afterwards. It looks like it will be BC break...

library/Zend/InputFilter/Input.php
Martin-P: I see your point, but the casting should be done in `isRequired()`. Checking for `false` here would mean `null` is considered `true` while `null` is normally considered `false`. That is probably also the reason you get many errors.

Instead of checking everything with `===` or `!==`, it is better to force the methods to return the value which is defined in the docblocks. A method that can return `null` when a the docblock explicitely says `@return bool` is wrong IMO.

weierophinney: I agree with @Martin-P here; the `isRequired()` method is defined to return a boolean, so we can safely use the `!` operator here without doing a strict comparison.

Strict comparisons are primarily of use when type may be in question; it's not here.

weierophinney:
> Should allow_empty be the opposite of required?

No. `required` means that the input *must* be present. However, by default, we handle that in three ways:

- If the input is not present at all, then it evaluates as invalid, always.
- If the input is present, but a `null` value, empty string, or other value that evaluates as `empty()`, we consider it "invalid". **UNLESS**
- `allow_empty` is `true`. Essentially, we're acknowledging that the input was provided, but that no value was present.

That also answers your third point: "Should null be considered an empty value (line 323 of Zend\InputFlter\Input?" Yes, forms cannot send a `null` value, but input filters can be used for more than forms; one very common use case is APIs, where the data may be coming from JSON, XML, etc.

If those conditions are met — `required` is `true` and `allow_empty` is `true` — we then have to determine how to validate the value. The default is that it's considered valid with the "empty" value provided. This is now where the `continue_if_empty` flag comes into play: if this is `true`, we run the filters and validators on the value.

To break down the rules prior to 2.4:

- If `required === true`:
  - `allow_empty === true` will allow `empty()` values to validate
  - `continue_if_empty === false && allow_empty === true` will mark `empty()` values as valid
  - `continue_if_empty === true && allow_empty === true` will run filters and validators on `empty()` values.
- If `required === false`:
  - `allow_empty` has no meaning, and can be ignored
  - `continue_if_empty`, when `true`, will execute filters and validators even when the value evaluates to `empty()`

That second-to-last one is tricky. We currently default `allow_empty` to `false`, and if `required === false`, that means an empty value is valid. This is where I think the whole "order of operations" bit became ambiguous, and a decision was made that `setRequired()` should toggle the value of `allow_empty`. My inclination is that as I have stated it above is correct: `allow_empty` should only have meaning when `required === true`.

As to number 4:

> Can you clarify what should happen in this case:
> ```php
> $input = new Input();
> $input->setName('foo')
>       ->setRequired(false)
>       ->setAllowEmpty(false)
>       ->setValue('');
> 
> $this->assertFalse($input->isValid());
> ```

Per my notes above, I disagree; the assertion should be `$this->assertTrue($input->isValid())`. `allow_empty` should have no meaning if `required === false`.

weierophinney:
@Martin-P Are you in IRC, by any chance? or can you be? I'd like to discuss in real-time today, if possible.

Issue #7450, opened 2015-04-15
PartialLoop undefined offset notice


vixriihi:
I'm having nesting partial loops and there seems to be quite a few notices appearing. Source for them seems to be [this line in PartialLoop helper](https://github.com/zendframework/zf2/blob/master/library/Zend/View/Helper/PartialLoop.php#L117).

PR #7451, opened 2015-04-15
partialLoop notice #7450

vixriihi:
Here is a test and a fix that will show the notice in the nested partialloop #7450

### DIFF
--- a/tests/ZendTest/View/Helper/PartialLoopTest.php
+++ b/tests/ZendTest/View/Helper/PartialLoopTest.php
@@ -348,6 +348,36 @@ class PartialLoopTest extends TestCase
         }
     }
 
+    /**
+     * @group 7450
+     */
+    public function testNestedPartialLoopsNestedArray()
+    {
+        $data = array(
+            array(
+                'obj' => array(
+                    'helper' => $this->helper,
+                    'message' => 'foo1',
+                    'data' =>
+                        array(
+                            array(
+                                'message' => 'foo2'
+                            )
+                        ),
+                    )
+                )
+            );
+
+        $view = new View();
+        $view->resolver()->addPath($this->basePath . '/application/views/scripts');
+        $this->helper->setView($view);
+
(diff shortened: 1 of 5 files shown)


tests/ZendTest/View/Helper/PartialLoopTest.php
```diff
@@ -348,6 +348,36 @@ public function testNestedCallsShouldNotOverrideObjectKey()
...
+            );
+
+        $view = new View();
+        $view->resolver()->addPath($this->basePath . '/application/views/scripts');
+        $this->helper->setView($view);
+
+        $result = $this->helper->__invoke('partialLoopParentObject.phtml', $data);
+        $this->assertContains('foo1', $result, $result);
+        $this->assertContains('foo2', $result, $result);
+
```
samsonasik: remove unused blank line before close brace <code>}</code>

tests/ZendTest/View/Helper/PartialLoopTest.php
```diff
@@ -348,6 +348,36 @@ public function testNestedCallsShouldNotOverrideObjectKey()
...
+                    'message' => 'foo1',
+                    'data' =>
+                        array(
+                            array(
+                                'message' => 'foo2'
+                            )
+                        ),
+                    )
+                )
+            );
```
malukenho: align?

adamlundrigan:
This looks good to me, with just one question:  The current behaviour is a notice is emitted and `objectKey` is set to null.  In your code the existing value of `objectKey` will remain unchanged.  I'm not too familiar with `partialLoop` so I'm unsure of whether or not this presents an issue.  If `objectKey` is always NULL preceding this point in it's execution then it does not, but if `objectKey` is reused through the nested stack it could be an issue.

vixriihi:
Nether am I :) But the existing ```objectKey``` is set to null just before checking if the key exist [on this line](https://github.com/vixriihi/zf2/blob/zf7450/library/Zend/View/Helper/PartialLoop.php#L129), so it shouldn't be a problem.

PR #7452, opened 2015-04-17
MvcEvent in listeners

snapshotpl:
I have tried to inject MvcEvent to listener by constructor in ServiceManager. Unfortunately it doesn't exists yet in Application. 

I have other idea to create MvcEvent by service manager. It will be easy to extend event if necessary. What do you think?

### DIFF
--- a/library/Zend/Mvc/Application.php
+++ b/library/Zend/Mvc/Application.php
@@ -141,20 +141,20 @@ class Application implements
 
         $listeners = array_unique(array_merge($this->defaultListeners, $listeners));
 
+        // Setup MVC Event
+        $this->event = new MvcEvent();
+        $this->event->setTarget($this);
+        $this->event->setApplication($this)
+                    ->setRequest($this->request)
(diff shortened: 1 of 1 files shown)


Martin-P:
`MvcEvent` is always injected in the callback function, that way it is already always available. Perhaps you can provide a use case?

Regarding the creation by `ServiceManager` and extending the `MvcEvent`: shouldn't you just use `SharedEventManager` for that situation? `SharedEventManager` can be used for triggering your custom events.

Besides the above this PR needs a unittest.

weierophinney:
You shouldn't compose the `MvcEvent` in your listeners, period.

The event is _passed_ to listeners when they are triggered. The event passed **can** be a different instance entirely. As such, the canonical version is the one provided at time of invocation, not at time of instantiation.

> I have other idea to create MvcEvent by service manager. It will be easy to extend event if necessary. What do you think?

No; that's completely against the design of an event system, per my explanation above.

PR #7457, opened 2015-04-22
Fix controller namespace in event manager

0xPaul:
Controller was only registering the root namespace, e.g. `Foo` instead of `Foo\Bar` for a class named `Foo\Bar\Baz`.

### DIFF
--- a/library/Zend/Mvc/Controller/AbstractController.php
+++ b/library/Zend/Mvc/Controller/AbstractController.php
@@ -162,7 +162,7 @@ abstract class AbstractController implements
     {
         $className = get_class($this);
 
-        $nsPos = strpos($className, '\\') ?: 0;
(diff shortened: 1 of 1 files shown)


Martin-P:
Perhaps you can add:
- a unittest which shows the correct namespace is returned
- a separate unittest which shows the wrong namespace is **not** returned

That way regressions can be avoided.

0xPaul:
If by *you*, you mean *someone*, then yes. I'm sure *someone* can do those things.

Martin-P:
> If by *you*, you mean *someone*, then yes. I'm sure *someone* can do those things.

I'm sure *someone* can create unittests, but this is not *someone*s PR, it is your PR. So I meant you, @0xPaul  :wink:

GeeH:
New issue can be found at: https://github.com/zendframework/zend-mvc/issues/171

Issue #7206, opened 2015-02-12
JsonModel for JsonStrategy should have ability to configure json_encoding params 


netroby:
JSON_UNESCAPED_UNICODE can prevent string be unescaped, it is very useful when develop web apps.

zf2/library/Zend/View/Model/JsonModel.php

But how could we setting this param for json_encode ?

netroby:
Currently, i can only extends JsonModel as MyJsonModel, then override the serialize() method with 
```php
namespace netroby\utils;
use Zend\View\Model\JsonModel as JsonModel;
use Traversable;
use Zend\Stdlib\ArrayUtils as ArrayUtils;

class NetrobyJsonModel extends JsonModel
{

    public function serialize()
    {
        $variables = $this->getVariables();
        if ($variables instanceof Traversable) {
            $variables = ArrayUtils::iteratorToArray($variables);
        }
        if (null !== $this->jsonpCallback) {
            return $this->jsonpCallback.'('.Json::encode($variables).');';
        }
        return JsonConverter::toJson($variables);
    }
}

```

Issue #6848, opened 2014-11-06
Is it possible to support JSON_UNESCAPED_UNICODE?


twmobius:
On Zend\Json\Json it uses ```json_encode()``` which encodes all UTF-8 to ```\uXXXX```. After PHP 5.4.0 there is the possibility of having ```JSON_UNESCAPED_UNICODE``` to avoid this encoding.

Is it somehow possible to add this functionality on Zend\Json\Json. Ideally I would like to have JSON generated from Apigility to be unescaped.

twmobius:
Are there any plans on adding support for JSON_UNESCAPED_UNICODE?

GeeH:
This issue has been closed as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

PR #7458, opened 2015-04-23
add support of JSON_UNESCAPED_UNICODE

ikhrustalev:
add unescapedUnicode option to Zend\Json::encode #6848 #7206

### DIFF
--- a/library/Zend/Json/Encoder.php
+++ b/library/Zend/Json/Encoder.php
@@ -255,7 +255,9 @@ class Encoder
         // 0x08 => \b
         // 0x0c => \f
         $string = str_replace(array(chr(0x08), chr(0x0C)), array('\b', '\f'), $string);
-        $string = self::encodeUnicodeString($string);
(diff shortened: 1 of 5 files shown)


Ocramius:
While I understand why this was implemented, I don't see an actual use-case.

What is it for, exactly? Seems like an un-needed security risk to me...

rznzippy:
Hello Marco,

First of all I'ld like to say thank you for your time.

We're using this option for two reasons:
- we have integration with android application that unable to recognise encoded unicode characters
- with this option we able to put human-readable request/response text to text log files

For now we have to have our own implementation of Json and JsonModel classes, because there is no easy way to pass **JSON_UNESCAPED_UNICODE** option into json_encode function inside Json class.

So main pros that I can see:
- be able to integrate with some applications that doesn't support encoded unicode characters
- human-readable output in console and log files
- reduced size of json string

weierophinney:
I agree with @Ocramius that this is a security risk. As such, you're already following the appropriate course of action, @rznzippy : extending and overriding the renderer and model present in the framework. (This is why we have an IoC container and configurable renderers!)

PR #7460, opened 2015-04-27
Update PHP CS Fixer to v1.7

keradus:


Maks3w:
This is supersede by #7477 and https://github.com/zendframework/zf2/commit/d2bda8f1a0b36222641fa79b6ca5c72c15cc1236

Thank you anyway

keradus:
@Maks3w , I can't see any relation with #7477.
Also, on develop branch there is no caching enabled.

Maks3w:
No there was not relation. Last commit in the branch do a global update.

About the caching thing if you want you can create a PR only with that. However I don't see benefits with the inminent component split

keradus:
If sth can be done faster why not take advantages of it?
Even after component split ;)

Maks3w:
I would like to hear how that cache works. With component split php-cs-fixer is only executed once and temporal files are not persisted across builds.

Maks3w:
Also in the actual configuration all commands are executed in parallel at the same time so a lot of threads are prone to rebuild the cache due race conditions.

keradus:
One can want to run the fixer locally before sending PR.
In brief when using cache file is not trying to fix again if file itself wasn't changed and using same PHP CS Fixer version. More info available at tool's page.

Maks3w:
@keradus If you think is needed please open a PR with only that feature and describe the benefits and possible issues.

keradus:
It is not needed, just cool to have.
If you see no point of it then cool, no need for new PR.

PR #7476, opened 2015-05-01
[test] Replace assert(True|False) with more appropriate assertions where possible

Maks3w:
This PR aims to remove uses of `asserTrue` and `assertFalse`.

PHPUnit provides alternative assertions specific for certain cases and provides useful information about the nature of the failure.


weierophinney:
Can we maybe leave the editorial comments out of the titles? I wouldn't say that `assertTrue/assertFalse` were _evil_, but maybe misguided. Additionally, recall that many of these tests pre-date ZF2 itself, and the number of assertions in PHPUnit 2 and 3 were far fewer than what we have today, and they've worked over a number of PHPUnit migrations.

Still merging, but I'm going to remove the editorialization from the PR title.

Maks3w:
@weierophinney Ok.

PR #7479, opened 2015-05-03
Proper URI rendering for Zend/Http

### LINKED ISSUE #7472: Zend/Http/Request::renderRequestLine() does not take query parameters into account
The [renderRequestLine()](https://github.com/zendframework/zf2/blob/master/library/Zend/Http/Request.php#L517) function of Zend/Http/Request counterintuitively does not include query parameters set via setQuery() in the request line it returns. Is this intentional or is it something for which a patch would likely be accepted?

afeder:
Fixes #7472.

### DIFF
--- a/library/Zend/Http/Request.php
+++ b/library/Zend/Http/Request.php
@@ -68,6 +68,16 @@ class Request extends AbstractMessage implements RequestInterface
     protected $fileParams = null;
 
     /**
+     * Configuration array, set using ::setOptions()
+     *
+     * @var array
+     */
+    protected $config = array(
+        'argseparator'    => null,
+        'rfc3986strict'   => false
+    );
+
+    /**
      * A factory that produces a Request object from a well-formed Http Request string
      *
      * @param  string $string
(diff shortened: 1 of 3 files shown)


Martin-P:
This PR needs unittests.

afeder:
Sorry, what is a "PR"? Does any guidance exist as to what kinds of unit tests would be needed?

samsonasik:
need rebase

afeder:
samsonasik: Indeed. Done.

afeder:
I added a test for the issue described in #7472.

samsonasik:
need rebase again

afeder:
Rebase done.

library/Zend/Http/Request.php
```diff
@@ -521,13 +580,60 @@ public function isFlashRequest()
...
+                }
+                return $queryString;
+            }
+        }
+    }
+
+    /**
+     * Return the formatted URI for this http request
+     *
+     * @return string
```
sasezaki: `@return HttpUri` , not ?

Martin-P:
IMO this needs PR needs many more tests.

There are multiple new methods introduced in `Zend\Http\Request` which are not tested at all:
- `setOptions`
- `setArgSeparator`
- `getArgSeparator`
- `renderQueryString`
- `renderUri`

A big chunk of code in `Zend\Http\Client` has been removed and new code is added, but no tests have been added for this.

The only test which has been added is a very basic query string example. What happens with more advanced query strings? Does that work too or will it fail?

afeder:
If you share with me what tests you are looking for I shall add them right away.

The big chunks of code being removed or added is little more than migrating the URI building logic from the Client class into the Request class.

I've found two tests pertaining to this code which I've now copied over to the tests for the Request class. I've also expanded the query test to cover the case of multiple query parameters being set, and duplicated this test for the renderUri() and renderQueryString() functions.

library/Zend/Http/Request.php
```diff
@@ -166,6 +176,55 @@ public static function fromString($string, $allowCustomMethods = true)
...
+     * @return Request
+     * @throws Request\Exception\InvalidArgumentException
+     */
+    public function setOptions($options = array())
+    {
+        if ($options instanceof Traversable) {
+            $options = ArrayUtils::iteratorToArray($options);
+        }
+        if (!is_array($options)) {
+            throw new Request\Exception\InvalidArgumentException('Config parameter is not valid');
```
sasezaki: should be `Exception\InvalidArgumentException`, please add test too.

afeder:
I fixed the exception in the setOptions() function per sasezaki's comment. As for test, none exist in the original code in Client.php, but I added a basic one anyway.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-http/issues/68

PR #7482, opened 2015-05-04
Fix undefined PGSQL_CONNECT_ASYNC only present in PHP 5.6

Maks3w:
Issue introduced in #7396


Maks3w:
Travis-CI build https://travis-ci.org/Maks3w/zf2/builds/61212315

PR #7483, opened 2015-05-04
[#7396] Functionality only works in PHP >= 5.6

weierophinney:
PR #7396 introduced code that only works in 5.6; this patch does the following:
- Raises a RuntimeException when the new `setType()` method is invoked in
  versions less than 5.6.
- Skips unit tests of that method when the PHP version is less than 5.6.

This patch also fixes several long lines, flagged by phpcs when I was editing.

### DIFF
--- a/tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
+++ b/tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
@@ -86,9 +86,14 @@ class ConnectionTest extends \PHPUnit_Framework_TestCase
      */
     public function testSetConnectionTypeException()
     {
+        if (version_compare(PHP_VERSION, '5.6', 'lt')) {
+            $this->markTestSkipped('Functionality under test only works in 5.6 and above');
+        }
+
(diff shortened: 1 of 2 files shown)


tests/ZendTest/Db/Adapter/Driver/Pgsql/ConnectionTest.php
```diff
@@ -86,9 +86,14 @@ public function testGetConnectionStringEncodeSpecialSymbol()
      */
     public function testSetConnectionTypeException()
     {
+        if (version_compare(PHP_VERSION, '5.6', 'lt')) {
```
samsonasik: there are <code>if (PHP_VERSION_ID) </code> check used instead of <code>version_compare</code> in other source, for example : https://github.com/zendframework/zf2/blob/master/library/Zend/Stdlib/JsonSerializable.php#L12. 

what we will be using for consistencies ?

Maks3w:
#7482 provides a fix too but from a "feature exists" approach

Maks3w:
Merged #7482 instead

PR #7484, opened 2015-05-04
[2.5][travis-ci] Remove PHP 5.3 and PHP 5.4 support

Maks3w:


.travis.yml
```diff
@@ -1,8 +1,6 @@
 language: php
 
 php:
-  - 5.3
-  - 5.4
```
weierophinney: We need to wait to remove these until we split the components. I'm not merging any commits to develop until after they are split, and will only be doing merges to develop on individual components once we do.

PR #7475, opened 2015-04-30
[test] Place test constants in travis.xml.dist and travis.xml.travis

Maks3w:
Close #7383

PR #7486, opened 2015-05-05
[test] Use environment variables instead constants

Maks3w:
This PR use environment vars instead constants. Supersede #7475

Thanks to this change bug #7485 has been disclosure


Maks3w:
Change applied in components repo

PR #7493, opened 2015-05-06
[travis-ci] Print list of PEAR packages installed and their versions

Maks3w:


weierophinney:
Why? We're not installing any at this time — unless this is to detail which extensions were installed?

weierophinney:
> unless this is to detail which extensions were installed?

This appears to be the motivation.

Maks3w:
@weierophinney This is for a PHP7 incompatibility with `Archive_Tar` and I wanna to know what version of the package is installed

PR #7507, opened 2015-05-08
[http] Allow serialize any character on cookies

Maks3w:


library/Zend/Http/Header/GenericHeader.php
```diff
@@ -53,6 +53,10 @@ public static function splitHeaderLine($headerLine)
             throw new Exception\InvalidArgumentException('Header must match with the format "name:value"');
         }
 
+        if (! HeaderValue::isValid($parts[1])) {
```
Maks3w: @weierophinney I applied the same control point as was in Mail

Header folding is not an issue here as wasn't be on Mail

PR #7508, opened 2015-05-08
Factory for Input filter factory

snapshotpl:


library/Zend/Mvc/Service/InputFilterFactoryFactory.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Zend\Mvc\Service;
+
+use Zend\InputFilter\Factory;
+use Zend\ServiceManager\FactoryInterface;
+use Zend\ServiceManager\ServiceLocatorInterface;
+
+class InputFilterFactoryFactory implements FactoryInterface
+{
```
samsonasik: I think can be added with
```php
/**
 * {@overrideDoc}
 */
```
here

Martin-P:
What is the use case for this PR? This single factory creates a new dependency to `Zend\Mvc` which was not there before. See: [composer.json](https://github.com/zendframework/zf2/blob/cabf3db7d94427598b0eff92c2c560a54c4a0047/library/Zend/InputFilter/composer.json#L17-L19)

snapshotpl:
I always need to inject dependencies by hand. 

I don't see any new dependencies in Input Filter

Martin-P:
The new dependency is introduced with [line 56 in `Zend\InputFilter\InputFilterAbstractServiceFactory`](https://github.com/snapshotpl/zf2/blob/6c2e90542fad2f7468089da560db047dd445f7d8/library/Zend/InputFilter/InputFilterAbstractServiceFactory.php#L56):
```php
$factory   = $services->get('InputFilterFactory');
```
The definition of `InputFilterFactory` is defined on [line 64 of `Zend\Mvc\Service\ServiceListenerFactory`](https://github.com/snapshotpl/zf2/blob/6c2e90542fad2f7468089da560db047dd445f7d8/library/Zend/Mvc/Service/ServiceListenerFactory.php#L64):
```php
'InputFilterFactory'             => 'Zend\Mvc\Service\InputFilterFactoryFactory',
```
Without `Zend\Mvc` the `ServiceManager` does not know about `InputFilterFactory` and the instantiation via `Zend\InputFilter\InputFilterAbstractServiceFactory` will fail because of that.

GeeH:
New issue can be found at: https://github.com/zendframework/zend-mvc/issues/169

PR #7510, opened 2015-05-10
[mail/mime] Fix content-type has invalid characters in field value. Fix #7503

### LINKED ISSUE #7503: Mail header - boundary issue (related to ZF2015-04)
The ZF2015-04 security fix (2.4.1) has broken my Mail implementation. I'm creating a 'multipart/alternative' MIME part with a boundary setting. (I basically copied the Mail spec of ZF1, since ZF2's impementation had a lot of issues for me)

I'm not really sure on how to set the boundary with the new checks in place.

Relevant snippet:
```php
$mail            = new \Zend\Mail\Message();
$bodyMessage     = new MimeMessage();

$multiPartContentMessage = new MimeMessage();
$multiPartContentMessage->addPart($this->text);
$multiPartContentMessage->addPart($this->html);
    
$multiPartContentMimePart          = new Part($multiPartContentMessage->generateMessage());
$multiPartContentMimePart->charset = $this->encoding;
// Following line breaks the GenericHeader check
$multiPartContentMimePart->type    = 'multipart/alternative;' . PHP_EOL . ' boundary="' .
    $multiPartContentMessage->getMime()->boundary() . '"';

$bodyMessage->addPart($multiPartContentMimePart);
$mail->getHeaders()->addHeaderLine('Content-Transfer-Encoding', Mime::ENCODING_QUOTEDPRINTABLE);

$mail->setBody($bodyMessage);
```
Exception thrown on last line of that snippet:
```
Zend\Mail\Header\Exception\InvalidArgumentException

File:
    project/vendor/zendframework/zendframework/library/Zend/Mail/Header/GenericHeader.php:65

Message:
    Invalid header value detected
```

Maks3w:
This bug has been introduced in v2.4.1 due the security patch ZF2015-04 for to prevent CRLF injection.

Zend\Mime\Part line was originally introduced in ZF1@583 with the following commit message

> - Correctly handle multipart/alternative to close #59:
>   - If both text and html body present, create multipart/alternative part
>   - If text, html, and attachments are present, create multipart/alternative
>     part for text+html, but mark email as multipart/mixed
> - Note: Zend_Mime::LINEEND as \r\n is causing issues on multipart emails (too
>   many line breaks); switching to \n fixes the issue, but doesn't follow
>   standards. Need reviewers.

Zend_Mime::LINEEND was replaced by $EOL argument on ZF1@598 with the following commit message

> - Refactoring of Zend_Mail/Zend_Mime; mail message construction happens in
>   transport. Fixes #134 and #106.

RFC 2045 section 5.1 defines the syntax of Content-Type Header field as this (only relevant parts of the syntax posted):

> In the Augmented BNF notation of RFC 822, a Content-Type header field
> value is defined as follows:
> 
> ```
> content := "Content-Type" ":" type "/" subtype
>            *(";" parameter)
>            ; Matching of media type and subtype
>            ; is ALWAYS case-insensitive.
> 
> parameter := attribute "=" value
> 
> attribute := token
>              ; Matching of attributes
>              ; is ALWAYS case-insensitive.
> 
> value := token / quoted-string
> 
> token := 1*<any (US-ASCII) CHAR except SPACE, CTLs,
>          or tspecials>
> 
> tspecials :=  "(" / ")" / "<" / ">" / "@" /
>               "," / ";" / ":" / "\" / <">
>               "/" / "[" / "]" / "?" / "="
>               ; Must be in quoted-string,
>               ; to use within parameter values
> ```

Finally on ZF2 has been duplicated the same behavior on Zend\Mail\Header\ContentType

https://github.com/zendframework/zf2/commit/297ba86090fda8ef4cc5d343e49869b196e158aa#diff-78a765eeca5dcb0c4d8fa594675622f6R76

### DIFF
--- a/library/Zend/Mail/Header/ContentType.php
+++ b/library/Zend/Mail/Header/ContentType.php
@@ -69,7 +69,7 @@ class ContentType implements HeaderInterface
             $values[] = sprintf('%s="%s"', $attribute, $value);
         }
 
-        return implode(';' . Headers::FOLDING, $values);
+        return implode(';', $values);
     }
 
     public function setEncoding($encoding)
@@ -165,7 +165,7 @@ class ContentType implements HeaderInterface
         if (isset($this->parameters[$name])) {
             return $this->parameters[$name];
         }
-        return;
+        return null;
     }
 
     /**
--- a/library/Zend/Mime/Part.php
+++ b/library/Zend/Mime/Part.php
@@ -439,8 +439,7 @@ class Part
         }
 
         if ($this->boundary) {
-            $contentType .= ';' . $EOL
-                          . " boundary=\"" . $this->boundary . '"';
+            $contentType .= '; boundary="' . $this->boundary . '"';
         }
 
         $headers[] = array('Content-Type', $contentType);
--- a/tests/ZendTest/Mail/Header/ContentTypeTest.php
+++ b/tests/ZendTest/Mail/Header/ContentTypeTest.php
@@ -63,7 +63,7 @@ class ContentTypeTest extends \PHPUnit_Framework_TestCase
         $string = $header->toString();
 
         $this->assertContains("Content-Type: application/x-unit-test;", $string);
-        $this->assertContains(";\r\n charset=\"us-ascii\"", $string);
+        $this->assertContains(';charset="us-ascii"', $string);
     }
 
     public function testExtractsExtraInformationFromContentType()
--- a/tests/ZendTest/Mail/MessageTest.php
+++ b/tests/ZendTest/Mail/MessageTest.php
@@ -538,7 +538,7 @@ class MessageTest extends \PHPUnit_Framework_TestCase
 
         $this->assertTrue($headers->has('content-type'));
         $header = $headers->get('content-type');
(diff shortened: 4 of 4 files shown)


Maks3w:
Seems break each parameter in his own landing is only a cosmetic change hacking the header folding feature.

Fix:
 - Generation of headers with EOL in `Zend\Mime\Part`

Don't fix:
 - Generation of headers with EOL in `Zend\Mail\Headers\Content-type`
 - Headers with folding sequences when call `HeaderInterface::fromString()`

weierophinney:
Just a note for those coming to this issue: ContentType currently uses `Headers::FOLDING`, and not a hard-coded inline sequence: https://github.com/zendframework/zf2/blob/master/library/Zend/Mail/Header/ContentType.php#L72

library/Zend/Mail/Header/ContentType.php
```diff
@@ -165,7 +165,7 @@ public function getParameter($name)
         if (isset($this->parameters[$name])) {
             return $this->parameters[$name];
         }
-        return;
+        return null;
```
weierophinney: Why have a return statement at all?

PR #7515, opened 2015-05-11
[ldap] fix case of Converter::toLdapBoolean calls

campersau:
PHP functions are not case sensitive but this looks better to me.

Is there a tool which can check for such things automatically?


Martin-P:
Because "looks better" is a very subjective description I think a more solid argument is the fact the method definition is spelled as [`toLdapBoolean`](https://github.com/zendframework/zf2/blob/014caed443a45d264830c6f8502f5c4987e1a0bd/library/Zend/Ldap/Converter/Converter.php#L159).

Issue #7502, opened 2015-05-08
Zend\Http\Header\SetCookie::setMaxAge doesn't conform to RFC6265 section 5.2.2


coud:
It should not throw an exception when passing a negative value or non integer.

Zend\Http\Client doesn't work as expected because of this.

Martin-P:
@Maks3w I think you added the wrong label to this issue. It is about `Http`, not `Mail`.

Martin-P:
For easy reference the exact link and text: http://tools.ietf.org/html/rfc6265#section-5.2.2
> If the first character of the attribute-value is not a DIGIT or a "-" character, ignore the cookie-av.

weierophinney:
Those of you watching this issue: please do the following in your project root:

```console
$ composer require zendframework/zendframework:dev-master@dev
```

and then test. #7507 was just merged, which should fix the issue.

Martin-P:
@weierophinney
Please reopen this issue. You have mistaken this with the security issue, but this issue is completely unrelated to that. This issue is about setting a negative or non integer value as max-age value. The PR you are referring to does not fix this:
- where is the unittest which tests a negative value for max-age does not throw an exception?
- where is the unittest which tests a non integer value for max-age does not throw an exception?

Maks3w:
@Martin-P Unittests are in the bag of the next person which want to help.

weierophinney:
@Martin-P Sorry; thought this was related to ZF2015-04; I see it actually was an issue previously as well. I'll not include a fix in 2.4.2, then.

nathanjosiah:
I'm not sure this is the right behavior change for this. Can somebody check my PR and provide some feedback?

Martin-P:
Looking at the specs for the second time I notice the specs are about the user agent: http://tools.ietf.org/html/rfc6265#section-5.2.2
> If the attribute-name case-insensitively matches the string "Max-Age", the **user agent** MUST process the cookie-av as follows.

The question is:
do we need server-side validation of `Max-Age` at all or should we just allow any value and let the user agent decide what to do with it according to the specs?

Martin-P:
@weierophinney @Maks3w You two are competing over this issue? Close it or open it? :smile:

Maks3w:
Nop, just Github perfom a cross repository close because 81ccf07 has been forward to zend-http

This may to be repeat again because is automatic

nathanjosiah:
[...]

PR #7517, opened 2015-05-12
Fix for 7502

nathanjosiah:
SetCookie Max-Age header is now RFC 6265 compliant and discards negative and non-numeric values.


nathanjosiah:
Related to #7502

PR #7525, opened 2015-05-14
Travis: add language translations monitor

Slamdunk:
Keeping up-to-date the translations is tough.
But at least we have to try.

I've added a test executed only in Travis that outputs languages with differences from `en` one.

I've used `markTestIncomplete`: the test will **always** pass, so the build result is never affected.

I'm not sure the test placement in the folder tree is elegant, and likely it needs to be rethought for ZF3 and component splitting.

### DIFF
--- a/tests/ResourcesTest.php
+++ b/tests/ResourcesTest.php
@@ -1,4 +1,11 @@
 <?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @link      http://github.com/zendframework/zf2 for the canonical source repository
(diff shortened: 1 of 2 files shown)


Slamdunk:
Result: https://travis-ci.org/zendframework/zf2/jobs/62553053#L1163

tests/ResourcesTest.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
```
samsonasik: missing license header here : 
```
/**
 *
 * @copyright Copyright (c) 2005-2014 Zend Technologies USA Inc. (http://www.zend.com)
 * @license   http://framework.zend.com/license/new-bsd New BSD License
 */
```

PR #7527, opened 2015-05-15
Reestablish the connection to the SMTP server if a timeout occurred

andrewisplinghoff:
Holding a SMTP connection for a long time without activity  (>= 5 min) causes a server-side timeout (https://tools.ietf.org/html/rfc1123#page-61). This pull request improves the Zend Mailer to automatically reconnect to the server if a timeout occurred. I added a unit test that verifies the new functionality. I also wrote a test application that connects to a real SMTP server and triggers the server-side timeout by  sleeping a few minutes between sending mails. You can find this test application here: https://github.com/andrewisplinghoff/zend-timeout-test

### DIFF
--- a/library/Zend/Mail/Protocol/Smtp.php
+++ b/library/Zend/Mail/Protocol/Smtp.php
@@ -299,8 +299,17 @@ class Smtp extends AbstractProtocol
     public function rset()
     {
         $this->_send('RSET');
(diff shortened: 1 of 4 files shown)


Maks3w:
I'll reject this because error management should be do on user code and not in the library.

The only thing I see useful its close the session when error occur. Could you detail more why was that needed?

PR #7529, opened 2015-05-15
update PHP CS Fixer to v1.8

keradus:


Maks3w:
I'll discard this since the project will change a lot and probaly the composer.lock will be removed

keradus:
I see. OK ;)
Still don't know where to send PR - here or to custom components ;) CONTRIBUTING on main zf2 repo don't tell me that ;)

Maks3w:
By the way we have to open a rfc about the CS tool to use. Compare performance, features, semantic versioning, ...

PR #7530, opened 2015-05-15
enable cache for PHP CS Fixer

keradus:
Replaces #7418


Maks3w:
Is it possible enable/disable cache with a command line switch?

keradus:
It is possible at v2.0@dev, not on v1.x@stable

PR #7532, opened 2015-05-15
fix project's CS

keradus:
why, why the CS on develop branch are broken :(


keradus:
hurray, dead on `phpcov`

gianarb:
I don't know the future of this repository.. But now all components are splitted in different repository.. 
I don't know if this fix is applicable

keradus:
In Symfony we have components splitted into different repositories, but they are subtrees and read-only, all patches go to the root project.
Don't know how it is here.

harikt:
@keradus it was the same a few days back. But now all components accepts PR and issues. So this repo may be staying for historical reasons probably from now on or may be removed ? But the PR can be send to individual component I guess.

prolic:
:+1:

PR #7536, opened 2015-05-17
Add isGone method to \Zend\Http\Response

asgrim:
As per the discussion in zendframework/zf2#7368 this change adds a new
method `\Zend\Http\Response->isGone()` to determine if the status code is
a "410 Gone". This works exactly the same as the similar methods like
`isNotFound` simply by checking the status code.

This change takes the opinion that a 410 Gone should not be included in
the `isNotFound()` method. Further discussion can be seen in the issue.

### DIFF
--- a/library/Zend/Http/Response.php
+++ b/library/Zend/Http/Response.php
@@ -400,6 +400,16 @@ class Response extends AbstractMessage implements ResponseInterface
     }
 
     /**
+     * Does the status code indicate the resource is gone?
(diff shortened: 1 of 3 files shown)


gianarb:
Hi! Can you push this PR into the correct repository?
https://github.com/zendframework/zend-http

Thanks! :)

asgrim:
This is now open on zendframework/zend-http#3

PR #7544, opened 2015-05-23
[Json/Server] Return the invalid parameters list

oswaldocarvalho:
If the user forgot or send some required parameter with a wrong name the server returns the error ERROR_INVALID_PARAMS.

This request added a invalid parameters list in the json response.

Result before

``` javascript
{
    "error":{
        "code":-32602,
        "message":"Invalid params",
        "data":null
    },
    "id":"8",
    "jsonrpc":"2.0"
}
```

Result after

``` javascript
{
    "error":{
        "code":-32602,
        "message":"Invalid param(s): [estimatedMonths, location, specieId, coatTypeId]",
        "data":null
    },
    "id":"8",
    "jsonrpc":"2.0"
}
```

*Please do not consider this pull request if this implementation has not been made before for security reasons.

### DIFF
--- a/library/Zend/Json/Server/Server.php
+++ b/library/Zend/Json/Server/Server.php
@@ -541,14 +541,21 @@ class Server extends AbstractServer
             }
 
             $orderedParams = array();
+            $invalidParams = array();
             foreach ($reflection->getParameters() as $refParam) {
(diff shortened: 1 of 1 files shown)


GeeH:
New issue can be found at: https://github.com/zendframework/zend-json/issues/25

Issue #7526, opened 2015-05-14
ValidatorChain::getValidators() returns array, not PriorityQueue


kehigginbotham:
Per Issue #6496, Zend\Validator\ValidatorChain was modified to use a PriorityQueue internally, as opposed to an array.  Per the documentation and annotations, ValidatorChain::getValidators() should return a PriorityQueue of the validators.  However, the method is written to return an array ( extracted from the PriorityQueue ). Is this intended functionality ( documentation lagging behind ), or should a Priority Queue indeed be returned?

Using zf2 v2.4.0, PHP v5.6.3.

Example ( $input is an object of Zend\InputFilter\Input ):
`gettype( $input->getValidatorChain()->getValidators() )`
returns
`string(5) "array"`

Martin-P:
The strange thing is the changes in #6496 introduced both the docblock `@return PriorityQueue` as well as `return $this->validators->toArray(PriorityQueue::EXTR_DATA);`. Perhaps it was done for BC purposes?

Maks3w:
Duplicate https://github.com/zendframework/zend-validator/issues/42. Follow https://github.com/zendframework/zend-validator/issues/42 for continue the discussion.

PR #7546, opened 2015-05-24
#7526 Update docblock @return array for getValidators() in ValidatorChain

MadCat34:


Martin-P:
Linking to #7526.

**Note**:
It is not yet clear if the docblock is wrong here or if the return value should be an instance of `PriorityQueue`. Ping @weierophinney

Maks3w:
Duplicate https://github.com/zendframework/zend-validator/pull/43

PR #7548, opened 2015-05-25
'type' in Input config may be plugin name

halaxa:
Support for input plugin name in `'type'` input configuration key for `Zend\InputFilter\Factory::createInput()`.


halaxa:
Moved to zendframework/zend-inputfilter#2.

PR #7549, opened 2015-05-25
Fix loading constraint data from Oracle

vixriihi:
This Fixes loading constraint data from Oracle database


samsonasik:
could you make a pr against https://github.com/zendframework/zend-db too ? thanks.

PR #7598, opened 2015-06-30
Added zend-ldap

heiglandreas:
Somehow the zend-ldap component was missing. 

If that was not done intentionally this PR fixes that issue. If it _was_ intentionally, a hint in the README.md would be nice to know that it is missing :wink:


YP28:
I'm not entirely sure, but I think it was moved to the suggest section due to the requirement of the PHP LDAP extension.

heiglandreas:
Yeah. I've seen that later, @YP28 :smile: That's why I've then closed the PR :wink:

PR #7612, opened 2015-07-22
Fixed typo

radarhere:


samsonasik:
not sure, but changelog entry _probably_ generated via pull requests in milestone

radarhere:
You're saying that if this PR was merged, it would simply be automatically overwritten in the future? If so, fair enough, and feel free to close this. Thanks.

samsonasik:
ping @weierophinney as should now about changelog generation :)

weierophinney:
@samsonasik We don't overwrite old entries; we only generate *new* entries.

PR #7665, opened 2016-01-27
Updated dependencies to `^2.5`

weierophinney:
Previously, we were using `~2.5.0`, which is equivalent to `>=2.5.0,<2.6.0`.  This is problematic, as a number of components are at 2.6 and 2.7 versions, and some of those newer versions include security fixes. By being too restrictive in the constraints, we are locking users to insecure versions, which is problematic.

As none of these new minor versions introduces backwards compatibility breaks, this patch updates to use the `^2.5` constraint to allow `>=2.5.0,<3.0.0` versions to be installed for all components.

The plan is to do an immediate 2.5.3 release with these changes.


JulienBreux:
@noopable Thx for this infos!

weierophinney:
@noopable Can you please provide an issue report against zend-mvc detailing the specific errors you're seeing?

The removal of `ServiceLocatorAwareInterface` implementation only occurred for *controllers*, and we modified the initializer that injects service locator aware controllers to also duck-type against the `setServiceLocator()` method, which ensures no breakage. If you're referring to the deprecation notice that occurs from *using* the composed service locator in controllers, this deprecation notice is intentional, to signal you to change your code to make it future-proof. Similarly, for general services that are `ServiceLocatorAware`, we raise a deprecation notice in the initializer that injects the service locator; again, this is intentional. For more information, please see [my blog post on deprecating ServiceLocatorAware](https://mwop.net/blog/2016-04-26-on-locators.html).

Finally, the `ViewManager` is intended as an internal detail for wiring listeners and services related to views, and wasn't really expected to be an extension point, so I'm curious to see what you may have been using!

If you're seeing any other errors, please report them against zend-mvc so we can evaluate them. Thanks!

noopable:
@weierophinney 
Thank you  for your blog post.
I think so too.

My issue was duplicate with this. It was already resolved. 
https://github.com/zendframework/zend-mvc/issues/86
I thought that removing Di settings for V3 was not bug  at that time. My posts was in just after..

Another,my private CMS changes the theme after routing.It sometimes needs to change the layout script or the view paths . My codes use ViewManager::getLayout() , ViewManager::getResolver() .
I was trembling, but it is ok now, I  can change my codes for these.

WelingtonM:
Thanks

Issue #7699, opened 2016-05-02
PHP Version 7.0.5 incompatible with ZF2 before 2.5


Douglasdc3:
As reported in zend-code package (https://github.com/zendframework/zend-code/issues/49) PHP version  7.0.5 is not compatible if you are using zend annotations.

PR #7700, opened 2016-05-02
Exclude PHP 7.0.5 from allowed PHP versions (#7699)

Douglasdc3:
This fixes issue #7699 same fix as applied in https://github.com/zendframework/zend-code/issues/49


grizzm0:
This repo is just a meta package. The fix should be included per component and not in the zf2 repo.

Douglasdc3:
You're right we'll update the zend-code legacy branch

PR #7715, opened 2016-06-28
Update CONTRIBUTING.md

FraGoTe:
Updating github join url.


darielopper (approved):

weierophinney:
This change would need to be propagated to every single repository. The better time to do that is when we migrate to Laminas. Closing for now.

michalbundyra:
@weierophinney I've applied that change to our migration script.

Issue #7746, opened 2016-12-09
Zend_Http_Client::setCookies fails with array argument, in release-2.4


niloct:
`Zend_Http_Client::setCookies` incorrectly loops through array argument, making `addCookie` to fail, because `setCookies` pass the array index, numeric, as the name of the cookie.

PR #7747, opened 2016-12-09
Fix bug #7746

niloct:
*Fixing bug: Zend_Http_Client::setCookies fails with array argument*

This bugfix makes it pass the entire array to `addCookie`.

### DIFF
--- a/library/Zend/Http/Client.php
+++ b/library/Zend/Http/Client.php
@@ -578,9 +578,7 @@ class Client implements Stdlib\DispatchableInterface
     {
         if (is_array($cookies)) {
             $this->clearCookies();
-            foreach ($cookies as $name => $value) {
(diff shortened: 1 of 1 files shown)


Ocramius:
Linking #7746

Ocramius:
This is a patch against a very old version of the framework.

We'd need:

 * tests
 * a patch against a newer version, if the bug applies there too

Also, cookies should indeed be indexed by name here. Maybe it's more of an API problem?

niloct:
Sorry for poor linking.

I just used composer to install the whole framework in a project:

```
{
    "require": {
        "zendframework/zendframework": "^3.0"
    }
}

```

darielopper (approved):

weierophinney:
The 2.4 release has reached EOL; closing.

PR #7758, opened 2017-02-17
apply zend-config ^3.0

samsonasik:


Ocramius (approved):

Ocramius:
Thanks @samsonasik!

Note: the milestone `4.0.0` was set because of BC breaks in the upgraded component.

PR #7784, opened 2018-07-02
Update hostname validator TLD

kachar:
Backport of https://github.com/zendframework/zend-validator/blob/master/src/Hostname.php


froschdesign:
The [Long Term Support](https://framework.zend.com/long-term-support) for version 2.4 ended on 2018-03-31. Therefore, I do not see a chance to backport the updates of the validator.

@weierophinney is [working](https://discourse.zendframework.com/t/zend-framework-2-5-x-support-status/586/2) on a new support policy, which will clarify the new time frame of support.

weierophinney:
@kachar As noted, 2.4 is past its EOL, and we will not be doing new releases of it.

We recommend upgrading to at least the v2.5 series of Zend Framework, and, ideally, installing only the packages you actually use. This will allow Composer to pull in updates as soon as they happen, and provide you clearer upgrade paths.

